PR #541, opened 2014-04-22
Fixed Utf8UrlEncoder to respect BNF for Uri syntax components on RFC3986

kelvl:
Previously Utf8UrlEncoder and RequestBuilderBase tries to percent encode everything other than a small set of ASCII characters (SAFE_ASCII). 

This adds a few more methods into Utf8UrlEncoder to encode specific Uri syntax components with the allowed characters specified on rfc3986.

RequestBuilderBase's toURI is also modified to use those methods when encoding different parts of the URI.

### DIFF
--- a/src/main/java/com/ning/http/util/UTF8UrlEncoder.java
+++ b/src/main/java/com/ning/http/util/UTF8UrlEncoder.java
@@ -124,6 +124,10 @@ public class UTF8UrlEncoder {
         USERINFO[':'] = true;
     }
 
+    /**
+     * host        = IP-literal / IPv4address / reg-name
+     * reg-name    = *( unreserved / pct-encoded / sub-delims )
+     */
     private final static boolean[] HOSTPORT = new boolean[128];
     static {
         for(int i=0; i < 128; i++)
@@ -131,6 +135,9 @@ public class UTF8UrlEncoder {
             HOSTPORT[':'] = true;
     }
 
+    /**
+     * pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
+     */
     private final static boolean[] PCHAR = new boolean[128];
     static {
         for(int i=0; i < 128; i++)
@@ -139,6 +146,9 @@ public class UTF8UrlEncoder {
         PCHAR['@'] = true;
     }
 
+    /**
+     * query       = *( pchar / "/" / "?" )
+     */
     private final static boolean[] QUERY = new boolean[128];
     static {
         System.arraycopy(PCHAR, 0, QUERY, 0, 128);
@@ -146,6 +156,11 @@ public class UTF8UrlEncoder {
         QUERY['?'] = true;
     }
 
+    /**
+     * used to encode individual name and value parts of query
+     * =& are characters used to separate name and value pairs
+     * + needs to be encoded to differentiate from encoded spaces
+     */
     private final static boolean[] QUERY_PARAM = new boolean[128];
     static {
         System.arraycopy(QUERY, 0, QUERY_PARAM, 0, 128);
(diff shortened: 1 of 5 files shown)


src/main/java/com/ning/http/util/UTF8UrlEncoder.java
```diff
@@ -23,25 +23,178 @@
...
+     * used to encode individual name and value parts of query
+     * =& are characters used to separate name and value pairs
+     * + needs to be encoded to differentiate from encoded spaces
+     */
+    private final static boolean[] QUERY_PARAM = new boolean[128];
+    static {
+        System.arraycopy(QUERY, 0, QUERY_PARAM, 0, 128);
+        QUERY_PARAM['='] = false;
+        QUERY_PARAM['+'] = false;
+        QUERY_PARAM['&'] = false;
```
chadselph: `;` also needs to be encoded here. Many severs will parse it the same as `&`.

src/main/java/com/ning/http/util/UTF8UrlEncoder.java
```diff
@@ -55,25 +208,57 @@ public static String encode(String input) {
         return sb.toString();
     }
 
+    public static StringBuilder encodeAppendScheme(StringBuilder sb, String scheme) {
```
chadselph: the existing method was `appendEncoded`. Was this supposed to be `appendEncodedScheme`? (same question for all of the following `encodeAppend*` methods.

src/main/java/com/ning/http/util/UTF8UrlEncoder.java
slandelle: Then those servers are not compatible with modern browsers? I just check Chrome 40 and foo=bar;baz is sent as is.

chadselph: How did you check this? I tested on Chrome 39 with this HTML:

```html
<form action="http://localhost:1190/">
  <input type="hidden" name="smile" value=";-) %">
  <input type="submit">
</form>
```

and loaded up netcat on port 1190. First line is

```
GET /?smile=%3B-%29+%25 HTTP/1.1
```

chadselph: If you just checked by typing that into the query string, that's kinda like saying that typing "foo=bar&baz" will send as is. Yes, it will send that, but a server will see `foo` as just `bar`, and if you wanted `"bar&baz"` you'd have to URL-encode the ampersand. Same with ";" for some servers.

slandelle:
@chadselph I realize several things.
First, AHC tries too to encode query.
Then, there seems to be a confusion between [RFC3986](http://www.ietf.org/rfc/rfc3986.txt) and slightly different [application/x-www-form-urlencoded encoding algorithm](http://www.w3.org/TR/html5/forms.html#application/x-www-form-urlencoded-encoding-algorithm) which is described in the HTML spec.

I plan on simplifying things:
* query encoding should only be performed when we're totally sure the character MUST be encoded (space, non US-ASCII... will sort this out). For example, let's just leave % as is. We should stop with this crazy decode-encode roundtrip. This is what browsers seem to be doing when one write some unencoded query in the address bar.
* disableUrlEncoding should only impact explicitly passed queryParams where we can tell (name, value) pairs appart (so we know `=` and `&` meaning).
* make sure  `x-www-form-urlencoded` bodies use HTML spec

@jfarcand WDYT?

jfarcand:
@slandelle +1 That code has always be problematic.

slandelle:
Closing as this PR's author never replied and there was some concerns about it.
Now that we have our own Uri implementation, it can't be merged.

I'll just come up with our own solution.

@chadselph Why did you become interested in this PR in the first place? Did you run into an issue?

chadselph:
The author is a friend and former co-worker. We were running this patch and a customer complained about the ";" character.

I think his approach would have been appropriate for an RFC3986 validator, but as you discovered, `x-www-form-urlencoded` is really a spec on top of the URI spec. It's not 100% clear which characters need to be encoded in form param key/values. Unfortunately I don't know the original reason for this PR (which characters he was trying to preserve) and I don't think he will remember either. As far as I can tell, encoding all "reserved" characters wouldn't be a problem.

slandelle:
@chadselph Sorry for taking too long to review this. By the time I could check it, code base had changed too much to make it mergeable.

Then, I think AHC can make its best to compensate for poorly built url. But then, it's still the user's responsibility. When in doubt, I really advice using a proper RFC3986 implementation like https://github.com/palominolabs/url-builder.

Then, I've cleaned up things on our side: https://github.com/AsyncHttpClient/async-http-client/issues/808.

PR #818, opened 2015-02-18
Pass the InetSocketAddress instance to the onDnsResolved extensions callback

mobileAgent:
In order to extract the IP address that was used for the connection without resolving it again from outside AHC, this callback can surface the remote address instance.


slandelle:
Good idea, thanks!

PR #820, opened 2015-02-20
Netty: Fix releasing limit for connections pool per host

kullfar:
I've found a bug in async-http-client.
1.set 'maxConnections' and 'maxConnectionsPerHost' simultaneously.
2.try to GET the same host multiple times. With channels auto-closed after each request (with no chance here: tryToOfferChannelToPool https://github.com/AsyncHttpClient/async-http-client/blob/async-http-client-1.9.10/src/main/java/com/ning/http/client/providers/netty/channel/ChannelManager.java#L285)
3.Semaphore freeChannelsForHost (here https://github.com/AsyncHttpClient/async-http-client/blob/async-http-client-1.9.10/src/main/java/com/ning/http/client/providers/netty/channel/ChannelManager.java#L142) won't be decremented on channel close and you'll get 
"Caused by: java.io.IOException: Too many connections per host XXX"

Because there is no mapping in channelId2KeyPool
(https://github.com/AsyncHttpClient/async-http-client/blob/async-http-client-1.9.10/src/main/java/com/ning/http/client/providers/netty/channel/ChannelManager.java#L140)

I think it should be added here, while "registerOpenChannel" (https://github.com/AsyncHttpClient/async-http-client/blob/async-http-client-1.9.10/src/main/java/com/ning/http/client/providers/netty/channel/ChannelManager.java#L384)

prove-of-fail can be found here https://github.com/kullfar/async-test
steps to reproduce:
git clone 'https://github.com/kullfar/async-test.git'
cd async-test
mvn clean install
java -jar target/async-test-1.0-SNAPSHOT.jar

You will get "Caused by: java.io.IOException: Too many connections per host 3" line shortly

If you'll compile my pull request version and change pom.xml to use 1.9.11-SNAPSHOT version, there will be no Exceptions.

ps. It's the pull request to the 1.9.x branch (just because I use it ritgh now in prod), but as I can see 2.x branch needs to be fixed as well

### DIFF
--- a/src/main/java/com/ning/http/client/providers/netty/channel/ChannelManager.java
+++ b/src/main/java/com/ning/http/client/providers/netty/channel/ChannelManager.java
@@ -155,8 +155,8 @@ public class ChannelManager {
         }
 
         if (maxConnectionsPerHostEnabled) {
-            freeChannelsPerHost = new ConcurrentHashMap<String, Semaphore>();
-            channelId2KeyPool = new ConcurrentHashMap<Integer, String>();
+            freeChannelsPerHost = new ConcurrentHashMap<>();
+            channelId2KeyPool = new ConcurrentHashMap<>();
         } else {
             freeChannelsPerHost = null;
             channelId2KeyPool = null;
@@ -381,8 +381,11 @@ public class ChannelManager {
             getFreeConnectionsForHost(poolKey).release();
     }
 
-    public void registerOpenChannel(Channel channel) {
+    public void registerOpenChannel(Channel channel, final String poolKey) {
         openChannels.add(channel);
+        if (maxConnectionsPerHostEnabled) {
+            channelId2KeyPool.put(channel.getId(), poolKey);
+        }
     }
 
     private HttpClientCodec newHttpClientCodec() {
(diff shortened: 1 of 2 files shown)


slandelle:
> 2.use apache http client

You mean Netty, right?

kullfar:
sorry, completetly misleading step. I mean
<dependency>
  <groupId>org.apache.httpcomponents</groupId>
  <artifactId>httpclient</artifactId>
  <version>4.3.6</version>
</dependency>

in pom.xml from prove-of-fail

have just removed it
https://github.com/kullfar/async-test/commit/96fb464f61f1ff94756b038a15c319ddc3dc14ae

just ignore ;-)

kullfar:
removed this step "use apache http client" from the pull request description

slandelle:
Fix seems legit, but could you add a test for it, please?

kullfar:
sure, will do.
I'll return to that on Tuesday.
Thx!

slandelle:
Thanks!

slandelle:
Adding the test myself. Thanks a lot for reporting and fixing!

kullfar:
Awesome! You are fast! ;-)Thanks a lot

PR #823, opened 2015-02-20
Check in additional oauth tests

wsargent:
Adds a package level constructor for OAuthSignatureCalculator to directly set the nonce and timestamp for testing, and adds some tests to ensure that the requestbuilder is using the signature calculator correctly.


slandelle:
I like the test, but I'm not found of adding 2 members just for tests. Why not simply delegate nonce and timestamp computation to protected methods that tests can override?

Note that you declared testTimestamp as a Long wrapper, but then handle it as a primitive, causing a possible NPE.

Thanks!

slandelle:
@wsargent I'll merge and clean up as I suggested. Thanks!

slandelle:
Ported on 1.9.x as of 692bad165714cb51f244eec6988781cd1f2204f1

wsargent:
Thanks!

PR #830, opened 2015-02-26
Fix #829 NullPointerException with GrizzlyAsyncHttpProvider

### LINKED ISSUE #829: NullPointerException with GrizzlyAsyncHttpProvider(v1.8) when WebSocket is closed
The Grizzly ProtocolHandler (grizzly-websockets-2.3.18) org.glassfish.grizzly.websockets.ProtocolHandler has a bug when it close the webSocket. 

```
public GrizzlyFuture<DataFrame> close(int code, String reason) {
    return send(new ClosingFrame(code, reason),
            new EmptyCompletionHandler<DataFrame>() {

                @Override
                public void failed(final Throwable throwable) {
                    webSocket.onClose(null);
                }

                @Override
                public void completed(DataFrame result) {
                    if (!maskData) {
                        webSocket.onClose(null);
                    }
                }
            });
}
```

 The DataFrame is not set and when it comes to the GrizzlyAsyncHttpProvider it throws NullPointerException. 

```
public void onClose(org.glassfish.grizzly.websockets.WebSocket gWebSocket, DataFrame dataFrame) {
            try {
                if (ahcListener instanceof WebSocketCloseCodeReasonListener) {
                    ClosingFrame cf = ClosingFrame.class.cast(dataFrame);
                    WebSocketCloseCodeReasonListener.class.cast(ahcListener).onClose(webSocket, cf.getCode(), cf.getReason());
                } else {
                    ahcListener.onClose(webSocket);
                }
            } catch (Throwable e) {
                ahcListener.onError(e);
            }
}
```

laszlohordos:
The DataFrame is not set and when it comes to the GrizzlyAsyncHttpProvider it throws NullPointerException.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -591,7 +591,7 @@
         <compiler.exclude>com/ning/http/client/providers/grizzly/*.java</compiler.exclude>
         <test.compiler.exclude>com/ning/http/client/async/grizzly/*.java</test.compiler.exclude>
         <javadoc.package.exclude>com.ning.http.client.providers.grizzly</javadoc.package.exclude>
-        <grizzly.version>2.3.16</grizzly.version>
+        <grizzly.version>2.3.19</grizzly.version>
         <source.property>1.5</source.property>
         <target.property>1.5</target.property>
         <surefire.version>2.12</surefire.version>
--- a/src/main/java/com/ning/http/client/FluentCaseInsensitiveStringsMap.java
+++ b/src/main/java/com/ning/http/client/FluentCaseInsensitiveStringsMap.java
@@ -184,7 +184,7 @@ public class FluentCaseInsensitiveStringsMap implements Map<String, List<String>
      * @return This object
      */
     public FluentCaseInsensitiveStringsMap replace(final String key, final String... values) {
-        return replace(key, Arrays.asList(values));
+        return replace(key, (Collection<String>)Arrays.asList(values));
     }
 
     /**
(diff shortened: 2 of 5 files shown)


oleksiys:
Hey, I'll review the fix and commit ASAP.
Thank you!

laszlohordos:
Hi, 

I created the [GRIZZLY-1746](https://java.net/jira/browse/GRIZZLY-1746) issue with the proposed fix. Unfortunately my commit has the hardcoded '1011' code (server is terminating the connection). It's mandatory because the interface requires an int but obviously this code is wrong. Normal local socket closure generates 1000 (normal closure) and this happens when I shut down my client normally. I think there is no good default number. I'd use -1 to indicate to the developer there is something wrong.

oleksiys:
I've integrated the Grizzly fix, let me release Grizzly 2.3.19 tomorrow, do the integration and then apply the AHC part.
Thanks!

oleksiys:
Hi, do you need the fix to be ported to 1.8.x, or you can use 1.9.x?
Thank you.

laszlohordos:
Hi, 

This fix is not necessary for 1.9 any more. The #833 pull request has been merged which upgrades 1.9 to Grizzly 2.3.19 and it has the GRIZZLY-1746 fix. The #832 pull request is still pending which does the same to 1.8. After this update the previous versions of Grizzly won't be compatible so this situation should not happen. I'm using the 1.8 because I may need Java 5 but if not I'd like to use 1.9.

Thank you fixing and releasing Grizzly 2.3.19.

oleksiys:
So we don't need this change anymore, right?

PR #832, opened 2015-03-04
Update Grizzly version

laszlohordos:
The latest 2.3.19 grizzly release contains very important fixes but this library is not compatible with that version.


oleksiys:
thank you for the patch!

PR #853, opened 2015-04-09
[1.9.x] Fix #852: query paramters should not be removed from the redirec...

### LINKED ISSUE #852: Inconsistent behavior in redirect after dropping setRemoveQueryParamsOnRedirect
That feature was dropped from 1.9.x in https://github.com/AsyncHttpClient/async-http-client/commit/e1233ac3249f11e827a599ea5b0c3f0d3672af4b but as you can see there on Netty's side  the false flag survived while on Grizzly's it was the true flag, so Grizzly removes query parameters on redirects.

afelisatti:
...t URI.


oleksiys:
thank you!

PR #863, opened 2015-04-22
ISSUES-860 - Grizzly 2.3.19 fails when using SSL

pablolagreca:


oleksiys:
Hi,
this code is needed because of the other issue.
It would really help if you can provide the reproducer. IMO they issue you observe should be fixed in different place.

Thanks.

slandelle:
No info from user for 6 months, closing.

PR #864, opened 2015-04-24
fixed method compatibility with java 8

andreaTP:
This prevent me to run "mvn install" with a Jdk 8, seems they broke compatibility over the method used.


slandelle:
@andreaTP I preferred casting into the super type instead of using old `Enumeration`, but thanks for reporting. Next step will be to fix Javadoc so one doesn't have to use `-Xdoclint:none`.

andreaTP:
Perfect! Thanks!

PR #902, opened 2015-06-02
Make SocketChannel and EventLoopGroup configurable

sschepens:
Add support for specifying SocketChannel class, allow any EventLoopGroup except Oio


slandelle:
Thanks!

PR #914, opened 2015-06-14
fix unnecessary percent encoding

alalwww:
#### before
```java
normalized parameters:
 csv=aaa%[25]2Cbbb&file=vacation.jpg&oauth_consumer_key=dpf43f3p2l4k3l03&oauth_nonce=9940pd%2F933%2B3jh%3D%3D&oauth_signature_method=HMAC-SHA1&oauth_timestamp=1191242096&oauth_token=nnch734d00sl2jdk&oauth_version=1.0&size=original

signature base string:
 POST&http%3A%2F%2Fphotos.example.net%2Fphotos&csv%3Daaa%25[25]2Cbbb%26file%3Dvacation.jpg%26oauth_consumer_key%3Ddpf43f3p2l4k3l03%26oauth_nonce%3D9940pd%252F933%252B3jh%253D%253D%26oauth_signature_method%3DHMAC-SHA1%26oauth_timestamp%3D1191242096%26oauth_token%3Dnnch734d00sl2jdk%26oauth_version%3D1.0%26size%3Doriginal
```

#### after
```java
normalized parameters:
 csv=aaa%[]2Cbbb&file=vacation.jpg&oauth_consumer_key=dpf43f3p2l4k3l03&oauth_nonce=9940pd%2F933%2B3jh%3D%3D&oauth_signature_method=HMAC-SHA1&oauth_timestamp=1191242096&oauth_token=nnch734d00sl2jdk&oauth_version=1.0&size=original

signature base string:
 POST&http%3A%2F%2Fphotos.example.net%2Fphotos&csv%3Daaa%25[]2Cbbb%26file%3Dvacation.jpg%26oauth_consumer_key%3Ddpf43f3p2l4k3l03%26oauth_nonce%3D9940pd%252F933%252B3jh%253D%253D%26oauth_signature_method%3DHMAC-SHA1%26oauth_timestamp%3D1191242096%26oauth_token%3Dnnch734d00sl2jdk%26oauth_version%3D1.0%26size%3Doriginal
```

### DIFF
--- a/api/src/main/java/org/asynchttpclient/oauth/OAuthSignatureCalculator.java
+++ b/api/src/main/java/org/asynchttpclient/oauth/OAuthSignatureCalculator.java
@@ -32,6 +32,7 @@ import org.asynchttpclient.SignatureCalculator;
 import org.asynchttpclient.uri.Uri;
 import org.asynchttpclient.util.Base64;
 import org.asynchttpclient.util.StringUtils;
+import org.asynchttpclient.util.Utf8UrlDecoder;
 import org.asynchttpclient.util.Utf8UrlEncoder;
 
 /**
@@ -139,12 +140,12 @@ public class OAuthSignatureCalculator implements SignatureCalculator {
 
         if (formParams != null) {
             for (Param param : formParams) {
-                allParameters.add(param.getName(), param.getValue());
+                allParameters.add(param.getName(), Utf8UrlDecoder.decode(param.getValue()).toString());
             }
         }
         if (queryParams != null) {
             for (Param param : queryParams) {
-                allParameters.add(param.getName(), param.getValue());
+                allParameters.add(param.getName(), Utf8UrlDecoder.decode(param.getValue()).toString());
             }
         }
         return allParameters.sortAndConcat();
--- a/api/src/test/java/org/asynchttpclient/oauth/OAuthSignatureCalculatorTest.java
+++ b/api/src/test/java/org/asynchttpclient/oauth/OAuthSignatureCalculatorTest.java
@@ -47,7 +47,7 @@ public class OAuthSignatureCalculatorTest {
 
     public static final String TOKEN_SECRET = "pfkkdhi9sl3r4s00";
 
-    public static final String NONCE = "kllo9940pd9333jh";
+    public static final String NONCE = "9940pd/933+3jh==";
 
     final static long TIMESTAMP = 1191242096;
 
@@ -83,10 +83,11 @@ public class OAuthSignatureCalculatorTest {
         List<Param> queryParams = new ArrayList<>();
         queryParams.add(new Param("file", "vacation.jpg"));
         queryParams.add(new Param("size", "original"));
+        queryParams.add(new Param("csv", "aaa%2Cbbb")); // add percent encoded test case
         String url = "http://photos.example.net/photos";
         String sig = calc.calculateSignature("GET", Uri.create(url), TIMESTAMP, NONCE, null, queryParams);
 
-        assertEquals(sig, "tR3+Ty81lMeYAr/Fid0kMTYa/WM=");
+        assertEquals(sig, "oc/GQouurhlY1gWFyaqz9/w7fAg=");
     }
 
     @Test(groups = "fast")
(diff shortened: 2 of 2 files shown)


slandelle:
@alalwww Could you elaborate on your fix, please? As [per spec](http://oauth.net/core/1.0/#anchor14): 

> All the request parameters MUST be encoded as described in Parameter Encoding prior to constructing the Signature Base String.

alalwww:
@slandelle 

### problem
It will fail OAuth authentication, if has percent encoded parameter.

### cause
The request parameters is percent-encoded three times to create the OAuth signature if when [`RequestBuilder#uriEncoder`](https://github.com/AsyncHttpClient/async-http-client/blob/master/api/src/main/java/org/asynchttpclient/RequestBuilderBase.java#L294) is `UriEncoder.FIXING`.

1. [1st percent encoding](https://github.com/AsyncHttpClient/async-http-client/blob/master/api/src/main/java/org/asynchttpclient/RequestBuilderBase.java#L655)
(and [`Request#getQueryParams()`](https://github.com/AsyncHttpClient/async-http-client/blob/master/api/src/main/java/org/asynchttpclient/RequestBuilderBase.java#L246-L261))
1. [2nd percent encoding](https://github.com/AsyncHttpClient/async-http-client/blob/master/api/src/main/java/org/asynchttpclient/oauth/OAuthSignatureCalculator.java#L233)
1. [3rd percent encoding](https://github.com/AsyncHttpClient/async-http-client/blob/master/api/src/main/java/org/asynchttpclient/oauth/OAuthSignatureCalculator.java#L171)

### my fix
My fix is anytime decode to `formParams` and `queryParams`, because calculator can't judgement those `Param`s has been percent-encoded.

I'm not good at English sorry. :sob:
thanks.

slandelle:
The [spec](http://oauth.net/core/1.0/#anchor14) is a bit confusing, but I tend to agree with @cowtowncoder's reading (the original author).

*From "9.1.1.  Normalize Request Parameters":*

> a=1&c=hi%20there&f=25&f=50&f=a&z=p&z=t

So query parameters are supposed to be encoded here (they are in AHC, from `request.getQueryParams()`).

*From "9.1.3.  Concatenate Request Elements":*

> Each item is encoded and separated by an ‘&’ character

So it looks to me that encoding is indeed supposed to happen twice, as weird as it seems.

So now:
* Do you have any reference that would explain why our reading is wrong and that encoding should only once for request parameters?
* If our reading is right, then the issue would be on the service provider side. Which one do you use?

Cheers

PS: Your English is fine, thanks for your help on this. :smile:

alalwww:
I tried OAuth to Twitter on the PlayFramework 2.3 and 2.4.
Play 2.3 can be authenticated, failed in 2.4.

Play 2.4's OAuth has been changed from Signpost to AHC's calculator. [[link](https://www.playframework.com/documentation/2.4.x/Migration24#WS-client)]
> WS has changed the OAuth signature calculator from Signpost to AsyncHttpClient’s OAuthCalculator. 

I found this problem when I investigating the cause of this authentication failure.

---
I think that this problem did not occur when quondam calculator has been added.
The query parameters has not been percent-encoded, [when the request builder calling calculator]
(https://github.com/AsyncHttpClient/async-http-client/blob/1.7.x/src/main/java/com/ning/http/client/AsyncHttpClient.java#L271-L284).

thanks!

slandelle:
@alalwww I'm very sorry it took me so long to figure out what you meant (my fault, your explanations were clear).
So you're absolutely right: we're encoding query params twice. As this only happens with query params, I'm going to fix this in a different way than the one you're suggesting.

Thanks a lot!

slandelle:
Ping @jroper @wsargent

alalwww:
@slandelle 
OK. Certainly my fix is not appropriate.

I think so too that [fix](https://github.com/AsyncHttpClient/async-http-client/commit/f01d8610b9ceebc1de59d42f569b8af3efbe0a0f) is that it most simple If case you do not care about the change `RequestBuilderBase#uriEncoder` case and use `RequestBuilderBase#(add|set)(Form|Query)Param(s)` methd with encoded `Param(s)` case.
maybe, but it's an irregular case. :)

Thanks a lot!! :star2:

wsargent:
@slandelle is there a test case we can use?

slandelle:
@wsargent See #921

PR #915, opened 2015-06-14
fix unnecessary percent encoding(backport)

alalwww:
backport #914


slandelle:
Fixing a bit differently, but thanks!

PR #916, opened 2015-06-16
"followRedirect" feature doesn't work in this condition

joymufeng:
"followRedirect" feature doesn't work in this condition:
GET /checkCookieUrl HTTP/1.1
Host: sample.com

HTTP/1.1 302 Moved Temporarily
Set-Cookie: _customId=qs7e95d91464;domain=sample.com;path=/
Location: http://sample.com/checkCookieUrl

Since we have maxRedirects to prevent dead loops, there is no need to detect if urls equals while redirect occurs.

### DIFF
--- a/providers/netty3/pom.xml
+++ b/providers/netty3/pom.xml
@@ -5,7 +5,7 @@
 		<version>2.0.0-SNAPSHOT</version>
 	</parent>
 	<modelVersion>4.0.0</modelVersion>
-	<artifactId>async-http-client-netty3-provider</artifactId>
+	<artifactId>async-http-client-netty3</artifactId>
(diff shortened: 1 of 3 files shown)


slandelle:
> "followRedirect" feature doesn't work in this condition:

What's the issue exactly? See point below, it's impossible to figure out your changes.

> Since we have maxRedirects to prevent dead loops, there is no need to detect if urls equals while redirect occurs.

Point taken, but that's a different one. Non atomic PR are impossible to review (just look at the diff). Could you please revert it?

Thanks

slandelle:
Maybe I didn't get it right (once again, the diff is impossible to review). Was removing the "same url" test the only thing you changed?

joymufeng:
Sorry, I will revert it soon.

PR #917, opened 2015-06-16
Remove the checking of "redirecting to the same url".

joymufeng:
Fix up for the following condition:
---Request---
GET /checkCookieUrl HTTP/1.1
Host: sample.com
---Response---
HTTP/1.1 302 Moved Temporarily
Set-Cookie: _customId=qs7e95d91464;domain=sample.com;path=/
Location: http://sample.com/checkCookieUrl


slandelle:
OK, so the only change was indeed just removing the "same url" test :)

Thanks!

joymufeng:
@slandelle Sorry, I just comment the "if" statements. I'm new to github and found that if I re-format the code, the diff result was in a mess.:)

slandelle:
No problem. Thanks for contributing!

Issue #922, opened 2015-06-23
Fix for #921 breaks all OAuth requests


henningp:
The fix for issue #921 breaks all OAuth requests, even those that were previously working.

The problem is that the nonce is not urlencoded, breaking the request (e.g. "&oauth_nonce=cDEwXIeEzLcTYN/tSKZOCg==").

slandelle:
@henningp Ouch. In the samples I got, the nonce token didn't need encoding... Fixed, thanks for reporting!

henningp:
Added pull request: https://github.com/AsyncHttpClient/async-http-client/pull/923

henningp:
Ah, that was fast! :) Looking forward to the release, we're running a monkeypatched version as of now.

slandelle:
@henningp I just released 1.9.29.

henningp:
Nice, thanks!

PR #923, opened 2015-06-23
Fix for #922: URL-encoding nonce, fixing OAuth.

henningp:
Urlencoding nonce, adapting test case.


slandelle:
I didn't know you would provide a fix and already pushed it.
Regarding the test, I'd rather not change this one as it's the reference one from the RFC. I'd rather add a new one.

PR #931, opened 2015-07-06
Cast correct object to WebSocketUpgradeHandler

gulrotkake:
The NettyResponseFuture can not be cast to a WebSocketUpgradeHandler, resulting in a cast class exception and the close-callback never being called. Calling getAsyncHandler() on the future and casting that results in the correct behaviour.


slandelle:
Thanks!

slandelle:
Note: Only Netty 4 provider was buggy, the Netty 3 one was fine.

PR #933, opened 2015-07-13
AsyncHttpProviderUtils  method parseCharset parse header charset with…

wagzhi:
Some server (Wechat public api) response header is "encoding=utf-8", but not "charset=utf-8"


slandelle:
@wagzhi Can you provide some reference stating that using `encoding` instead of `charset` is legit, please?

If it's not legit, this is a bug that should be reported against Wechat. I'm really not in favor of introducing workarounds for every wrong implementation one could find in the wild.

wagzhi:
@slandelle You are right!  it's not legit. 
I'll report it to Tencent. Thanks!

slandelle:
Thanks you for your feedback. :)

PR #946, opened 2015-08-13
Send END_PADDING to server

bomgar:
In the finished state -1 is returned. So the caller assumes that nothing was written to the outgoing buffer
and misses the end padding.


slandelle:
Thanks!

PR #947, opened 2015-08-14
Fix stream suspending with FeedableBodyGenerator.

bomgar:
netty expects null as outgoing buffer to suspend the stream.

see ChunkedWriteHandler#222


bomgar:
I'm not sure if the instanceof is necessary. I added it because i don't want to cause NPE for other callers.

bomgar:
I backported it to 1.9. But it still doesn't work.

PR #948, opened 2015-08-14
Fix stream suspending with FeedableBodyGenerator.

bomgar:
netty expects null as outgoing buffer to suspend the stream.

see ChunkedWriteHandler#222

Also empty buffers need to be ignored to avoid suspending the stream unnecessarily.

### DIFF
--- a/api/src/main/java/org/asynchttpclient/request/body/generator/FeedableBodyGenerator.java
+++ b/api/src/main/java/org/asynchttpclient/request/body/generator/FeedableBodyGenerator.java
@@ -32,6 +32,8 @@ public final class FeedableBodyGenerator implements BodyGenerator {
     private final Queue<BodyPart> queue = new ConcurrentLinkedQueue<>();
     private FeedListener listener;
 
+    private boolean writeChunkBoundaries = true;
+
     @Override
     public Body createBody() {
         return new PushBody();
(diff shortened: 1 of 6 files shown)


bomgar:
I have backported the code to 1.9 and tested it with netty3. I'm not sure if the instanceof is necessary but i want to avoid breaking other callers with NPEs.

slandelle:
Thanks, but would it be possible for you to add some tests, please?

bomgar:
I added some test for the feedable body generator.

bomgar:
I'm actually wondering how this thing was tested before. I had a lot of trouble analyzing the errors. I have no idea if this is compatible with netty 4.

slandelle:
Currently, there's no test for FeedableBodyGenerator. But it would be great to have one. Please have a look at our test suite: you can start a Jetty server, and configure how it replies (eg echo).

bomgar:
The unit test in my second commit should cover the body generator. Unfortunately the BodyChunkedInput is still lacking one. 

Without this change the netty thread did non stop call read on the body. I will see what I can do to make a better test. But the change is necessary.

bomgar:
I have a test now with netty 3 and 4. With netty 3 everything seems to work. With netty 4 the server replies then chunk boundaries. Any ideas?

bomgar:
Now it seems to work with both version. I copied the netty 3 issue hack from the InputStreamBodyGenerator.

slandelle:
@bomgar LGTM, thanks a lot! I'll probably change a few bits, like making the default behavior target Netty 4 and not old Netty 3.

PR #949, opened 2015-08-16
Rework FeedableBodyGerator

bomgar:
the read method became really big.

### DIFF
--- a/api/src/main/java/org/asynchttpclient/request/body/generator/FeedableBodyGenerator.java
+++ b/api/src/main/java/org/asynchttpclient/request/body/generator/FeedableBodyGenerator.java
@@ -29,6 +29,7 @@ import org.asynchttpclient.request.body.Body;
 public final class FeedableBodyGenerator implements BodyGenerator {
     private final static byte[] END_PADDING = "\r\n".getBytes(US_ASCII);
     private final static byte[] ZERO = "0".getBytes(US_ASCII);
+    private final static ByteBuffer EMPTY_BUFFER = ByteBuffer.allocate(0);
     private final Queue<BodyPart> queue = new ConcurrentLinkedQueue<>();
     private FeedListener listener;
 
@@ -60,7 +61,7 @@ public final class FeedableBodyGenerator implements BodyGenerator {
     }
 
     private enum PushBodyState {
-        ONGOING, CLOSING, FINISHED;
+        ONGOING, FINISHED;
     }
     
     public final class PushBody implements Body {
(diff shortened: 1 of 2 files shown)


slandelle:
I think there's a bug in `BodyChunkedInput` because of the way the ByteBuffer is computed.
If body.contentLength is 10 or less, we end up suspending or finishing because this code in `readBodyPart`:

```java
int capacity = buffer.remaining() - 10; // be safe (we'll have to add size, ending, etc.)
int size = Math.min(nextPart.buffer.remaining(), capacity);
```

WDYT?

bomgar:
At the moment the whole thing relies on the outgoing buffer being big enough to hold the boundaries and at least one byte. The "closing" part has the same problem.

bomgar:
I rewrote parts of it. What do you think?

api/src/main/java/org/asynchttpclient/request/body/generator/FeedableBodyGenerator.java
```diff
@@ -74,51 +75,47 @@ public long getContentLength() {
...
+                if (nextPart == null) {
+                    // Nothing in the queue. suspend stream if nothing was read. (reads == 0)
+                    return reads;
+                } else if (!nextPart.buffer.hasRemaining() && !nextPart.isLast) {
+                    // skip empty buffers
+                    // if we return 0 here it would suspend the stream - we don't want that
+                    queue.remove();
+                } else {
+                    readBodyPart(buffer, nextPart);
+                    reads++;
```
slandelle: This looks wrong, should probably be `reads += readBodyPart(buffer, nextPart);`

slandelle: Looks wrong to me: `read` must return the number of bytes read, while this seems to be returning the number of parts read (and still assuming buffer is large enough).

api/src/main/java/org/asynchttpclient/request/body/generator/FeedableBodyGenerator.java
```diff
@@ -127,13 +124,56 @@ public void close() {
 
     }
 
-    private final static class BodyPart {
+
+    private void move(ByteBuffer destination, ByteBuffer source) {
```
slandelle: This can be optimized when source has an array.

api/src/main/java/org/asynchttpclient/request/body/generator/FeedableBodyGenerator.java
```diff
@@ -74,51 +75,47 @@ public long getContentLength() {
...
+                    readBodyPart(buffer, nextPart);
+                    reads++;
                 }
-                if (writeChunkBoundaries)
-                    buffer.put(END_PADDING);
             }
-            if (!nextPart.buffer.hasRemaining()) {
-                if (nextPart.isLast) {
-                    state = writeChunkBoundaries ? PushBodyState.CLOSING : PushBodyState.FINISHED;
+            return reads;
```
slandelle: I guess it could just return the difference between the buffer current and original positions.

slandelle:
I've added some comments.
Then, I'm still concerned with buffer overflow (I didn't implement this code) that would happen if a huge pile of parts arrive at the same time.

api/src/main/java/org/asynchttpclient/request/body/generator/FeedableBodyGenerator.java
```diff
@@ -127,13 +124,56 @@ public void close() {
 
     }
 
-    private final static class BodyPart {
+
+    private void move(ByteBuffer destination, ByteBuffer source) {
+        while(destination.hasRemaining() && source.hasRemaining()) {
```
slandelle: Why not put a whole slice?

bomgar:
It doesn't have to return the exact amount of bytes. A positive number is enough. The amount is represented in the ByteBuffer. 
To solve the overflow problem one would have to implement backpressure. That would be great to have but this should not be part of this refactoring.

bomgar:
The code also does not assume the outgoing buffer is large enough. It stops writing as soon as it is full and continues with the next buffer at the exact same position. That's why i initialize the buffers in the BodyPart.

Also the loop is fine. The copy method from one buffer to another does exactly the same except it checks the destination buffer is large enough (not what i want).

slandelle:
> It doesn't have to return the exact amount of bytes. A positive number is enough.

Oh, right. Then, as AHC2 is still in alpha, I think we should change `Body.read` signature to return a State enum: `CONTINUE, SUSPEND, FINISHED`. WDYT?

slandelle:
> To solve the overflow problem one would have to implement backpressure. That would be great to have but this should not be part of this refactoring.

Ultimately, yes (ping @jroper).

bomgar:
> Oh, right. Then, as AHC2 is still in alpha, I think we should change Body.read signature to return a State enum: CONTINUE, SUSPEND, FINISHED. WDYT?

Sounds good to me.

slandelle:
Merging, I'll take care of some clean up. Thanks!

slandelle:
Optimized move in af0df14e63198bb90ebccf5961e6799d77109970

api/src/main/java/org/asynchttpclient/request/body/generator/FeedableBodyGenerator.java
slandelle: Done in af0df14e63198bb90ebccf5961e6799d77109970

PR #954, opened 2015-08-22
AsyncHttpClient `name` option

### LINKED ISSUE #952: Option to specify http client instance name
Requesting a `String` option for http client instance name.

(a field in `AsyncHttpClientConfig.Builder`)

It has two applications: for thread names, and for debugging.

## Thread name

Currently I have hunders of threads like "New I/O worker 120". That's because I have several async http client instances, and "New I/O worker" is a default thread name for Netty.

Requesting async http client simple `String` option: `name`, which is a name of http client instance.

And that name (if specified) should be used by AsyncHttpProvider implementations as a prefix for thread names/pool names.

It is currently possible to specify thread names by manually supplying AsyncHttpProviderConfig with thread factories, but that's too much work for the most use cases.

## Debugging

String field in http client would be helpful to distinguish between multiple instance of http client when setting a breakpoint inside AsyncHttpClient.

stepancheg:
Fixes #952

Name option is useful for two purposes:

## Thread name

Typical application with several netty and/or async-http-client
instances may have hundreds of threads like "New I/O worker 120",
which makes it hard to understand which thread belongs to which
part of the application.

Specified name is used in AsyncHttpClient to name threads created
by AsyncHttpClient and netty.

If name is not specified, `AsyncHttpClient` string is used.

## Debugging

`name` is stored in `AsyncHttpClientConfig`, and available in
debugger debugging when application has multiple client instances.
It can be useful to distinguish between instance, for example, when
breakpoint is set inside of AsyncHttpClient.

### DIFF
--- a/api/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
+++ b/api/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
@@ -34,6 +34,7 @@ import org.asynchttpclient.filter.RequestFilter;
 import org.asynchttpclient.filter.ResponseFilter;
 import org.asynchttpclient.proxy.ProxyServer;
 import org.asynchttpclient.proxy.ProxyServerSelector;
+import org.asynchttpclient.util.PrefixIncrementThreadFactory;
 import org.asynchttpclient.util.ProxyUtils;
 
 /**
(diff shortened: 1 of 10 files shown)


slandelle:
LGTM, thanks!

Issue #953, opened 2015-08-22
Add Rx Support


drmaas:
Add optional Rx support to the AsyncHttpClient, e.g.
```
...
AsyncHttpClient asyncHttpClient = new AsyncHttpClient();
Observable<Response> observable = asyncHttpClient.prepareGet("http://www.ning.com/").observe();
observable.subscribe((response) -> {
  System.out.println(response.getResponseBody());
});
...
```
The RxJava dependency could be optional, such that consumers would provide it only if they desire to use it. 

Another option would be to create a new client class that could inherit most behavior from the existing client, but provide the additional Rx support.

slandelle:
Is this something you plan on contributing?

drmaas:
I certainly can start on it. I have a simple version running locally. I will need to flesh out the design and add tests.

slandelle:
Yep. We need a new adapter module so that core API doesn't depend on Rx.

slandelle:
Closed by #961

PR #961, opened 2015-08-28
#953 - Adding extra support for rxjava

drmaas:


jfarcand:
@drmaas Can you add the proper license in your Java classes?

extras/rxjava/src/main/java/org/asynchttpclient/extras/rxjava/AsyncHttpClientErrorException.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2015 Target, Inc. All rights reserved.
```
slandelle: Sorry, we can't accept PRs where code isn't donated to "The AsyncHttpClient Project".

extras/rxjava/src/main/java/org/asynchttpclient/extras/rxjava/AsyncHttpObservable.java
```diff
@@ -0,0 +1,123 @@
...
+            @Override
+            public void call(final Subscriber<? super Response> subscriber) {
+                try {
+                    AsyncCompletionHandler<Void> handler = new AsyncCompletionHandler<Void>() {
+                        @Override
+                        public State onStatusReceived(HttpResponseStatus status) throws Exception {
+                            State state = super.onStatusReceived(status);
+                            if (abortOnErrorStatus) {
+                                int code = status.getStatusCode();
+                                if (code >= 400) {
```
slandelle: I think this is for the subscriber to decide which status code is an error. From the HTTP standpoint, there wasn't any error and a proper response was received.

extras/rxjava/src/main/java/org/asynchttpclient/extras/rxjava/AsyncHttpClientErrorException.java
```diff
@@ -0,0 +1,27 @@
...
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+package org.asynchttpclient.extras.rxjava;
+
+/**
+ *
+ */
+public class AsyncHttpClientErrorException extends RuntimeException {
```
slandelle: No need for this Exception, as below comment.

extras/rxjava/src/main/java/org/asynchttpclient/extras/rxjava/AsyncHttpClientErrorException.java
drmaas: I can change that, but can you clarify the policy - many other files have copyright headers for corporations e.g. Sonatype, Ning, Google etc.

extras/rxjava/src/main/java/org/asynchttpclient/extras/rxjava/AsyncHttpObservable.java
drmaas: The goal is to allow processing to abort on failed status codes. Would it be better to pass in a set of success status codes instead of a Boolean? 

Another idea is to allow someone to pass in their own handler. That way they would still be shielded from observable creation.

slandelle: > The goal is to allow processing to abort on failed status codes.

Depending on the use case, some status codes <= 400 could also be failures.

> Another idea is to allow someone to pass in their own handler.

Well, a Subscriber is a handler, just a different API.

drmaas: Once you subscribe to a response, you will be forced to read the entire response body. Unless there's something I'm not considering?

It almost seems that configuring status codes on which to abort should be the domain of either the client or the builder configuration, not the observable.

slandelle: If you're looking for full control, are you sure you don't want https://github.com/jroper/async-http-client/commit/1da72d579996d9529b9abdb3da4bc5530ca9cf45 that's about to be contributed, and that will bring support for reactive streams (that RxJava implements)?

slandelle: And then, do you feel like you'll really save a lot by not reading the body for status codes >= 400? I mean that bodies for such status codes are typically less that 1 MTU and will be read along the status line and headers.

drmaas:
I removed the abort logic - I think doing something like what is being done with reactive streams is possible but would tie you to rxjava and is not really what I had in mind for this request.

slandelle:
Great, thanks!

I'll release alpha9 once the reactive-streams support gets merged too.

PR #963, opened 2015-09-01
Issue/544 reactive streams support

### LINKED ISSUE #544: Backpressure in AsyncHandler
AsyncHandler provides no mechanism to send back pressure on receiving the body parts.

Imagine you have a server that stores large files on Amazon S3, and streams them out to clients, using async http client to connect to S3.  Now imagine you have a very slow client, that connects and downloads a file.  The slow client pushes back on the server via TCP.  However, async http client will keep on calling onBodyPartReceived as fast as S3 provides it with data.  The AsyncHandler implementation will have three choices:

1. Block. Then it's blocking a worker thread, preventing other concurrent operations from happening.  This is not an option.
2. Buffer. Eventually this will cause an OutOfMemoryError.  This is not an option.
3. Drop. Then the client gets a corrupted file.  This is not an option.

AsyncHandler therefore needs a mechanism to propagate back pressure when handling body parts.  One possibility here is to provide a method to say whether you are interested in receiving more data or not.  This would correspond to a Channel.setReadable(true/false) in the netty provider, which will push back via TCP flow control.  This could either be provided by injecting some sort of "channel" object into the AsyncHandler, or, since HttpResponseBodyPart already provides mechanisms for talking back to the channel (eg closeUnderlyingConnection()), it could be provided there.

dotta:
Here is an initial proposal to fix #544.

I'm entirely new to both Netty and AHC, so if you notice anything weird it's very likely that I haven't understood something :-)

~~Other than that, something is puzzling me with the tests I wrote. Basically, if I change the `CHUNK_SIZE` in https://github.com/AsyncHttpClient/async-http-client/compare/AsyncHttpClient:master...dotta:issue/544-reactive-streams-support?expand=1#diff-5133f4e3fed98f7a99081a1ae643c4c0R104 to a value around `20000` (or lower), the tests start to fail because there is a mismatch between the received bytes and the expected ones. I honestly have no good explanation for this, but I thought it may have something to do with the internals of Netty (or maybe a config flag?). I'd really appreciate it if someone more experienced with Netty could have a look.~~
EDIT: This issue is now fixed after using RxJava instead of providing my own implementation of `Publisher` in the test :-)

This PR also includes @jroper commit to add support for handling response bodies.

### DIFF
--- a/api/src/main/java/org/asynchttpclient/handler/StreamedAsyncHandler.java
+++ b/api/src/main/java/org/asynchttpclient/handler/StreamedAsyncHandler.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2015 AsyncHttpClient Project. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+package org.asynchttpclient.handler;
+
+import org.asynchttpclient.AsyncHandler;
+import org.asynchttpclient.HttpResponseBodyPart;
+import org.reactivestreams.Publisher;
+
+/**
+ * AsyncHandler that uses reactive streams to handle the request.
+ */
+public interface StreamedAsyncHandler<T> extends AsyncHandler<T> {
+
+    /**
+     * Called when the body is received. May not be called if there's no body.
+     *
+     * @param publisher The publisher of response body parts.
+     * @return Whether to continue or abort.
+     */
+    State onStream(Publisher<HttpResponseBodyPart> publisher);
+}
--- a/api/src/test/java/org/asynchttpclient/reactivestreams/ReactiveStreamsTest.java
+++ b/api/src/test/java/org/asynchttpclient/reactivestreams/ReactiveStreamsTest.java
@@ -1,0 +1,263 @@
+/*
+ * Copyright (c) 2015 AsyncHttpClient Project. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+package org.asynchttpclient.reactivestreams;
+
+import static org.asynchttpclient.test.TestUtils.LARGE_IMAGE_BYTES;
+import static org.testng.Assert.assertEquals;
+
+import java.io.ByteArrayOutputStream;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+
+import org.asynchttpclient.AbstractBasicTest;
+import org.asynchttpclient.AsyncHttpClient;
+import org.asynchttpclient.HttpResponseBodyPart;
+import org.asynchttpclient.HttpResponseHeaders;
+import org.asynchttpclient.HttpResponseStatus;
+import org.asynchttpclient.ListenableFuture;
+import org.asynchttpclient.handler.StreamedAsyncHandler;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import org.testng.annotations.Test;
+
+public abstract class ReactiveStreamsTest extends AbstractBasicTest {
+
+    @Test(groups = { "standalone", "default_provider" })
+    public void streamedResponseTest() throws Throwable {
+        try (AsyncHttpClient c = getAsyncHttpClient(null)) {
+
+            ListenableFuture<SimpleStreamedAsyncHandler> future = c.preparePost(getTargetUrl())
+                    .setBody(LARGE_IMAGE_BYTES)
+                    .execute(new SimpleStreamedAsyncHandler());
+
+            assertEquals(future.get().getBytes(), LARGE_IMAGE_BYTES);
+
+            // Run it again to check that the pipeline is in a good state
+            future = c.preparePost(getTargetUrl())
+                    .setBody(LARGE_IMAGE_BYTES)
+                    .execute(new SimpleStreamedAsyncHandler());
+
+            assertEquals(future.get().getBytes(), LARGE_IMAGE_BYTES);
+
+            // Make sure a regular request still works
+            assertEquals(c.preparePost(getTargetUrl())
+                    .setBody("Hello")
+                    .execute().get().getResponseBody(), "Hello");
+
+        }
+    }
+
+    @Test(groups = { "standalone", "default_provider" })
+    public void cancelStreamedResponseTest() throws Throwable {
+        try (AsyncHttpClient c = getAsyncHttpClient(null)) {
+
+            // Cancel immediately
+            c.preparePost(getTargetUrl())
+                    .setBody(LARGE_IMAGE_BYTES)
+                    .execute(new CancellingStreamedAsyncProvider(0)).get();
+
+            // Cancel after 1 element
+            c.preparePost(getTargetUrl())
+                    .setBody(LARGE_IMAGE_BYTES)
+                    .execute(new CancellingStreamedAsyncProvider(1)).get();
+
+            // Cancel after 10 elements
+            c.preparePost(getTargetUrl())
+                    .setBody(LARGE_IMAGE_BYTES)
+                    .execute(new CancellingStreamedAsyncProvider(10)).get();
+
+            // Make sure a regular request works
+            assertEquals(c.preparePost(getTargetUrl())
+                    .setBody("Hello")
+                    .execute().get().getResponseBody(), "Hello");
+
+        }
+    }
+
+    static protected class SimpleStreamedAsyncHandler implements StreamedAsyncHandler<SimpleStreamedAsyncHandler>{
+        private final SimpleSubscriber<HttpResponseBodyPart> subscriber;
+
+        public SimpleStreamedAsyncHandler() {
+            this(new SimpleSubscriber<HttpResponseBodyPart>());
+        }
+
+        public SimpleStreamedAsyncHandler(SimpleSubscriber<HttpResponseBodyPart> subscriber) {
+            this.subscriber = subscriber;
+        }
+        @Override
+        public State onStream(Publisher<HttpResponseBodyPart> publisher) {
+            publisher.subscribe(subscriber);
+            return State.CONTINUE;
+        }
+
+        @Override
+        public void onThrowable(Throwable t) {
+            throw new AssertionError(t);
+        }
+
+        @Override
+        public State onBodyPartReceived(HttpResponseBodyPart bodyPart) throws Exception {
+            throw new AssertionError("Should not have received body part");
+        }
+
+        @Override
+        public State onStatusReceived(HttpResponseStatus responseStatus) throws Exception {
+            return State.CONTINUE;
+        }
+
+        @Override
+        public State onHeadersReceived(HttpResponseHeaders headers) throws Exception {
+            return State.CONTINUE;
+        }
+
+        @Override
+        public SimpleStreamedAsyncHandler onCompleted() throws Exception {
+            return this;
+        }
+
+        public byte[] getBytes() throws Throwable {
+            List<HttpResponseBodyPart> bodyParts = subscriber.getElements();
+            ByteArrayOutputStream bytes = new ByteArrayOutputStream();
+            for (HttpResponseBodyPart part : bodyParts) {
+                part.writeTo(bytes);
+            }
+            return bytes.toByteArray();
+        }
+    }
+
+    /**
+     * Simple subscriber that requests and buffers one element at a time.
+     */
+    static protected class SimpleSubscriber<T> implements Subscriber<T> {
+        private volatile Subscription subscription;
+        private volatile Throwable error;
+        private final List<T> elements = Collections.synchronizedList(new ArrayList<T>());
+        private final CountDownLatch latch = new CountDownLatch(1);
+
+        @Override
+        public void onSubscribe(Subscription subscription) {
+            this.subscription = subscription;
+            subscription.request(1);
+        }
+
+        @Override
+        public void onNext(T t) {
+            elements.add(t);
+            subscription.request(1);
+        }
+
+        @Override
+        public void onError(Throwable error) {
+            this.error = error;
+            latch.countDown();
+        }
+
+        @Override
+        public void onComplete() {
+            latch.countDown();
+        }
+
+        public List<T> getElements() throws Throwable {
+            latch.await();
+            if (error != null) {
+                throw error;
+            } else {
+                return elements;
+            }
+        }
+    }
+
+    static class CancellingStreamedAsyncProvider implements StreamedAsyncHandler<CancellingStreamedAsyncProvider> {
+        private final int cancelAfter;
+
+        public CancellingStreamedAsyncProvider(int cancelAfter) {
+            this.cancelAfter = cancelAfter;
+        }
+
+        @Override
+        public State onStream(Publisher<HttpResponseBodyPart> publisher) {
+            publisher.subscribe(new CancellingSubscriber<HttpResponseBodyPart>(cancelAfter));
+            return State.CONTINUE;
+        }
+
+        @Override
+        public void onThrowable(Throwable t) {
+            throw new AssertionError(t);
+        }
+
+        @Override
+        public State onBodyPartReceived(HttpResponseBodyPart bodyPart) throws Exception {
+            throw new AssertionError("Should not have received body part");
+        }
+
+        @Override
+        public State onStatusReceived(HttpResponseStatus responseStatus) throws Exception {
+            return State.CONTINUE;
+        }
+
+        @Override
+        public State onHeadersReceived(HttpResponseHeaders headers) throws Exception {
+            return State.CONTINUE;
+        }
+
+        @Override
+        public CancellingStreamedAsyncProvider onCompleted() throws Exception {
+            return this;
+        }
+    }
+
+    /**
+     * Simple subscriber that cancels after receiving n elements.
+     */
+    static class CancellingSubscriber<T> implements Subscriber<T> {
+        private final int cancelAfter;
+
+        public CancellingSubscriber(int cancelAfter) {
+            this.cancelAfter = cancelAfter;
+        }
+
+        private volatile Subscription subscription;
+        private volatile int count;
+
+        @Override
+        public void onSubscribe(Subscription subscription) {
+            this.subscription = subscription;
+            if (cancelAfter == 0) {
+                subscription.cancel();
+            } else {
+                subscription.request(1);
+            }
+        }
+
+        @Override
+        public void onNext(T t) {
+            count++;
+            if (count == cancelAfter) {
+                subscription.cancel();
+            } else {
+                subscription.request(1);
+            }
+        }
+
+        @Override
+        public void onError(Throwable error) {
+        }
+
+        @Override
+        public void onComplete() {
+        }
+    }
+}
--- a/api/src/test/java/org/asynchttpclient/request/body/ReactiveStreamsTest.java
+++ b/api/src/test/java/org/asynchttpclient/request/body/ReactiveStreamsTest.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2015 AsyncHttpClient Project. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+package org.asynchttpclient.request.body;
+
+import static org.asynchttpclient.test.TestUtils.LARGE_IMAGE_BYTES;
+import static org.asynchttpclient.test.TestUtils.LARGE_IMAGE_PUBLISHER;
+import static org.testng.Assert.assertEquals;
+
+import java.nio.ByteBuffer;
+import java.util.concurrent.ExecutionException;
+
+import org.asynchttpclient.AbstractBasicTest;
+import org.asynchttpclient.AsyncHttpClient;
+import org.asynchttpclient.AsyncHttpClientConfig;
+import org.asynchttpclient.BoundRequestBuilder;
+import org.asynchttpclient.Response;
+import org.reactivestreams.Publisher;
+import org.testng.annotations.Test;
+
+import rx.Observable;
+import rx.RxReactiveStreams;
+
+public abstract class ReactiveStreamsTest extends AbstractBasicTest {
+
+    @Test(groups = { "standalone", "default_provider" }, enabled = true)
+    public void testStreamingPutImage() throws Exception {
+        try (AsyncHttpClient client = getAsyncHttpClient(new AsyncHttpClientConfig.Builder().setRequestTimeout(100 * 6000).build())) {
+            Response response = client.preparePut(getTargetUrl()).setBody(LARGE_IMAGE_PUBLISHER).execute().get();
+            assertEquals(response.getStatusCode(), 200);
+            assertEquals(response.getResponseBodyAsBytes(), LARGE_IMAGE_BYTES);
+        }
+    }
+
+    @Test(groups = { "standalone", "default_provider" }, enabled = true)
+    public void testConnectionDoesNotGetClosed() throws Exception { // test that we can stream the same request multiple times
+        try (AsyncHttpClient client = getAsyncHttpClient(new AsyncHttpClientConfig.Builder().setRequestTimeout(100 * 6000).build())) {
+            BoundRequestBuilder requestBuilder = client.preparePut(getTargetUrl()).setBody(LARGE_IMAGE_PUBLISHER);
+            Response response = requestBuilder.execute().get();
+            assertEquals(response.getStatusCode(), 200);
+            assertEquals(response.getResponseBodyAsBytes(), LARGE_IMAGE_BYTES);
+            
+            response = requestBuilder.execute().get();
+            assertEquals(response.getStatusCode(), 200);
+            assertEquals(response.getResponseBodyAsBytes(), LARGE_IMAGE_BYTES);
+        }
+    }
+
+    @Test(groups = { "standalone", "default_provider" }, enabled = true, expectedExceptions = ExecutionException.class)
+    public void testFailingStream() throws Exception {
+        try (AsyncHttpClient client = getAsyncHttpClient(new AsyncHttpClientConfig.Builder().setRequestTimeout(100 * 6000).build())) {
+            Observable<ByteBuffer> failingObservable = Observable.error(new FailedStream());
+            Publisher<ByteBuffer> failingPublisher = RxReactiveStreams.toPublisher(failingObservable);
+
+            client.preparePut(getTargetUrl()).setBody(failingPublisher).execute().get();
+        }
+    }
+
+    @SuppressWarnings("serial")
+    private class FailedStream extends RuntimeException {}
+}
--- a/api/src/test/java/org/asynchttpclient/test/TestUtils.java
+++ b/api/src/test/java/org/asynchttpclient/test/TestUtils.java
@@ -20,6 +20,10 @@ import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.SslConnectionFactory;
 import org.eclipse.jetty.util.security.Constraint;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.reactivestreams.Publisher;
+
+import rx.Observable;
+import rx.RxReactiveStreams;
 
 import javax.net.ssl.*;
 
@@ -32,12 +36,15 @@ import java.net.ServerSocket;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.security.*;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
@@ -52,7 +59,8 @@ public class TestUtils {
     private static final File TMP_DIR = new File(System.getProperty("java.io.tmpdir"), "ahc-tests-" + UUID.randomUUID().toString().substring(0, 8));
     public static final byte[] PATTERN_BYTES = "FooBarBazQixFooBarBazQixFooBarBazQixFooBarBazQixFooBarBazQixFooBarBazQix".getBytes(Charset.forName("UTF-16"));
     public static final File LARGE_IMAGE_FILE;
-    public static byte[] LARGE_IMAGE_BYTES;
+    public static final byte[] LARGE_IMAGE_BYTES;
+    public static final Publisher<ByteBuffer> LARGE_IMAGE_PUBLISHER;
     public static final File SIMPLE_TEXT_FILE;
     public static final String SIMPLE_TEXT_FILE_STRING;
     private static final LoginService LOGIN_SERVICE = new HashLoginService("MyRealm", "src/test/resources/realm.properties");
@@ -63,6 +71,7 @@ public class TestUtils {
             TMP_DIR.deleteOnExit();
             LARGE_IMAGE_FILE = new File(TestUtils.class.getClassLoader().getResource("300k.png").toURI());
             LARGE_IMAGE_BYTES = FileUtils.readFileToByteArray(LARGE_IMAGE_FILE);
+            LARGE_IMAGE_PUBLISHER = createPublisher(LARGE_IMAGE_BYTES, /*chunkSize*/ 1000);
             SIMPLE_TEXT_FILE = new File(TestUtils.class.getClassLoader().getResource("SimpleTextFile.txt").toURI());
             SIMPLE_TEXT_FILE_STRING = FileUtils.readFileToString(SIMPLE_TEXT_FILE, UTF_8);
         } catch (Exception e) {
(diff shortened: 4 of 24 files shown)


api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
```diff
@@ -0,0 +1,154 @@
...
+            boolean isCompleted = subscriber.readChunk(buffer);
+            return (isCompleted) ? State.Stop : State.Continue;
+        }
+    }
+
+    private static class QueueBackedSubscriber implements Subscriber<ByteBuffer> {
+
+        private static final Logger LOGGER = LoggerFactory.getLogger(QueueBackedSubscriber.class);
+
+        private final BlockingQueue<ByteBuffer> queue;
```
slandelle: Blocking, really?

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        @Override
+        public long getContentLength() {
+            return -1L;
+        }
+
+        @Override
+        public State read(ByteBuffer buffer) throws IOException {
+            boolean isCompleted = subscriber.readChunk(buffer);
```
slandelle: Correct me if I'm wrong, but your subscriber is going to block if there's nothing in the queue, causing the reading thread to block.

slandelle:
Hi Mirko,

I think your ReactiveBodyGenerator is broken. Are you sure it cannot be implemented on top of `FeedableBodyGenerator`?

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
```diff
@@ -0,0 +1,154 @@
...
+
+                // If we are at the end of the stream, then this condition must be true.
+                // In fact, if we have received an empty chunk, but the stream is not completed, then we handle it 
+                // as any other chunk.
+                if(lastElement == EMPTY && completed.get()) return true;
+
+                // don't overflow the passed buffer
+                if(buffer.remaining() <= lastElement.remaining()) {
+                    int bufferSize = buffer.remaining();
+                    buffer.put(lastElement.array(), lastElement.position(), bufferSize);
```
slandelle: What if the buffer doesn't have an array?

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
```diff
@@ -0,0 +1,154 @@
...
+                }
+                return false;
+            } catch (InterruptedException ie) {
+                LOGGER.debug("Queue was interrupted while waiting for next element.", ie);
+                subscription.cancel();
+                return true; // returning true will provoke the channel handler to stop  
+            }
+        }
+    }
+}
```
slandelle: empty line

api/src/test/java/org/asynchttpclient/reactivestreams/ReactiveStreamsTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+        @Override
+        public void onError(Throwable error) {
+        }
+
+        @Override
+        public void onComplete() {
+        }
+    }
+}
```
slandelle: empty line

api/src/test/java/org/asynchttpclient/request/body/ReactiveStreamsTest.java
```diff
@@ -0,0 +1,50 @@
...
+            Response response = requestBuilder.execute().get();
+            assertEquals(response.getStatusCode(), 200);
+            assertEquals(response.getResponseBodyAsBytes(), LARGE_IMAGE_BYTES);
+            
+            response = requestBuilder.execute().get();
+            assertEquals(response.getStatusCode(), 200);
+            assertEquals(response.getResponseBodyAsBytes(), LARGE_IMAGE_BYTES);
+        }
+    }
+}
```
slandelle: empty line

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
dotta: The alternative is queuing up elements in the subscriber, but the risk with that is overflowing the queue (dropping elements is even worse) when the consumer is much slower than the producer. Also, note this implementation is only used when the provider doesn't support reactive streams.

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
dotta: That is avoided by adding `EMPTY` in the queue on stream completion (see https://github.com/AsyncHttpClient/async-http-client/pull/963/files#diff-2fcc9d1538924b58a722dbc03fe9c2b6R116). I agree this is a bit convoluted, I'll think if there is a better way but I'm open to suggestions.

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
dotta: Oh, is that possible? Any idea how I should handle that? (should I create the array myself, or fail?)

slandelle: > Oh, is that possible?

Of course: DirectBuffers, composites, etc.

> See `FeedableBodyGenerator`, you can put a ByteBuffer into another.

dotta:
Hi @slandelle, thanks a lot for the feedback!

> I think your ReactiveStreamBodyGenerator is broken. Are you sure it cannot be implemented on top of FeedableBodyGenerator?

Do you mean the implementation of `ReactiveStreamBodyGenerator.StreamedBody`? After looking at the `FeedableBodyGenerator.PushBody`, I feel like I should be able to reuse it. I'll have a deeper look now.

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
hepin1989: @dotta the *netty in action* cover this:).composites could have array if it have only one element and the element have array or it was empty.

slandelle:
> Do you mean the implementation of ReactiveStreamBodyGenerator.StreamedBody? After looking at the FeedableBodyGenerator.PushBody, I feel like I should be able to reuse it. I'll have a deeper look now.

Yeah, they serve the same purpose, so I guess existing code could be reuse/extended/wrapped. `FeedableBodyGenerator.PushBody` currently has a known issue that the reading buffer has to be larger than the payload, but that's something that can/should be fixed.

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
slandelle: @hepin1989 Don't confuse Netty's `ByteBuf` and Java's `ByteBuffer` :)

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
slandelle: > Also, note this implementation is only used when the provider doesn't support reactive streams.

I'd rather throw an `UnsupportedOperationException` then. :)

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
hepin1989: @slandelle haha,sorry,I don't see the whole context:).

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
dotta: Mmm, I'm looking at the implementation of `FeedableBodyGenerator` and realise I probably don't need blocking. Forgive me :-)

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
```diff
@@ -0,0 +1,152 @@
...
+
+        @Override
+        public void onNext(ByteBuffer t) {
+            if (t == null) throw null;
+            try {
+                feeder.feed(t, false);
+                executor.execute(new Runnable() {
+                    @Override
+                    public void run() {
+                      subscription.request(1);
```
dotta: @jroper I'm not so convinced about requesting only one element at a time. Any thought?

dotta:
@slandelle I think you should be more pleased now (I for sure am :-)).

api/src/main/java/org/asynchttpclient/request/body/generator/IFeedableBodyGenerator.java
```diff
@@ -0,0 +1,11 @@
+package org.asynchttpclient.request.body.generator;
+
+public interface IFeedableBodyGenerator {
```
slandelle: For consistency, I'd rather remove the `I` prefix and prefix the default implementation with `Default` or `Simple`.

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/HttpProtocol.java
```diff
@@ -400,6 +401,20 @@ private boolean exitAfterHandlingHeaders(Channel channel, NettyResponseFuture<?>
         return false;
     }
 
+    private boolean exitAfterHandlingReactiveStreams(Channel channel, NettyResponseFuture<?> future, HttpResponse response, AsyncHandler<?> handler) throws IOException {
+        if (handler instanceof StreamedAsyncHandler) {
```
slandelle: Anyone missing pattern matching, raise a hand!

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/HttpProtocol.java
```diff
@@ -425,7 +440,8 @@ private boolean handleHttpResponse(final HttpResponse response, final Channel ch
                 exitAfterHandlingRedirect(channel, future, response, request, statusCode, realm) || //
                 exitAfterHandlingConnect(channel, future, request, proxyServer, statusCode, httpRequest) || //
                 exitAfterHandlingStatus(channel, future, response, handler, status) || //
-                exitAfterHandlingHeaders(channel, future, response, handler, responseHeaders);
+                exitAfterHandlingHeaders(channel, future, response, handler, responseHeaders) ||
+                exitAfterHandlingReactiveStreams(channel, future, response, handler);
```
slandelle: Anyone missing monadic composition, raise a hand...

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/Processor.java
```diff
@@ -71,11 +75,41 @@ public void channelRead(final ChannelHandlerContext ctx, Object msg) throws Exce
...
                 Channels.setDiscard(channel);
             }
+            ReferenceCountUtil.release(msg);
+
 
         } else if (attribute instanceof NettyResponseFuture) {
             NettyResponseFuture<?> future = (NettyResponseFuture<?>) attribute;
             protocol.handle(channel, future, msg);
 
+        } else if (attribute instanceof StreamedResponsePublisher) {
```
slandelle: I missed this one. I suspect not having a `NettyResponseFuture` attribute would lead to tons of issues with retry and exception handling.

slandelle: I think the `StreamedResponsePublisher` could be stored in the `NettyResponseFuture`.
Yes, I know it's a huge pile of junk, better design suggestions mostly welcome.

api/src/main/java/org/asynchttpclient/request/body/generator/IFeedableBodyGenerator.java
dotta: :+1:

api/src/main/java/org/asynchttpclient/BoundRequestBuilder.java
```diff
@@ -93,6 +97,16 @@ public BoundRequestBuilder setBody(String data) {
...
 
     @Override
+    public BoundRequestBuilder setBody(Publisher<ByteBuffer> publisher) {
+        return super.setBody(new ReactiveStreamBodyGenerator(client.getConfig().getExecutorService(), publisher));
+    }
+
+    @Override
+    public BoundRequestBuilder setBody(BodyGenerator bodyGenerator) {
+        return super.setBody(bodyGenerator);
+    }
```
jroper: This isn't needed to be overridden since it just delegates straight to super is it?

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
```diff
@@ -0,0 +1,152 @@
...
+              subscription.request(1);
+            }
+        }
+
+        @Override
+        public void onNext(ByteBuffer t) {
+            if (t == null) throw null;
+            try {
+                feeder.feed(t, false);
+                executor.execute(new Runnable() {
```
jroper: This is not necessary, reactive streams spec [3.3](https://github.com/reactive-streams/reactive-streams-jvm/tree/v1.0.0#3-subscription-code) says it's the responsibility of `Subscription.request` to put a bound on the recursion, so it is allowed to call `request` from `onNext` as much as you like.

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
jroper: Considering this is a fallback implementation, I think it's fine.  Another option, seeing as the provider is going to end up buffering infinitely anyway, is you can invoke `subscription.request(Long.MAX_VALUE)` in `onSubscribe`, this is allowed according to reactive streams spec [3.17](https://github.com/reactive-streams/reactive-streams-jvm/tree/v1.0.0#3.17).

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/HttpProtocol.java
jroper: :hand:

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/HttpProtocol.java
jroper: :hand:

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/Processor.java
jroper: If you look in the exception handling below on line 190 (the diff doesn't make it obvious that the code I added is part of the exception handling), I have handled it - the exception gets sent to the stream as an error, and then the future is sent to `requestSender.abort`.

As to whether it should be replayed or not - in Play for example, if it fails during the stream, we never want it to be replayed.  Why?  Because our API essentially looks something like this:

```java
Future<ResponseHeader, Publisher<ByteBuffer>> executeStream();
```

This future will be redeemed when `onStream` is invoked.  All subsequent calls to `onStream` will be ignored - since there is literally nothing we can do with them.  And, to imagine the most common use case for this, where you're streaming a response body out to another response body (acting essentially as a proxy), if the stream fails halfway through, what are you going to do?  You can only forward the failure to the client.  You can't retry, because your opportunity to send a response has gone.  Basically, if you're handling the response as a stream, then this very likely means that consuming the body is side effecting, in which case, it's very unlikely that you want to retry, in fact it could often be dangerous - this is a very different thing to, for example, buffering the body in memory, where there are no negative consequences to retrying.

providers/netty4/src/main/java/org/asynchttpclient/netty/request/body/NettyReactiveStreamBody.java
```diff
@@ -0,0 +1,140 @@
...
+            LOGGER.warn("Stream has already been consumed and cannot be reset");
+        } else {
+            future.setStreamWasAlreadyConsumed(true);
+            NettySubscriber subscriber = new NettySubscriber(channel);
+            channel.pipeline().addLast(NAME_IN_CHANNEL_PIPELINE, subscriber);
+            publisher.subscribe(new SubscriberAdapter(subscriber));
+        }
+    }
+
+    private static class SubscriberAdapter implements Subscriber<ByteBuffer> {
```
jroper: A better "Netty" way to implement this would be to insert another handler in the chain, just before the `HandlerSubscriber` that intercepts calls to `write`, and converts it to a `DefaultHttpContent`.

providers/netty4/src/main/java/org/asynchttpclient/netty/request/body/NettyReactiveStreamBody.java
```diff
@@ -0,0 +1,140 @@
...
+
+        private final Channel channel;
+
+        public NettySubscriber(Channel channel) {
+            super(channel.eventLoop());
+            this.channel = channel;
+        }
+
+        @Override
+        public void onComplete() {
```
jroper: `HandlerSubscriber` actually provides `protected complete()` for the purposes of overriding.  If you override that, then lets say in future we find a bug that needs to be fixed by doing some internal processing in `onComplete`, overriding `complete` won't interfere with that.  A good example of this is that currently we let netty buffer all writes, but maybe in future we might find that we need to do our own buffering, in which case, our `onComplete` handling will need to flush before invoking `complete`, but if we change that this will break, since it will bypass the flush.

providers/netty4/src/main/java/org/asynchttpclient/netty/request/body/NettyReactiveStreamBody.java
```diff
@@ -0,0 +1,140 @@
...
+                        public void operationComplete(ChannelFuture future) throws Exception {
+                            removeFromPipeline();  
+                        }
+                    });
+                }
+            });
+        }
+
+        @Override
+        public void onError(Throwable error) {
```
jroper: Same with `onComplete`, `HandlerSubscriber` provides an `error` method for the purposes of overriding.

providers/netty4/src/main/java/org/asynchttpclient/netty/request/body/NettyReactiveStreamBody.java
```diff
@@ -0,0 +1,140 @@
...
+    public void write(Channel channel, NettyResponseFuture<?> future) throws IOException {
+        if (future.isStreamWasAlreadyConsumed()) {
+            LOGGER.warn("Stream has already been consumed and cannot be reset");
+        } else {
+            future.setStreamWasAlreadyConsumed(true);
+            NettySubscriber subscriber = new NettySubscriber(channel);
+            channel.pipeline().addLast(NAME_IN_CHANNEL_PIPELINE, subscriber);
+            publisher.subscribe(new SubscriberAdapter(subscriber));
+        }
+    }
```
jroper: Shouldn't this method be capturing the `NettyResponseFuture`, and at very least invoking `future.abort` in the error callback?

api/src/test/java/org/asynchttpclient/test/TestUtils.java
```diff
@@ -92,6 +99,57 @@ public static File createTempFile(int approxSize) throws IOException {
...
+
+        @Override
+        public void subscribe(final Subscriber<? super ByteBuffer> subscriber) {
+            subscriber.onSubscribe(new Subscription() {
+                // This index is used to keep track of bytes that have been sent.
+                private volatile int currentIndex = 0;
+                // This flag will track whether this `Subscription` is to be considered cancelled or not
+                private volatile boolean cancelled = false;
+
+                public void request(long n) {
```
jroper: This will need to be modified to avoid blowing the stack in case `request` is invoked by `onNext` - the reactive stream spec requires that this be handled here, not in `Subscriber.onNext`.  A better option here might be to pull in a reactive streams implementation for the tests, eg rxjava or akka streams. @slandelle do you have a preference as to which implementation should be used?

api/src/main/java/org/asynchttpclient/Request.java
```diff
@@ -112,6 +113,13 @@
     InputStream getStreamData();
 
     /**
+     * Return the current request's body as a Reactive Stream Publisher.
+     *
+     * @return a Reactive Stream Publisher representation of the current request's body.
```
jroper: This will return null if no publisher was set, right?  That should probably be documented, we definitely don't want to expose any arbitrarily set body as a publisher as implementing that will require either depending on a reactive streams implementation from the async-http-client API, or implementing reactive streams ourselves.

api/src/main/java/org/asynchttpclient/handler/StreamedAsyncHandler.java
```diff
@@ -0,0 +1,19 @@
+package org.asynchttpclient.handler;
```
slandelle: Missing license header. Please grant copyright to "the AsyncHttpClient Project".

api/src/main/java/org/asynchttpclient/request/body/generator/IFeedableBodyGenerator.java
```diff
@@ -0,0 +1,11 @@
+package org.asynchttpclient.request.body.generator;
```
slandelle: Missing license header. Please grant copyright to "the AsyncHttpClient Project".

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
```diff
@@ -0,0 +1,152 @@
+/*
+ * Copyright (c) 2010-2012 Sonatype, Inc. All rights reserved.
```
slandelle: Please change copyright to "the AsyncHttpClient Project".

api/src/test/java/org/asynchttpclient/reactivestreams/ReactiveStreamsTest.java
```diff
@@ -0,0 +1,251 @@
+/*
+ * Copyright (c) 2010-2012 Sonatype, Inc. All rights reserved.
```
slandelle: Please change copyright to "the AsyncHttpClient Project".

api/src/test/java/org/asynchttpclient/request/body/ReactiveStreamsTest.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2010-2012 Sonatype, Inc. All rights reserved.
```
slandelle: Please change copyright to "the AsyncHttpClient Project".

providers/netty3/src/test/java/org/asynchttpclient/netty/request/body/NettyReactiveStreamsTest.java
```diff
@@ -0,0 +1,13 @@
+package org.asynchttpclient.netty.request.body;
```
slandelle: Missing license header. Please grant copyright to "the AsyncHttpClient Project".

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
```diff
@@ -0,0 +1,46 @@
+package org.asynchttpclient.netty.handler;
```
slandelle: Missing license header. Please grant copyright to "the AsyncHttpClient Project".

providers/netty4/src/main/java/org/asynchttpclient/netty/request/body/NettyReactiveStreamBody.java
```diff
@@ -0,0 +1,140 @@
+/*
+ * Copyright (c) 2010-2012 Sonatype, Inc. All rights reserved.
```
slandelle: Please change copyright to "the AsyncHttpClient Project".

providers/netty4/src/test/java/org/asynchttpclient/netty/reactivestreams/NettyReactiveStreamsTest.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright (c) 2010-2012 Sonatype, Inc. All rights reserved.
```
slandelle: Please change copyright to "the AsyncHttpClient Project".

providers/netty4/src/test/java/org/asynchttpclient/netty/request/body/NettyReactiveStreamsTest.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) 2010-2012 Sonatype, Inc. All rights reserved.
```
slandelle: Please change copyright to "the AsyncHttpClient Project".

api/src/test/java/org/asynchttpclient/test/TestUtils.java
slandelle: If we were to go with Akka, I guess we would have to turn the maven project into a Scala/Java mixed one so we can use TestKit. I'm afraid this would make the project more difficult to build for most people. So, even thought my heart would love to go with Akka, I think it's more reasonable to stay with a simple Java project and go with rx. WDYT?

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/Processor.java
slandelle: Actually, there's 2 cases where retry makes sense. Sadly, AHC doesn't currently tell them appart, as it was originally intended to deal with file downloading.

1. The request was sent, and the connection crashed while the response body chunks were being received.
2. The request was not sent, because the connection crashed first.

The second case typically happens when reusing a pooled connection, and the remote peer closes it (keep-alive idle timeout, server being overloaded). We do check that the connection is alive when pooling it, and just before writing, but there's always a time window where this could happen. Under heavy load, this happens all the time, as soon as the server implements idle timeout. Browsers do implement a retry to deal with such use case. AHC has a maxRetry default value of 4.

Note that you can be notified of a retry when you implement `AsyncHandlerExtensions`, such you can clean/reset chunks that you've received so far.

What's the expected behavior of reactive-streams when the connection crashes? Is it up to the user to implement retry logic?

api/src/main/java/org/asynchttpclient/BoundRequestBuilder.java
dotta: I've done so for consistency with regard to other methods overridden in this same source. See this comment https://github.com/dotta/async-http-client/blob/issue/544-reactive-streams-support/api/src/main/java/org/asynchttpclient/BoundRequestBuilder.java#L49-L52 for context.

slandelle: I'm going to remove all of those from master:
* [CLJ-126](http://dev.clojure.org/jira/browse/CLJ-126) seems to be related to Java 5, that we don't support.
* [CLJ-259](http://dev.clojure.org/jira/browse/CLJ-259) has been open in 2010. Come on, 5 years to come up with a fix!
* [https://github.com/cch1/http.async.client](http.async.client) has just been upgraded to AHC 1.8. So they won't upgrade to AHC 2 any time soon.
* I just hate it with my guts when people push their own mess to others.

slandelle: Done: e8060d166b4581be25deccd8f5474ee47dbfc1d0

slandelle: @dotta Please remove those too.

api/src/main/java/org/asynchttpclient/Request.java
dotta: > This will return null if no publisher was set, right?

I'm not sure what is the contract of `Request`. Though, by looking at `RequestBuilder`, it seems `null` is an acceptable value to return in all the getters in `Request`.

>  we definitely don't want to expose any arbitrarily set body as a publisher as implementing that will require either depending on a reactive streams implementation from the async-http-client API, or implementing reactive streams ourselves.

Here I'm not following. The AHC client already depends on reactive-stream, and the `setBody(Publisher<T>)` would be only a conveniency method so that you can write `client.preparePut(getTargetUrl()).setBody(LARGE_IMAGE_PUBLISHER)` instead of `client.preparePut(getTargetUrl()).setBody(new ReactiveStreamBodyGenerator(LARGE_IMAGE_PUBLISHER))`. That said, I can see the argument of not polluting the `Request` interface with yet another method.

@slandelle WDYT?

api/src/main/java/org/asynchttpclient/BoundRequestBuilder.java
dotta: :+1:

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
dotta: Ah, I've copied the license note from another source. Grepping on `2010-2012 Sonatype` reveals quite a few hits, just letting you know in case this is something you want to update.

slandelle: Yeah, Ning and Sonatype have successively employed @jfarcand to develop this project. I can't change copyrights without the owner donating.

api/src/main/java/org/asynchttpclient/Request.java
slandelle: Sadly, I don't see any convenient solution for not having core API depend on reactive-streams.

dotta: Any thought on having `getPublisherData()` in `Request`?

api/src/main/java/org/asynchttpclient/request/body/generator/ReactiveStreamBodyGenerator.java
dotta: Got it, thanks.

providers/netty4/src/main/java/org/asynchttpclient/netty/request/body/NettyReactiveStreamBody.java
dotta: I did not notice `complete` and `error` where `protected`. Thanks!

providers/netty4/src/main/java/org/asynchttpclient/netty/request/body/NettyReactiveStreamBody.java
dotta: But I have a `publisher` of type `Publisher<ByteBuffer>` that wants to subscribe to the `NettySubscriber`. See https://github.com/AsyncHttpClient/async-http-client/pull/963/files#diff-201ec22bdb5bc3211518d720cfce85d3R67

api/src/test/java/org/asynchttpclient/test/TestUtils.java
dotta: > If we were to go with Akka, I guess we would have to turn the maven project into a Scala/Java mixed one so we can use TestKit.

We don't need the TestKit. A dependency to akka-streams would be enough. But, even if we had to use TesKit, I don't see why we would then need to turn this project into a Scala/Java mixed one.

That said, I think we should go with rxjava because with akka-streams turning a `Source` into a `Publisher` requires materialization (meaning we need to create an `ActorMaterializer` and hence also an `ActorSystem`). Consider that with rxjava it's straightforward to convert an `Observer` into a `Publisher`.

api/src/main/java/org/asynchttpclient/Request.java
dotta: Alright, forget this, turns out I don't need it. See the upcoming commit for details.

providers/netty4/src/main/java/org/asynchttpclient/netty/request/body/NettyReactiveStreamBody.java
dotta: Absolutely. I'll add a test and fix it.

dotta:
@slandelle I believe I've addressed all comments. The only pending one is https://github.com/AsyncHttpClient/async-http-client/pull/963#discussion_r38438885, which you are discussing with @jroper.

Let me know if you have other comments, or if I should rebase and squash commits.

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/Processor.java
jroper: In this case the stream will simply be notified with the exception via `Subscriber.onError`.

If we did attach the publisher to the future, I *think* we'd have to make a mutable field in `NettyResponseFuture`.  It appeared to me that everything in `NettyResponseFuture` is immutable, and I assumed there was a reason for that - by keeping the only mutable thing the channel attribute, that makes it easier to reason about the state of the system.

On the two crash conditions - the second one is not impacted here.  It's only once we start receiving a body that we change the attribute, before that, all the handling is the same.  For the first crash condition, I think it would be quite straight forward to modify the exception handling such that connections are still retried.  Note that unlike other `AsyncHandlers`, the `StreamedResponsePublisher` changes the state of the pipeline (inserts a handler), and requires events, such as the exception itself, to be forwarded down the channel so that the subscriber that's handling the body receives them.  This is why I think the handling for this should (or must) be in the `Processor`, it requires custom handling at that level.

I'll have a closer look at fixing the exception handling to implement retry - I've realised a `StreamedAsyncHandler` can always return `ABORT` from `onStream` if it's invoked a second time, and that's what we'll do in Play, so I don't think retrying is that big an issue.

slandelle: The second one is the real important one. It would be great if AHC could tell them appart, maybe some day...

Please let me know when you guys are done so I can merge.

Thanks a bunch!

dotta: @jroper Isn't this https://github.com/dotta/async-http-client/blob/issue/544-reactive-streams-support/providers/netty4/src/main/java/org/asynchttpclient/netty/handler/Processor.java#L158-L159 already taking care of retrying for the first condition. If not, could you or @slandelle give me a head start on this.

jroper: @dotta I believe https://github.com/dotta/async-http-client/blob/issue/544-reactive-streams-support/providers/netty4/src/main/java/org/asynchttpclient/netty/handler/Processor.java#L177 is where the retry is initiated, and that code is only executed if the attribute is a `NettyResponseFuture`, so that's where we need to change things.

dotta: @jroper I've tried to write a test for it, but haven't succeeded. Is the following what you had in mind to fix it? https://github.com/dotta/async-http-client/commit/eab7a33413b418f69eb99f37c2b4fb1da52f3961 \cc @slandelle

slandelle:
@dotta @jroper Sorry, busy day. Will have a look this week-end.

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/Processor.java
dotta: Is there a way to emulate a connection crash? As said, I've tried to write a test for this, but I can't make the connection crash (so that I can confirm the retry logic is run). My attempt was to call `channel.close().await()`, but that blocks indefinitely.

dotta: By the way, I was calling `await()` to force [Processor.exceptionCaught](https://github.com/dotta/async-http-client/blob/issue/544-reactive-streams-support/providers/netty4/src/main/java/org/asynchttpclient/netty/handler/Processor.java#L155) to be called. If I don't call `await`, then of course execution continues, but the `Processor.exceptionCaught` method is still never called.

dotta: Alright, I've managed to write a test, see https://github.com/dotta/async-http-client/commit/a373a21e393ca5d071fd0cbb60a6fd92f8d4f92d (not really a straightforward and easy-to-read one. I'm all hears if there is a better way of writing it).

dotta:
@slandelle I'm happy to squash commits and rebase if you give me your bless :-) In the meanwhile, I'll update Play master to take advantage of the changes in this PR. Whohoo!

providers/netty4/src/main/java/org/asynchttpclient/netty/handler/Processor.java
slandelle: Yeah, those are not easy to test... What I did in the past was to do a kind of load test to try to reproduce.

slandelle:
@dotta Sorry, had some unexpected things to deal with first... Yes, could you please squash and rebase? Let's merge this so everyone can move forward and get everything stable in the next weeks so we can release 2.0.0.

dotta:
Whohooo! I'm on it, thanks! :-)

slandelle:
@jroper @dotta  Thanks guys!

dotta:
@slandelle Thank you for merging!

I'm updating the Play codebase to use AHC master, will let you know if I hit any issue.

slandelle:
Great!
What's your timeline for Play 2.5?

dotta:
I believe it's November 2015, but I'll let @jroper confirm.

Once I've verified that things work properly on the Play side, I'd really appreciate if you could release a 2.0.0 milestone (in case you would like to wait a bit more before cutting a 2.0.0 final).

slandelle:
I can easily cut an alpha.

I'm still considering some refactorings (maybe Proxies) + I might want to break a few things to improve DNS handling in Gatling.

dotta:
Sounds good. I'll ping you soon :-)

dotta:
@slandelle I've a branch in Play that is now using the new `StreamedAsyncHandler`, and it seems to be working just fine (and I'm really pleased with the result, the code is much, much, much easier to read now!). Could you cut a release so that I can push that work in Play?

slandelle:
I just released 2.0.0-alpha9.
Cheers!

jroper:
Thanks @slandelle!

dotta:
@slandelle Thank you!

PR #964, opened 2015-09-08
Fix NPE in ThreadSafeHMAC when secret is not set

bomgar:
apparently the '&' is important even if no secret is present


slandelle:
Sorry, but what does this fix exactly? Could you provide a test case that fails without it?

bomgar:
Test added.
The OAuthSignatureCalculator already handles it correctly if the user key is null. But the secret causes a NPE in the ThreadSafeHMAC.

slandelle:
Isn't the secret mandatory?

bomgar:
The user key and secret are not necessary. It is possible to authenticate a consumer only.

slandelle:
Thanks!

PR #967, opened 2015-09-10
update a comment in the scheduleTimeouts method.

dapchen:


slandelle:
Thanks for reporting (I fixed with a comment more to my liking)!

PR #977, opened 2015-09-23
ISSUES-976: FluentCaseInsensitiveStringsMap keyLookup un-sync with values after iterator remove 

ben-tsai:
Just proposing a quick fix unless we're planning to do something fancier.

Thanks,
Ben


slandelle:
FluentCaseInsensitiveStringsMap was dropped.

slandelle:
But thanks anyway :)

PR #1019, opened 2015-10-26
add a unit test to test reactive stream support for downloading files

Tianwei-Li:
I did some test about reactive stream support in AHC and found a wield issue.
The test is to get a file from a http file server, in order to implement back pressure from the client to server, I used the StreamedAsyncHandler.

The issue is if the file I want to get is large (above 10K), it works fine.
But if the file is small (below 50 bytes), the onComplete of Subscriber got called, but the onNext never got called, and then get zero bytes from the server.

The unit test file is ReactiveStreamsDownLoadTest.java, it includes two tests, one is to download a large file, the other is to download a small file.

The HttpStaticFileServer code is copied from unit test in Netty.


slandelle:
@jroper @dotta Ping :smile:

dotta:
@slandelle Haven't forgot about this, but I'm a bit swamped. Hopefully, I'll be able to have a look at it tomorrow.

slandelle:
@dotta No worries. Just that I'm so swamped too that I definitively need you guys to help maintaining the reactive stream support :)

dotta:
@slandelle No worries. I've pinged the rest of the Typesafe Play team to see if someone has time to look into this.

Issue #648, opened 2014-07-24
Better Continuous Integration 


slandelle:
There's a BuildHive account: https://buildhive.cloudbees.com/job/AsyncHttpClient/job/async-http-client/ but it's no longer used.

Let's switch to Travis and let's deploy snapshots on Sonatype.

slandelle:
The main issue is with the tests that rely on external websites, whose behavior is different depending on the client IP location. Typically, google.com redirects to google.fr when calling from France.

There's a plan for stopping using external websites in our integration tests suite, see discussion in #1069.
If we fail to do so, we can still isolate those tests into a dedicated maven profile that wouldn't run on Travis.

slandelle:
@dotta ping

dotta:
:fireworks:

Issue #496, opened 2014-03-05
Fix microsoft.com based tests in RemoteSiteTest


slandelle:
microsoft.com has frequent DNS issues that make the test fails. This usually appears randomly on Cloudbees, but I was able to reproduce locally here in France.
I've disabled them, but we should try to depend less on remote website.

PR #1026, opened 2015-10-31
Use travis CI

dotta:
Hey Stephane, I thought it could be useful to use Travis for validating PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: java
+jdk:
+  - oraclejdk8
+
+script: mvn clean install


dotta:
@slandelle If you like the idea of using Travis, you'll need to do a few things to set it up (see http://docs.travis-ci.com/user/getting-started/). But it should be a fairly quick process.

slandelle:
@dotta You just treated me noob :) Gatling is built with Travis.
There's a big issue with the AHC tests: some old yet important ones depend on external websites, and sometimes, behaviour depends on your IP zone. Typically, those will break when running from North America. Guess where Travis run? See #496 and #648.

dotta:
@slandelle Sorry Stephane, I really did not mean to treat you as a noob. I just wanted to make it extremely simple to get this in, and knowing you are pretty busy I thought a link could be handy. Please, again, forgive me :-)

Thanks for the background info about why travis wouldn't work with ahc.

slandelle:
@dotta no pro :) Moving to Travis is something I have been wanting to do for a very long time, but I could never find time to fix those tests. Originally, they could only succeed when run from NA. I went with the quick and dirty so they could succeed from anywhere but NA...

The other bad point about AHC is the lack of doc, and things will get worse with AHC2.

I guess the main issue with AHC is that it's always been developed by people whose main goal wasn't this project: JFA changed from Ning to Sonatype to his own stuff on Atmo, I'm here because of Gatling, you Typesafe guys are here because of Play, Oracle people were there because of Grizzly and Jersey and dropped as their projects went down, etc.

dotta:
> Originally, they could only succeed when run from NA. I went with the quick and dirty so they could succeed from anywhere but NA...

LOL, pragmatic :-)

> I guess the main issue with AHC is that it's always been developed by people whose main goal wasn't this project

I hear you.

slandelle:
@dotta There's a chance we'll be able to fix the integration test suite for AHC2. If it's still not possible, we could move the problematic tests to a dedicated profile that wouldn't run on Travis.

So we'll be able to make it!

dotta:
nice!

PR #1031, opened 2015-11-09
Add URI to message of TimeoutException when timeout

Atry:


slandelle:
Hi @Atry,

What's the motivation for this change? I'd tend to disagree. The point of having the remoteAddress is to be able to detect a failing node in a cluster.

Atry:
Isn't the URI the more significant information for a normal request?

As a client, how do you feel when you caught an exception and did not know which URI you were requesting

slandelle:
URI wouldn't be sufficient: what about posting a body, what about headers (eg auth token)?

Your real problem is that your losing request context: which you could still have if you were mapping the failure.

The current design:
* let you have consistent exception messages
* reduce memory pressure by not storing references to the requests (that can have huge payloads) in possibly expired timeout tasks that have to been pushed out yet out of the timer
* give you the proper remote address so you can tell cluster nodes

I really need all of those for Gatling, so sorry, I'm going to reject this one, as there's an easy workaround.

Atry:
OK, thank you

PR #1032, opened 2015-11-09
Add URI to message of TimeoutException when timeout

Atry:


slandelle:
Hey @Atry !
Sorry for the delay :(

I have 2 issues with your PR (appart from fact that your branch has conflicts and can't be merged).

First, I display those Exception messages in Gatling, and having the full url in the message would be both super long and would blow off the number of different messages that get displayed. I would be fine with either having a config option to switch between current behaviour and the one you suggest.

Then, it's a bad idea to store the full Request inside the Task. Cancelled tasks could be lingering some time in the Timer, increasing memory pressure (e.g. the request body won't be quickly gc'ed). Only storing the Uri should suffice.

Atry:
This PR is a duplicate of https://github.com/AsyncHttpClient/async-http-client/pull/1031

slandelle:
Oh right, sorry.
So actually, I think you're right, and providing the url might prove useful, even if not sufficient for every use case (think ElasticSearch that always send a JSON body, even for GET requests).

If you agree with my suggestions, please feel free to provide a PR.

Atry:
I was convinced by your design mentioned in #1031 that the users could store URL in their error handler.

slandelle:
OK, then :)

PR #1038, opened 2015-11-17
Code quality fix - IP addresses should not be hardcoded

faisal-hameed:
Resolving occurrences of Sonar rule squid:S1313 - “IP addresses should not be hardcoded"
You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S1313
We have replaced a bunch of references to 127.0.0.1 in the test code with a utility method to lookup the 
IP address to use for unit tests. This is useful because it leaves you with one place to maintain this constant and makes the code slightly more readable.

Please let me know if you have any questions.

Faisal.

### DIFF
--- a/client/src/test/java/org/asynchttpclient/AbstractBasicTest.java
+++ b/client/src/test/java/org/asynchttpclient/AbstractBasicTest.java
@@ -21,6 +21,7 @@ import static org.asynchttpclient.test.TestUtils.newJettyHttpServer;
 import static org.testng.Assert.fail;
 
 import org.asynchttpclient.test.EchoHandler;
+import org.asynchttpclient.test.TestUtils;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.handler.AbstractHandler;
 import org.slf4j.Logger;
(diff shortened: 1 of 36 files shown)


slandelle:
IMO, it looks less readable and make it harder to tell appart tests that use a local server and ones that use a remote one.

Why not just replace `127.0.0.1` with `localhost`?
I have no idea why those tests use "127.0.0.1", that's something I inherited.

faisal-hameed:
Hi @slandelle 
My intention was to fix following issue to improve code quality.
1. squid:S1313 : “IP addresses should not be hardcoded"
2. squid:S1192 :  "String literals should not be duplicated"
3. Maintainability : We can easily change the IP address at one place.
But IMHO by using "localhost" we are still duplicating the string literals.
Instead of using "localhost" we can rename getUnitTestIpAddress() to getLocalhostIP(), which makes code readable and resolve all aformentioned issues.
What's you suggestion?

slandelle:
IMO, this loses a lot in readability.

faisal-hameed:
Ok. I will will replace "127.0.0.1" with "localhost".

slandelle:
Closing. Those are tests where hardcoding 127.0.0.1 and localhost really makes sense.

Thanks anyway.

PR #1041, opened 2015-11-19
Code quality fix - Strings literals should be placed on the left side when checking for equality

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rule squid:S1132 - “Strings literals should be placed on the left side”. 
You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S1132

Please let me know if you have any questions.

Faisal.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/Realm.java
+++ b/client/src/main/java/org/asynchttpclient/Realm.java
@@ -323,12 +323,12 @@ public class Realm {
 
             // prefer auth over auth-int
             for (String rawServerSupportedQop : serverSupportedQops) {
-                if (rawServerSupportedQop.equals("auth"))
+                if ("auth".equals(rawServerSupportedQop))
                     return rawServerSupportedQop;
             }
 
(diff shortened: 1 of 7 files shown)


slandelle:
I know this comes from good will, but...
Just applying generic rules without further thought has nothing to do with improving code quality.

Here, specifically:
* Yoda style is everything but readable
* some of the code you suggest to change is already protected against NPE when the parameter is expected to be optional. ex: turning `algorithm == null || algorithm.equals("MD5")` into `algorithm == null || "MD5".equals(algorithm)`
* trying to blindly change code so that it doesn't crash with an NPE when an expected parameter is missing, but then behave in an unexpected manner is so very dangerous. I'd rather have an NPE, and then decide if it was a bug and that this parameter could rightfully be optional, instead of never being aware that this parameter is sometimes missing, being possibly a user error.

Those qualimetry tools can only raise some warnings that have then to be evaluated in the context of the application and the code. They are by no way silver bullets, and some of their rules are plain BS.

PR #1045, opened 2015-11-23
Keep Content-Type when redirecting POST request with body

rodrigosetti:
For the POST redirects that keeps the body (302 strict and 307), also keep the Content-Type header.


slandelle:
Thanks!

PR #1056, opened 2015-12-09
Add setter for 'usePooledMemory' parameter at DefaultAsyncHttpClientConfig.Builder

mielientiev:
I've added setter for 'usePooledMemory' parameter at DefaultAsyncHttpClientConfig.Builder


slandelle:
lol thanks

Issue #1057, opened 2015-12-15
AHC should accept a classloader at creation


dotta:
When creating an AHC instance, the classloader used to load the ahc-default.properties is the thread's context classloader. While this is a good default, it would be good if AHC would offer an explicit constructor taking a classloader.

slandelle:
Hey @dotta 
You know the usual answer: contributions welcome! ;-)

dotta:
@slandelle Of course! :-)

hepin1989:
@dotta for Play's need for dynamic classloader?

dotta:
@hepin1989 I'd say just to make things explicit and not having to side-effect the context classloader to successfully create AHC instances when done from a different classloader, e.g., when starting a Play server from a sbt task.

slandelle:
ping @dotta :)

dotta:
@slandelle Hey Stephane, this is not a priority for me, I just filed a ticket to keep track of it but I don't have spare cycles atm to work on this. Hope it's alright.

slandelle:
No pro. Just that I expect to be able to release AHC 2 within one month, so if there's changes you want to push for Play, now's the time :)

dotta:
Alright. Thanks for pinging me!

slandelle:
Feel free to ping if you want to contribute and I'll reopen.

PR #1075, opened 2016-01-08
Property loading fix for OSGi

paulbakker:
In an OSGi environment, the ContextClassLoader is not by default set to a classloader that will read the property files from the bundle. Instead, the classloader of the bundle should be used.

It is possible to work around this issue by setting the ContextClassLoader explicitly, but that's confusing (unless someone really understand the classloader model in OSGi). This PR is an easy workaround for the issue.


slandelle:
@dotta was suggesting letting one pass a ClassLoader, see #1057.
Not sure what's the better approach.

paulbakker:
Is there a reason we can't have both? I can see the use cases for setting a class loader specifically in some environments, but it's not a very nice default if you must do this if you didn't create your own properties.

slandelle:
Yeah. Merged, thanks!

paulbakker:
That's fast, thanks!

PR #1082, opened 2016-01-15
Fixed an error in the README.md

thisismana:
the `S` in line 2 of the configuration example shouldn't be there


slandelle:
Thanks

PR #1098, opened 2016-02-19
fixed NPE that occured when setting query params before url was set

truslan:
Hello!
I have some cases with RequestBuilder when query parameters are set before setting uri. Current builder implementation produces NullPointerException in those scenarios. I've made little fix.


slandelle:
Thanks!

PR #1105, opened 2016-03-02
Introduce AsyncHttpSingle

twz123:
Since `AsyncHandlers` only produce a single value, it seems appropriate to use RxJava's `Single<T>` instead of `Observable<T>`. This PR introduces new factory methods for `Single` comparable to those offered by `AsyncHttpObservable`.

In addition to that, the `Single` instances created by `AsyncHttpSingle` honor the subscription state, so if a subscriber unsubscribed from the `Single` during request execution, this will be forwarded to AHC as `State.ABORT` in the next call to one of the `onXXX` methods. Besides `AsyncCompletionHandlerBase`, arbitrary `AsyncHandlers` can be used for request processing, too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,12 @@ after_success: |
 
 sudo: false
 
+# https://github.com/travis-ci/travis-ci/issues/3259
+addons:
+  apt:
+    packages:
+      - oracle-java8-installer
+
 # Cache settings
 cache:
   directories:
--- a/extras/rxjava/src/main/java/org/asynchttpclient/extras/rxjava/UnsubscribedException.java
+++ b/extras/rxjava/src/main/java/org/asynchttpclient/extras/rxjava/UnsubscribedException.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2015 AsyncHttpClient Project. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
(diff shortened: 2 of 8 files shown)


slandelle:
> Since AsyncHandlers only produce a single value

That's the case for HTTP/1, but not for WebSockets, nor HTTP/2 (when we'll support it). So I wonder...

twz123:
I see... But that would mean that the `AsyncHandler` API needs to be changed or extended in some way?

My initial impulse for this change was to implement unsubscription support and to use custom AsyncHandlers. So if you're in doubt, would it make sense to rework this PR to extend the AsyncHttpObservable class instead? This would be quite tricky to get backpressure right (at least for the multiple events case).

.travis.yml
```diff
@@ -11,6 +11,12 @@ after_success: "[[ $TRAVIS_PULL_REQUEST == \"false\" ]] && mvn deploy"
 
 sudo: false
 
+# https://github.com/travis-ci/travis-ci/issues/3259
+addons:
+  apt:
+    packages:
+      - oracle-java8-installer
```
slandelle: Is this necessary? I wonder if Travis has to install it on every build over and over again, making the build longer.

slandelle:
@twz123 Could you please answer to comment, and then rebase so I can merge?

> But that would mean that the AsyncHandler API needs to be changed or extended in some way?

Yes. My intent was to was you that I absolutely no idea what the future API will look like, but probably that I'll have to break things for HTTP/2.

.travis.yml
twz123: Indeed this is necessary for the build to succeed. With JDK 8u31, the
sources won't compile, since the compiler fails to infer some types. Either
I'll have to rework the PR and introduce some boilerplate, or we use a
newer JDK.

The issue in the comment states other, possibly faster ways to get a newer
JDK, for instance to use a newer Ubuntu version. Shall I investigate
further?

twz123:
I'll rebase tomorrow morning, thanks for your patience!

.travis.yml
slandelle: No, that's fine. Honestly, I had missed this JDK bug. Mostly do Scala :)

slandelle:
Cool, thanks!

twz123:
@slandelle rebased…

slandelle:
Thanks!

PR #1106, opened 2016-03-03
Make Travis available for all branches

twz123:
I changed the Travis script a bit to be usable for all branches by changing the predicates for automatic deployment and removing the branch whitelist.

I think this is helpful, since I like to test Travis builds in my forks before creating pull requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,20 @@
 language: java
 jdk:
   - oraclejdk8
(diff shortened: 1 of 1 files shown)


slandelle:
Thanks! Beware that the test suite won't run properly for pre-AHC2 branches (some tests depend on external sites whose behavior depends on IP: works from Europe, not from the US where Travis hosts are).

PR #1112, opened 2016-03-11
Update README with org.asynchttpclient

wsargent:
Unscarcifying documentation!

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,19 +15,19 @@ Latest `version`: [![Maven][mavenImg]][mavenLink]
 
 ## Installation
 
-First, in order to add it to your Maven project, simply add this dependency:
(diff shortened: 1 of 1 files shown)


slandelle:
> Unscarcifying documentation!

LOL.

Let's change the package names in the README when 2.0.0 is out and AHC1 is officially deprecated. I plan to do that along with Gatling 2.2.0 at the end of the month. I'll hold this PR on until then.

But then, I guess this isn't sufficient for most users. A second step would probably to migrate the sample from JFA's blog in here.

wsargent:
> sample from JFA's blog 

What blog post sample?  I haven't heard of this.

slandelle:
See links in one of the first lines in this README.

wsargent:
I think the 2011 blog post isn't current for 1.9.x, even.

There's also https://asynchttpclient.github.io/async-http-client/ site which still has the 1.8.0-SNAPSHOT.

wsargent:
woo!

Issue #1113, opened 2016-03-16
Race condition can cause Netty connection pool to close currently active connections


malaporte:
Under load we noticed that we sometimes get unexplainable read timeouts on connections on some connections for which we can confirm that a successful response was indeed returned by the server.

After a sizeable amount of investigation where we doubted everything ranging from our NATs to AWS ELBs, it turns out that sometimes `DefaultChannelPool` will close a connection that is in fact actively being used, causing and eventual read timeout in the code that waits for a result. This is because nothing prevents a channel that has been identified as expired from being handed out of the pool before it reaches the point where it is indeed closed. There is in fact a check there to prevent connections for which a Future is active from being closed, but this don't cover the case where the connection has been taken out of the pool but hasn't sent a request yet.

We get this problem in 1.9.33 but it also exists in the 2.0 version.

I have implemented a fix for this; I'll be submitting a PR shortly.

slandelle:
Fixed on master w/ #1114

slandelle:
Ping @wsargent @jroper 
Wasn't something similar reported on Play?

wsargent:
Don't remember anything off the top of my head. There's https://github.com/AsyncHttpClient/async-http-client/issues/1027 but that's sort of the inverse of this, where threads stay open after they should be closed.

wsargent:
There is https://github.com/AsyncHttpClient/async-http-client/issues/1027#issuecomment-175249025 though...

slandelle:
Yes, that's the one I was looking for. Really seems to be the same, thanks!

slandelle:
backported on 1.9 with 1d5b880e7e07b53942ded35bad82f39f65ba5761

slandelle:
@malaporte just released 1.9.35

PR #1114, opened 2016-03-16
Prevent Netty connection pool from killing active connections

malaporte:
This fixes the issue https://github.com/AsyncHttpClient/async-http-client/issues/1113 I reported previously by ensuring that a channel can't be both allocated to a caller and expired by the timer task. It uses a simple atomic boolean to grant "ownership" of the channel to one or another.

While testing I also found another issue: channels that are rejected by the pool are not closed properly by `ChannelManager`. This is particularly impactful on the test I wrote to reproduce the initial issue, as it quickly leads to all available sockets to be exhausted.

And finally, the code that does check for TTL is broken in 2.0 because the creation time for a channel is never put into the map due to the wrong map being used when checking if the partition exists. I'm pretty sure the change I made is correct although I haven't researched this part yet (my initial fix was made on top of 1.9.33 which doesn't have this issue).

I added a test that can be used to reproduce the initial problem (`TimeToLiveIssue`) but now works OK once the fix is made.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
@@ -97,10 +97,15 @@ public final class DefaultChannelPool implements ChannelPool {
     private static final class IdleChannel {
         final Channel channel;
         final long start;
+        final AtomicBoolean owned = new AtomicBoolean(false);
 
         IdleChannel(Channel channel, long start) {
             this.channel = assertNotNull(channel, "channel");
             this.start = start;
+        }
+
+        public boolean takeOwnership() {
+            return owned.compareAndSet(false, true);
         }
 
         @Override
@@ -148,25 +153,15 @@ public final class DefaultChannelPool implements ChannelPool {
             return idleTimeoutChannels != null ? idleTimeoutChannels : Collections.<IdleChannel> emptyList();
         }
 
-        private boolean isChannelCloseable(Channel channel) {
-            Object attribute = Channels.getAttribute(channel);
-            if (attribute instanceof NettyResponseFuture) {
-                NettyResponseFuture<?> future = (NettyResponseFuture<?>) attribute;
-                if (!future.isDone()) {
-                    LOGGER.error("Future not in appropriate state {}, not closing", future);
-                    return false;
-                }
-            }
-            return true;
-        }
-
         private final List<IdleChannel> closeChannels(List<IdleChannel> candidates) {
 
             // lazy create, only if we have a non-closeable channel
             List<IdleChannel> closedChannels = null;
             for (int i = 0; i < candidates.size(); i++) {
+                // We call takeOwnership here to avoid closing a channel that has just been taken out
+                // of the pool, otherwise we risk closing an active connection.
                 IdleChannel idleChannel = candidates.get(i);
-                if (isChannelCloseable(idleChannel.channel)) {
+                if (idleChannel.takeOwnership()) {
                     LOGGER.debug("Closing Idle Channel {}", idleChannel.channel);
                     close(idleChannel.channel);
                     if (closedChannels != null) {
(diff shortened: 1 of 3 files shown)


slandelle:
Hi @malaporte

Thanks a lot! Reviewing atm.

One first question: can the test you added actually fail, as a unit test, or is it something that's intended to create some conditions that you have to check, eg with netstat?

malaporte:
It fails if the fix isn't present, for example if you comment out the part that calls `takeOwnership`, etc.

malaporte:
(I should say: it eventually fails after some time running)

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -257,7 +252,7 @@ private boolean offer0(Channel channel, Object partitionKey, long now) {
     }
 
     private void registerChannelCreation(Channel channel, Object partitionKey, long now) {
-        if (channelId2Creation.containsKey(partitionKey)) {
+        if (partitions.containsKey(partitionKey)) {
```
slandelle: Actually, there's a stupid bug here: the test was supposed to be `if (!channelId2Creation.containsKey(partitionKey))`. The goal was to avoid creating a `new ChannelCreation` if the entry is already there. And you don't gain anything here with using `computeIfAbsent` instead of `putIfAbsent` because the lambda would be capturing, and that the key usually exists.

Could you please revert your change and fix the if test?

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -148,25 +153,15 @@ private boolean isIdleTimeoutExpired(IdleChannel idleChannel, long now) {
...
         private final List<IdleChannel> closeChannels(List<IdleChannel> candidates) {
 
             // lazy create, only if we have a non-closeable channel
             List<IdleChannel> closedChannels = null;
             for (int i = 0; i < candidates.size(); i++) {
+                // We call takeOwnership here to avoid closing a channel that has just been taken out
+                // of the pool, otherwise we risk closing an active connection.
                 IdleChannel idleChannel = candidates.get(i);
-                if (isChannelCloseable(idleChannel.channel)) {
+                if (idleChannel.takeOwnership()) {
```
slandelle: So your opinion is that it's safe to drop `isChannelCloseable` because ownership guarantees that we can no longer never end up with a incomplete future?

client/src/test/java/org/asynchttpclient/netty/TimeToLiveIssue.java
```diff
@@ -0,0 +1,40 @@
+package org.asynchttpclient.netty;
```
slandelle: Please add package header

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
malaporte: You mean `if (!channelId2Creation.containsKey(channelId(channel))) {` right? Aka the key is not the partition key as in your sample? This plus the absence of the `!` is what got me confused initially...

slandelle: Yeah, right (was looking at old code and got confused).
So yes. And please compute channelId only once.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
malaporte: Well my very much uninformed opinion was that this check existed to prevent a channel with an incomplete future (e.g. one that is in use) from being closed, but this failed because that future is only registered when the request is sent. Normally if I haven't screwed up my change will prevent that from happening, so yes it's probably safe, unless there was some hidden meaning there that I didn't grasp...

malaporte:
Should the UT be flagged with `@ignore`? Since it's somewhat long to run..

slandelle:
Absolutely please

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
slandelle: Let's drop it

slandelle:
Once you're done, please rebase into one single commit.

slandelle:
Thanks a lot!

slandelle:
@malaporte Do you want to push the fix on 1.9 too?

malaporte:
If you'd left me some more time I'd have gladly provided that :) Thanks!

malaporte:
Wow and thanks for the fast release, much appreciated.

slandelle:
No pro. This bug has been lurking around for some time, but I didn't get a chance to reproduce it, so  I really appreciate your help here! Thanks again!

PR #1122, opened 2016-03-29
Allow AhcEndpoint address override in grizzly provider

elrodro83:
Add usage of Request#getInetAddress() in grizzly provider.

This allows for overriding the address where the physical connection is established, by setting it in the request instead of letting the provider use the host from the url or the proxy.

Netty is already doing this (https://github.com/AsyncHttpClient/async-http-client/blob/1.9.x/src/main/java/com/ning/http/client/providers/netty/request/NettyRequestSender.java#L354) among with other name resolution stuff, and it is a feature i am needing in grizzly provider.

### DIFF
--- a/src/main/java/com/ning/http/client/providers/grizzly/ConnectionManager.java
+++ b/src/main/java/com/ning/http/client/providers/grizzly/ConnectionManager.java
@@ -129,12 +129,12 @@ class ConnectionManager {
             port = getPort(scheme, uri.getPort());
         }
         
-        final String partitionId = getPartitionId(request, proxy);
+        final String partitionId = getPartitionId(request.getInetAddress(), request, proxy);
         Endpoint endpoint = endpointMap.get(partitionId);
         if (endpoint == null) {
             final boolean isSecure = Utils.isSecure(scheme);
             endpoint = new AhcEndpoint(partitionId,
-                    isSecure, host, port, request.getLocalAddress(),
+                    isSecure, request.getInetAddress(), host, port, request.getLocalAddress(),
                     defaultConnectionHandler);
 
             endpointMap.put(partitionId, endpoint);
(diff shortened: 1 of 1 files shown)


src/main/java/com/ning/http/client/providers/grizzly/ConnectionManager.java
```diff
@@ -219,10 +219,11 @@ static boolean isKeepAlive(final Connection connection) {
         return !IS_NOT_KEEP_ALIVE.isSet(connection);
     }
     
-    private static String getPartitionId(Request request,
+    private static String getPartitionId(InetAddress overrideAddress, Request request,
             ProxyServer proxyServer) {
-        return request.getConnectionPoolPartitioning()
-                .getPartitionKey(request.getUri(), proxyServer).toString();
+        return overrideAddress != null ? overrideAddress.toString() + "_" : "" +
```
oleksiys: is it supposed to be
(overrideAddress != null ? overrideAddress.toString() + "_" : "") +
           request.getConnectionPoolPartitioning()
           .getPartitionKey(request.getUri(), proxyServer).toString();

?

slandelle:
First, before going into reviewing the implementation (I'll leave this for @oleksiys as this is on the Grizzly side), there's a very important step! You didn't describe the purpose of your proposed change!!! What is this for? Is it a bug? A new feature?

src/main/java/com/ning/http/client/providers/grizzly/ConnectionManager.java
elrodro83: Yes, thanks for pointing out. Fixed it.

elrodro83:
@slandelle Thanks, i just edited the first comment in the PR.

oleksiys:
@slandelle can we ignore that integration failure? not sure if it's related to the change at all.
Thanks.

slandelle:
@oleksiys Yes, you can ignore Travis on 1.9. Travis is enabled globally on the repo but there's no yaml file in 1.9 as tests would fail anyway (some old tests are localized and only work outside NA, where Travis hosts are of course based).

oleksiys:
thank you!

elrodro83:
Excellent. Would it be possible to get a version out with this change?
Thanks in advance.

slandelle:
I juste release 1.9.37

oleksiys:
thank you!

elrodro83:
Thanks to both of you.

PR #1124, opened 2016-03-30
HTTPS requests over HTTP CONNECT proxies should send relative URI

jacobtolar:
If we're sending a HTTPS request and using a proxy, we must be sending over a CONNECT tunnel. In that case, we only need to send the relative URI, not the full URI.

With this change, if I have a proxy configured and want to GET

https://some-url.com/the-path

the server behind some-url.com sees: 

```
GET /the-path
```

Prior to this change the server will see: 

```
GET https://some-url.com/the-path
```

The new behavior matches other clients (e.g. curl).


slandelle:
Thanks!

Issue #1111, opened 2016-03-09
Uppercase URIs result in invalid requests (1.8 only)


gmethvin:
It seems like executing a request to a URI such as `HTTP://EXAMPLE.COM/` will try to send the request to port 443 rather than 80, resulting in various kinds of errors. This was originally reported as a Play Framework issue at playframework/playframework#5820.

I created a project demonstrating the issue: https://github.com/gmethvin/asynchttpclient-uppercase

This seems to be fixed in 1.9 but I'm not sure fixing it was intentional. I didn't see any pull requests or issues mentioning the problem.

slandelle:
1.9 was a major rewrite to pave the way for 2.0, so there's a good chance is just cleaned this up without much thought. Is should be just a matter of `equalsIgnoreCase` instead of `equals`.

I personally won't put any effort in maintaining the 1.8 branch. But feel free to provide a fix and I'll gladly ship and release.

slandelle:
Fixed by #1125.

Released as of 1.8.17

PR #1125, opened 2016-03-31
getPort should work with uppercase scheme

jacobtolar:
Fixes AsyncHttpClient/async-http-client#1111

It appears that the 1.8.x branch:
* Only builds under JDK7
* Has a bunch of flakey tests (like *RemoteSiteTest)


slandelle:
Thanks.

As explained in #1111, 1.8 has reached eol.
I can accept contribs and release, though.

slandelle:
Released as of 1.8.17

PR #1133, opened 2016-04-06
Fix leak when cleaning transactions up

elrodro83:
With the commit https://github.com/AsyncHttpClient/async-http-client/commit/d500dd7351ee93f0c203582cf983e3b131e24968 that was release on 1.9.34, there are cases in which the cleanup is not properly done or not done at all, causing a degradation in performance over time.
This PR addresses that issue.


oleksiys:
Good point!
I've decided to rework the previous fix, will ping you once it's commited.

Thanks a lot!

oleksiys:
please let me know if the fix works for you.
thank you!

elrodro83:
That commit fixed the leak, thanks!

PR #1139, opened 2016-04-07
Added a way to define a body for ReactiveStreams

### LINKED ISSUE #1136: Allow to manually specify a Content-Length on setBody(Publisher<ByteBuffer>)

schmitch:
Will Fix #1136

### DIFF
--- a/client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
+++ b/client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
@@ -308,7 +308,11 @@ public abstract class RequestBuilderBase<T extends RequestBuilderBase<T>> {
     }
 
     public T setBody(Publisher<ByteBuffer> publisher) {
-        return setBody(new ReactiveStreamsBodyGenerator(publisher));
+        return setBody(publisher, -1L);
+    }
+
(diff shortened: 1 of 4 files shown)


client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
```diff
@@ -311,6 +311,10 @@ public T setBody(Publisher<ByteBuffer> publisher) {
         return setBody(new ReactiveStreamsBodyGenerator(publisher));
```
slandelle: delegate to setBody(publisher, -1)

slandelle:
Please address comment and I'll merge.

Thanks!

schmitch:
Done @slandelle and thanks for your fast comments.

PR #1145, opened 2016-04-19
Minor correction to readme code sample

anuragkapur:
AsyncHttpClient is an interface. Instantiate with concrete implementation instead.


slandelle:
Thanks!

PR #1150, opened 2016-05-03
Fix. retry is incremented by twice

charliechang:
There is a code defect in counting retry.

There are two places which can increment the counter.
_NettyResponseFuture.java_
```java
    public boolean canRetry() {
        return maxRetry > 0 && CURRENT_RETRY_UPDATER.incrementAndGet(this) <= maxRetry;
    }
    public boolean canBeReplayed() {
        return !isDone() && canRetry() && !(Channels.isChannelValid(channel) && !getUri().getScheme().equalsIgnoreCase("https")) && !inAuth.get() && !inProxyAuth.get();
    }
```

_NettyConnectListener.java_
```java
        boolean canRetry = future.canRetry(); //Add retry counter here
        LOGGER.debug("Trying to recover from failing to connect channel {} with a retry value of {} ", channel, canRetry);
        if (canRetry//
                && cause != null 
                && (future.getChannelState() != ChannelState.NEW || StackTraceInspector.recoverOnNettyDisconnectException(cause))) {

            if (requestSender.retry(future)) { //Add retry counter here either, since canBeReplayed will be called
                return;
            }
        }
```
The retry counter is added in `future.canRetry()` and `requestSender.retry(future)`
This PR fixed the problem and rename the method from `canRetry` to `incRetryAndCheck` for better naming convention

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
+++ b/client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
@@ -425,7 +425,7 @@ public final class NettyResponseFuture<V> extends AbstractListenableFuture<V> {
         return reuseChannel;
     }
 
-    public boolean canRetry() {
+    public boolean incRetryAndCheck() {
         return maxRetry > 0 && CURRENT_RETRY_UPDATER.incrementAndGet(this) <= maxRetry;
     }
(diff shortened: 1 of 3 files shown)


slandelle:
Thanks!

Issue #119, opened 2012-05-24
Implement exponential backoff


gonzajg:
Could be a great for the project.
Here's an implementation:
http://code.google.com/p/google-http-java-client/wiki/ExponentialBackoff

PR #1155, opened 2016-05-17
Add feature exponential backofff

charliechang:
I implemented the exponential backoff retry and refine the retry mechanism. In this change, retry will happened in one of following conditions.

1. ConnectionException 
Connection failure, e.g. Connection timeout
2. ReadTimeout happened.
When read is timeout, it will trigger retry
3. Channel Inactive
e.g Connection reset.

I also added the test cases for scenarios listed above.


slandelle:
Hi.

This is interesting, thanks!

But sadly, you can't just drop a current behavior that does match your needs for a new one that does.

For example, I mainly develop AHC as the HTTP engine for Gatling, where the desired behavior is the one of a browser:
* no exponential back-off
* retry only happens when a pooled keep-alive connection is closed by the remote peer when trying to write the request

I'm perfectly open to bringing your behavior as long it's not enabled by default and that the current behavior isn't impacted in terms of performance.

charliechang:
Understood. 

Unfortunately we need the retry on read timeout and connection failure due to unstable network condition. We will maintain this change in private repository.

Thanks

nithril:
Hi,

I was looking for pluggable Retry.

> Unfortunately we need the retry on read timeout and connection failure due to unstable network condition. We will maintain this change in private repository.

I think these feature can be pluggable.  You have created a `RetryHandler`, maybe you could add some flags to this handler to enable/disable retry for read timeout/connection failure. 

In my opinion It would be unfortunate to lost this PR.

slandelle:
Then, I'm not sure this feature has to be plugged deep into AHC internals. It could perfectly be implemented as a wrapper.

nithril:
A wrapper around NettyRequestSender ?

slandelle:
Nah, around AHC.

nithril:
My OO design principles bleed ;)

slandelle:
Really? Decorator pattern? AHC is an interface.

nithril:
In my opinion, speaking about a public library, it should be something pluggable, an extension point which will allow different strategy (exponential backoff, failover...) that could be easily implemented, composed, focusing on the feature. A wrapper around AHC is a sledgehammer.  I agree with [https://github.com/AsyncHttpClient/async-http-client/issues/119#issuecomment-18255569](cowtowncoder) comment.

slandelle:
> In my opinion, speaking about a public library, it should be something pluggable, an extension point which will allow different strategy (exponential backoff, failover...) that could be easily implemented, composed, focusing on the feature.

Why not. I'm fine with a contribution as long as:
* the current behavior is still available
* changes are explained/discussed

But I won't do the hard work myself as I simply don't need it.

> A wrapper around AHC is a sledgehammer.

Here, I disagree. It would be quite easy to implement, and you would have fine control.

nithril:
> Here, I disagree. It would be quite easy to implement, and you would have fine control.

I agree, should be understood the same way but with the cons: some fine control not needed and with code not directly linked to the feature.

nithril:
`ResponseFiltersInterceptor` already allows pluggable replay, but only for successful response.

There is some case there pluggable retry is not possible. Usage of 
`NettyRequestSender#abort` are:
- `AsyncHttpClientHandler#exceptionCaught `: possible in case of IOException 
- `HttpHandler#handleRead`: not possible for all error eg. [[1]](https://github.com/AsyncHttpClient/async-http-client/blob/e418d627107e1e797ea4d03e6b93e2a356ee995e/client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java#L180)
- Not possible for 401 and 407 interceptor. May not make sense.
-  `NettyRequestSender#handleUnexpectedClosedChannel` : 
  - `AsyncHttpClientHandler#channelInactive` : possible
  - `NettyRequestSender#sendRequestWithOpenChannel`: not possible
- `NettyRequestSender#sendRequestWithNewChannel` : seems not possible, I do not really understand the purpose of the method, especially with the FIXME and the `ChannelPool`
- `NettyRequestSender#writeRequest` : Seems possible with `channelInactive`
- `TimeoutTimerTask#expire` not possible

To allow retry, these methods may called pluggable interceptors (eg. TimeoutFilter...). To allow custom states, `NettyResponseFuture` may contain a context map.

Does it seems consistent to you?

nithril:
Any feedback?

slandelle:
Too swamped atm, sorry.

Issue #1156, opened 2016-05-17
Netty dependencies of different versions used


Spikhalskiy:
Now async http client uses netty-reactive-streams 1.0.4 which depends on io.netty:netty-handler:4.0.34.Final, while the main netty version in project is 4.0.36.Final

Suggested fix: exclude netty-handler from netty-reactive-streams dependency and add separately with correct version

slandelle:
Thanks!

BTW, versions are properly picked in my env as `netty-code-http` also depends on `netty-handler`, but arrives first in the build chain, before `netty-reactive-streams`. At least that's how maven/aether works, but maybe other tools such as gradle have an issue.

Spikhalskiy:
I found it by:

[INFO] --- maven-enforcer-plugin:1.4:enforce (enforce) @ our_artifact ---
[WARNING]
Dependency convergence error for io.netty:netty-handler:4.0.36.Final paths to dependency are:
+-our_artifact
  +-org.asynchttpclient:async-http-client:2.0.2
    +-io.netty:netty-codec-http:4.0.36.Final
      +-io.netty:netty-handler:4.0.36.Final
and
+-our_artifact
  +-org.asynchttpclient:async-http-client:2.0.2
    +-com.typesafe.netty:netty-reactive-streams:1.0.4
      +-io.netty:netty-handler:4.0.34.Final
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.DependencyConvergence failed with message:
Failed while enforcing releasability the error(s) are [
Dependency convergence error for io.netty:netty-handler:4.0.36.Final paths to dependency are:
+-our_artifact
  +-org.asynchttpclient:async-http-client:2.0.2
    +-io.netty:netty-codec-http:4.0.36.Final
      +-io.netty:netty-handler:4.0.36.Final
and
+-our_artifact
  +-org.asynchttpclient:async-http-client:2.0.2
    +-com.typesafe.netty:netty-reactive-streams:1.0.4
      +-io.netty:netty-handler:4.0.34.Final

slandelle:
OK. So actually, this is more a warning that the dependency tree is not clean (I fully agree), than a bug that actually happen (conflicting versions at runtime).

Spikhalskiy:
Right, no actual bugs based on it. For us, it just caused build fails because of our build flow setup.

Spikhalskiy:
Actually, it didn't fix this warning. With new release it's still presented.
```
[...]

PR #1157, opened 2016-05-17
Fix inherited netty dependency version to 4.0.36.Final

Spikhalskiy:
Suggested patch for Issue #1156

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -260,6 +260,11 @@
 				<artifactId>netty-transport</artifactId>
 				<version>${netty.version}</version>
 			</dependency>
+			<dependency>
+				<groupId>io.netty</groupId>
+				<artifactId>netty-handler</artifactId>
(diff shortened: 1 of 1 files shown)


client/pom.xml
```diff
@@ -43,6 +43,17 @@
...
+				<exclusion>
+					<!--Added behind explicitly with consistent version-->
+					<groupId>io.netty</groupId>
+					<artifactId>netty-handler</artifactId>
+				</exclusion>
+			</exclusions>
+		</dependency>
+		<dependency>
+			<groupId>io.netty</groupId>
+			<artifactId>netty-handler</artifactId>
```
slandelle: Those changes are useless. Dependency management in parent should suffice.

Spikhalskiy: You are right, fixed

PR #1164, opened 2016-05-23
Fix proxy when using JDK provider

malaporte:
Right now it always throws an NPE if it's not bypassed. The culprit is the `ka` variable (that's in fact always null but hey). There is an `if` on the first call to `setRequestProxy`, but none on the 2nd, which is spurious anyway, so I removed it.


slandelle:
Merging. But honestly, you shouldn't be using the JDK provider. It's a mere toy. I advise you switch to the Netty provider, or better upgrade to AHC2.

slandelle:
But thanks anyway :)

malaporte:
Oh I'm using the Netty provider alright, except in this one place where someone requires the use of a proxy server using Windows Authentication (gotta love those guys). There the JDK provider works better than Netty, for some reason.

slandelle:
> There the JDK provider works better than Netty, for some reason.

Even with latest versions (there were several fixes in the area)? Would be great to have the expected and actual HTTP sequences so I can fix (you can guess I can't run such cases on my side).

JDK provider went away with AHC2.

malaporte:
Well I did wanted to test but then I found this:

image

Shoot.

slandelle:
Pity indeed

PR #1167, opened 2016-05-24
always create a mutable collection for the request

efwe:
I'm not sure if the cookies of a request are meant to be modified from within a filter. But if one wants to modify cookies from within a filter it is not possible right now because the emptyList() is immutable.

Please decide if this is a desired behaviour or a bug. I'm happy to check the other lists also then.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
+++ b/client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
@@ -527,7 +527,7 @@ public abstract class RequestBuilderBase<T extends RequestBuilderBase<T>> {
         Charset finalCharset = rb.computeCharset();
 
         // make copies of mutable internal collections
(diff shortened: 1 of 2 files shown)


slandelle:
Hi,

Sorry, but you're not supposed to mutate `Request` instances but use a `RequestBuilder`.
Also, I'm pretty against adding more allocations on a pretty likely path (no cookies).

As `preProcessRequest` returns a `FilterContext` and not void, it means you're supposed to create a new one instead of mutating the input parameter. You can create a new `RequestBuilder` from the current `Request`, add your cookie and create a new `FilterContext` from the associated builder.

efwe:
Understood. Thanks for clarification. Are you positive about clarifying JavaDoc PRs?

slandelle:
Sure, any help on documentation is very welcome!

PR #1169, opened 2016-05-25
Using .equals for String comparison in validateWebSocketRequest

andreaskofer:
Hi, fixing an issue where string comparison is done using "!=" which can cause erratic problems.


slandelle:
Hi,

Thanks for the contribution.
However, I'm sorry but i won't merge.

* The current active version of AsyncHttpClient is 2.0 (latest is 2.0.4).
* The 1.9 branch is only maintained by the community. I'll gladly merge contributed fixes and release but won't fix bugs myself.
* The 1.8 branch is frozen. There won't be any new release. Users are required to upgrade.

I hope you'll understand that keeping too many active branches is harmful for an open-source project.

Regards

Issue #1170, opened 2016-06-01
Allow customizing SslContext after default config is performed


marshallpierce:
I want to customize the trusted CA certs to only trust an internal CA (for service-to-service use). It seems like what I want is to use any of the `SslContextBuilder#trustManager()` methods when building the `SslContext`. However, I'd like to re-use the existing default setup that's done in `DefaultSslEngineFactory`. (Side note: When would `JsseSslEngineFactory` be used? It's not clear what it's for.) Right now my workaround is to copy-and-paste all the setup done in `DefaultSslEngineFactory` and `SslEngineFactoryBase` when creating my own `SslContext`, which isn't ideal.

One way to do this would be to allow customizing the `SslContextBuilder` created in `DefaultSslEngineFactory` right before `build()` is called, maybe by providing a `Function<SslContextBuilder,SslContextBuilder>` or `Consumer<SslContextBuilder>` in `AsyncHttpClientConfig`. This would allow for re-use of the existing config logic but also make it fairly easy to provide a hook that sets up the trust manager. However, it would mean that the config option would be meaningless if a custom SslEngineFactory is provided. I don't know how common that is in practice.

Anyway, I'm happy to provide a PR to get this fixed so I can delete my copy and pasted code...

slandelle:
`JsseSslEngineFactory` takes a `javax.net.ssl.SSLContext` while `DefaultSslEngineFactory` is based on `io.netty.handler.ssl.SslContext`.

> One way to do this would be to allow customizing the SslContextBuilder created in DefaultSslEngineFactory right before build() is called, maybe by providing a Function<SslContextBuilder,SslContextBuilder> or Consumer<SslContextBuilder> in AsyncHttpClientConfig.

Better: have a protected method that you could override and that doesn't break API.

>  However, it would mean that the config option would be meaningless if a custom SslEngineFactory is provided. I don't know how common that is in practice.

I don't get you.

> Anyway, I'm happy to provide a PR to get this fixed so I can delete my copy and pasted code...

Sure, go ahead.

marshallpierce:
- Jsse vs Default: makes sense, thanks.
- Protected method: will do, PR forthcoming.
- Config changes: I was referring to how only the Default factory impl would make use of the config option, which could be confusing to users. But, not an issue with the protected method route.

slandelle:
> Config changes: I was referring to how only the Default factory impl would make use of the config option, which could be confusing to users. But, not an issue with the protected method route.

[...]

PR #1171, opened 2016-06-01
Allow DefaultSslEngineFactory subclass customization of the SslContext

marshallpierce:
See #1170 for context.

If you have ideas for how to usefully test this, I'm happy to write them up, but it wasn't obvious to me how to usefully test this change.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
@@ -111,10 +111,11 @@ public class ChannelManager {
     public ChannelManager(final AsyncHttpClientConfig config, Timer nettyTimer) {
 
         this.config = config;
+        this.sslEngineFactory = config.getSslEngineFactory() != null ? config.getSslEngineFactory() : new DefaultSslEngineFactory();
         try {
-            this.sslEngineFactory = config.getSslEngineFactory() != null ? config.getSslEngineFactory() : new DefaultSslEngineFactory(config);
+            this.sslEngineFactory.init(config);
         } catch (SSLException e) {
-            throw new ExceptionInInitializerError(e);
+            throw new RuntimeException("Could not initialize sslEngineFactory", e);
         }
 
         ChannelPool channelPool = config.getChannelPool();
--- a/client/src/main/java/org/asynchttpclient/netty/ssl/DefaultSslEngineFactory.java
+++ b/client/src/main/java/org/asynchttpclient/netty/ssl/DefaultSslEngineFactory.java
@@ -26,13 +26,9 @@ import org.asynchttpclient.AsyncHttpClientConfig;
 
 public class DefaultSslEngineFactory extends SslEngineFactoryBase {
 
-    private final SslContext sslContext;
+    private volatile SslContext sslContext;
 
-    public DefaultSslEngineFactory(AsyncHttpClientConfig config) throws SSLException {
-        this.sslContext = getSslContext(config);
-    }
-
-    private SslContext getSslContext(AsyncHttpClientConfig config) throws SSLException {
+    private SslContext buildSslContext(AsyncHttpClientConfig config) throws SSLException {
         if (config.getSslContext() != null)
             return config.getSslContext();
 
@@ -44,7 +40,7 @@ public class DefaultSslEngineFactory extends SslEngineFactoryBase {
         if (config.isAcceptAnyCertificate())
             sslContextBuilder.trustManager(InsecureTrustManagerFactory.INSTANCE);
 
-        return sslContextBuilder.build();
+        return configureSslContextBuilder(sslContextBuilder).build();
     }
 
     @Override
@@ -54,4 +50,22 @@ public class DefaultSslEngineFactory extends SslEngineFactoryBase {
         configureSslEngine(sslEngine, config);
         return sslEngine;
     }
+
+    @Override
+    public void init(AsyncHttpClientConfig config) throws SSLException {
+        sslContext = buildSslContext(config);
+    }
+
+    /**
+     * The last step of configuring the SslContextBuilder used to create an SslContext when no context is provided in
+     * the {@link AsyncHttpClientConfig}. This defaults to no-op and is intended to be overridden as needed.
+     *
+     * @param builder builder with normal configuration applied
+     * @return builder to be used to build context (can be the same object as the input)
+     */
+    protected SslContextBuilder configureSslContextBuilder(SslContextBuilder builder) {
+        // default to no op
+        return builder;
+    }
+
 }
--- a/client/src/main/java/org/asynchttpclient/SslEngineFactory.java
+++ b/client/src/main/java/org/asynchttpclient/SslEngineFactory.java
@@ -14,6 +14,7 @@
 package org.asynchttpclient;
 
 import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLException;
 
 public interface SslEngineFactory {
 
@@ -26,4 +27,16 @@ public interface SslEngineFactory {
      * @return new engine
      */
     SSLEngine newSslEngine(AsyncHttpClientConfig config, String peerHost, int peerPort);
+
+    /**
+     * Perform any necessary one-time configuration. This will be called just once before {@code newSslEngine} is called
+     * for the first time.
+     *
+     * @param config the client config
+     * @throws SSLException if initialization fails. If an exception is thrown, the instance will not be used as client
+     *                      creation will fail.
+     */
+    default void init(AsyncHttpClientConfig config) throws SSLException {
+        // no op
+    }
 }


client/src/main/java/org/asynchttpclient/netty/ssl/DefaultSslEngineFactory.java
```diff
@@ -26,13 +26,11 @@
...
-    private final SslContext sslContext;
+    private volatile SslContext sslContext;
 
-    public DefaultSslEngineFactory(AsyncHttpClientConfig config) throws SSLException {
-        this.sslContext = getSslContext(config);
-    }
+    public DefaultSslEngineFactory() { }
 
-    private SslContext getSslContext(AsyncHttpClientConfig config) throws SSLException {
+    private SslContext buildSslContext(AsyncHttpClientConfig config) throws SSLException {
```
marshallpierce: I left this code where it was in the source but if you don't mind the reshuffling, I'll move it down to the bottom with the other private stuff

client/src/main/java/org/asynchttpclient/netty/ssl/DefaultSslEngineFactory.java
```diff
@@ -44,14 +42,44 @@ private SslContext getSslContext(AsyncHttpClientConfig config) throws SSLExcepti
...
     }
+
+    /**
+     * The last step of configuring the SslContextBuilder used to create an SslContext when no context is provided in
+     * the {@link AsyncHttpClientConfig}. This defaults to no-op and is intended to be overridden as needed.
+     *
+     * @param builder builder with normal configuration applied
+     * @return builder to be used to build context (can be the same object as the input)
+     */
+    protected SslContextBuilder configureSslContextBuilder(SslContextBuilder builder) {
```
marshallpierce: I made this return `SCB` instead of `void` to allow someone to make a new one if they *really* needed to

client/src/main/java/org/asynchttpclient/netty/ssl/DefaultSslEngineFactory.java
```diff
@@ -44,14 +42,44 @@ private SslContext getSslContext(AsyncHttpClientConfig config) throws SSLExcepti
...
+        // default to no op
+        return builder;
+    }
+
+    private SslContext getSslContext(AsyncHttpClientConfig config) {
+        SslContext localContext = sslContext;
```
marshallpierce: As per http://shipilev.net/blog/2014/safe-public-construction/#_safe_publication

slandelle:
Headache... Actually, it bothers me if we only detect SslContext building crash on first request instead of client instantiation...

marshallpierce:
Yeah, that's really not very nice at all. :( If breaking compatibility is out of the question, one thing we could do is add a pre-emptive initialization call in ChannelManager in the typical case where the factory is the default impl (either a custom subclass or not).

Perhaps at the next opportunity to break compatibility we could add some sort of `init()` to the factory interface. This would allow us to cleanly throw from such initialization errors, and also remove the double-checked locking and have simply a volatile.

marshallpierce:
Oh wait, this project is Java 8 -- couldn't we just add an `init(config)` with a default no-op impl to the factory interface?

slandelle:
Oh right! Can you give it a try?

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -110,10 +110,11 @@
     public ChannelManager(final AsyncHttpClientConfig config, Timer nettyTimer) {
 
         this.config = config;
+        this.sslEngineFactory = config.getSslEngineFactory() != null ? config.getSslEngineFactory() : new DefaultSslEngineFactory();
         try {
-            this.sslEngineFactory = config.getSslEngineFactory() != null ? config.getSslEngineFactory() : new DefaultSslEngineFactory(config);
+            this.sslEngineFactory.init(config);
         } catch (SSLException e) {
-            throw new ExceptionInInitializerError(e);
+            throw new RuntimeException("Could not initialize sslEngineFactory", e);
```
marshallpierce: I thought this was a more helpful exception

client/src/main/java/org/asynchttpclient/netty/ssl/DefaultSslEngineFactory.java
```diff
@@ -26,13 +26,11 @@
 
 public class DefaultSslEngineFactory extends SslEngineFactoryBase {
 
-    private final SslContext sslContext;
+    private volatile SslContext sslContext;
```
marshallpierce: To be able to make this final we'd need some sort of FactoryFactory that could create a Factory from the Config. I'm not heartbroken over ~2ns volatile read cost per ssl connection though.

slandelle: Yeah, I have been trying to reject this FactoryFactory idea since yesterday and come with another solution. I've failed so far :(

marshallpierce:
btw, I can't run tests on my Linux box  - `BasicHttpProxyToHttpTest.nonPreemptyProxyAuthWithPlainHttpTarget` consistently hangs. It works on my OS X laptop. Is this a known issue?

slandelle:
> BasicHttpProxyToHttpTest.nonPreemptyProxyAuthWithPlainHttpTarget consistently hangs. It works on my OS X laptop. Is this a known issue? It happens in untouched upstream as well as my fork.

Oh my... I run on OSX and was trusting Travis to execute the tests on Linux, but I just realized that Travis skip tests by default!!! Investigating.

slandelle:
Actually, there was an issue with the Travis set up, tests are currently running: https://travis-ci.org/AsyncHttpClient/async-http-client/builds/134807126

marshallpierce:
FWIW I get the following on a full run:

```

Failed tests:
  BasicHttpProxyToHttpTest.nonPreemptyProxyAuthWithPlainHttpTarget:119 expected [200] but found [504]
  ProxyTest.testRequestNonProxyHost:176 » IllegalState failed to create a child ...
  ReactiveStreamsTest.testConnectionDoesNotGetClosed:66 arrays differ firstly at element [12000]; expected value is <111> but was <45>.
```

The last one only shows up sometimes. The second one shows up every time I do a full run, but doesn't fail when run by itself.

slandelle:
How do you run maven? By any chance, do you use parallel builds?

marshallpierce:
`mvn clean install`. No settings.xml. No parallel builds, version 3.3.9.

marshallpierce:
I'll see if I can reproduce this in a VM.

slandelle:
I'll investigate tomorrow at the office too. Those issues are very weird. The `IllegalStateException` usually means connectivity issues or ports starvation.

marshallpierce:
In case it comes in handy, I pushed https://github.com/marshallpierce/async-http-client/tree/debug-test-failures-vagrant which has a vagrant box. Curiously, I get different issues in the VM:

```
Running TestSuite
Tests run: 395, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 134.372 sec <<< FAILURE! - in TestSuite
testListenableFutureAfterCompletion(org.asynchttpclient.ListenableFutureTest)  Time elapsed: 0.028 sec  <<< FAILURE!
java.lang.AssertionError: expected [1] but found [0]
        at org.testng.Assert.fail(Assert.java:94)
        at org.testng.Assert.failNotEquals(Assert.java:496)
        at org.testng.Assert.assertEquals(Assert.java:125)
        at org.testng.Assert.assertEquals(Assert.java:372)
        at org.testng.Assert.assertEquals(Assert.java:382)
        at org.asynchttpclient.ListenableFutureTest.testListenableFutureAfterCompletion(ListenableFutureTest.java:63)


```

slandelle:
WAT?!

I only see 2 changes that could be related:
* I modified `AbstractListenableFuture` to use the same design as your PR for safe publication, see 2addbc04cb94c59c67b94101cb305e2fe72979b7 (you don't have this commit in your fork).
* I updated our internal copy of `ExecutionList` (which comes from Guava), see f4f6129ce90b8dcb67c637fd4b60a67d5cdeeab7

Could you try adding the former AND/OR reverting the latter, please?

I'll try to reproduce on my side today.

slandelle:
I could reproduce the 504 on Travis: https://travis-ci.org/AsyncHttpClient/async-http-client#L5629

slandelle:
I could get the Travis build to pass if I ask surefire to not fork and execute tests sequentially, see 8c054a3ae9142933aee1bf70fb2045f9daeab0b3 and https://travis-ci.org/AsyncHttpClient/async-http-client/builds/134987736.

And I can't reproduce the issue on my local Linux box. Will try your vagrant.

slandelle:
I can't reproduce on VirtualBox either :(

And actually build failed again on Travis... I'm at a loss.
What's your host config?

slandelle:
>   ProxyTest.testRequestNonProxyHost:176 » IllegalState failed to create a child ...
shows up every time I do a full run, but doesn't fail when run by itself.

I would be awesome if you could get the full stacktrace. In order to get surefire to output logs, you have to enable a profile: `mvn test -Ptest-output`.

This message comes from Netty, and I think the root issue is there, like lacking/leaking native memory.

slandelle:
Pinned it: #1172

marshallpierce:
I'll dig in once I'm at the office (about an hour). The host is an E5-1650 v3 w/ 64GiB, running up-to-date Gentoo Linux with Oracle JDK 1.8u92. It's not running with SELinux or restrictive sysctl params or anything like that.

slandelle:
Check last comment: issue solved :)
https://github.com/AsyncHttpClient/async-http-client/commit/8951a89621fc978fa8227cec24bd15e48f994ea6#diff-09326dc85cb5946ac041536f7475c385R90

slandelle:
(cont)

At least the `BasicHttpProxyToHttpTest`.
I couldn't figure out exactly where the lock was as I could only reproduce on Travis, where I can't of course make a thread dump. But tests are all green on Travis now.

I couldn't reproduce the other ones.

marshallpierce:
On 9f817eb I get: 

```
Failed tests: 
  RemoteSiteTest.stripQueryStringTest:164 expected [200] but found [503]
  ProxyTest.testRequestNonProxyHost:176 » IllegalState failed to create a child ...
```

Stack traces:

```
 testRequestNonProxyHost
java.lang.IllegalStateException: failed to create a child event loop
	at io.netty.util.concurrent.MultithreadEventExecutorGroup.<init>(MultithreadEventExecutorGroup.java:68)
	at io.netty.channel.MultithreadEventLoopGroup.<init>(MultithreadEventLoopGroup.java:49)
	at io.netty.channel.nio.NioEventLoopGroup.<init>(NioEventLoopGroup.java:68)
	at io.netty.channel.nio.NioEventLoopGroup.<init>(NioEventLoopGroup.java:63)
	at io.netty.channel.nio.NioEventLoopGroup.<init>(NioEventLoopGroup.java:54)
	at org.asynchttpclient.netty.channel.ChannelManager.<init>(ChannelManager.java:173)
	at org.asynchttpclient.DefaultAsyncHttpClient.<init>(DefaultAsyncHttpClient.java:81)
	at org.asynchttpclient.DefaultAsyncHttpClient.<init>(DefaultAsyncHttpClient.java:63)
	at org.asynchttpclient.Dsl.asyncHttpClient(Dsl.java:25)
	at org.asynchttpclient.proxy.ProxyTest.testRequestNonProxyHost(ProxyTest.java:176)
Caused by: io.netty.channel.ChannelException: failed to open a new selector
	at io.netty.channel.nio.NioEventLoop.openSelector(NioEventLoop.java:143)
	at io.netty.channel.nio.NioEventLoop.<init>(NioEventLoop.java:134)
	at io.netty.channel.nio.NioEventLoopGroup.newChild(NioEventLoopGroup.java:94)
	at io.netty.util.concurrent.MultithreadEventExecutorGroup.<init>(MultithreadEventExecutorGroup.java:64)
	... 37 more
Caused by: java.io.IOException: Too many open files
	at sun.nio.ch.IOUtil.makePipe(Native Method)
	at sun.nio.ch.EPollSelectorImpl.<init>(EPollSelectorImpl.java:65)
	at sun.nio.ch.EPollSelectorProvider.openSelector(EPollSelectorProvider.java:36)
	at io.netty.channel.nio.NioEventLoop.openSelector(NioEventLoop.java:141)
	... 40 more
... Removed 28 stack frames
```

and

```
 stripQueryStringTest
java.lang.AssertionError: expected [200] but found [503]
	at org.asynchttpclient.RemoteSiteTest.stripQueryStringTest(RemoteSiteTest.java:164)
... Removed 33 stack frames
```

The first one I cannot reproduce when running by itself. The latter one I did on the third try. However, that one appears to be hitting the live freakonomics.com site, so it's probably unavoidable without reimplementing using a local http server.

`ulimit -n` reports `1024` on my Linux box, which is pretty typical. I can certainly make it a massive number to make the tests pass, but the fact that it fails in an all-tests run and passes by itself makes your resource leak theory seem pretty likely. Some casual `lsof`ing while running tests does show a *lot* of pipes allocated. Annoyingly, setting `ulimit -n 1024` on OS X doesn't make the build fail.

marshallpierce:
Oh, apparently the hard limit on OS X is unlimited anyway.

Re-running on linux with `ulimit -n 4096` still fails on testRequestNonProxyHost. `lsof`ing for pipes owned by `java` showed counts as high as 36000 during the test run, and was typically above 10000, so clearly something is unhappy.

slandelle:
> The first one I cannot reproduce when running by itself. The latter one I did on the third try. However, that one appears to be hitting the live freakonomics.com site, so it's probably unavoidable without reimplementing using a local http server.

Yeah, some tests where originally introduced this way (before I joined here). There are some issues open for refactoring them into proper standalone one, such as #496. Help welcome. The "freakonomics.com" website is sometimes unstable.

> ulimit -n reports 1024 on my Linux box, which is pretty typical. I can certainly make it a massive number to make the tests pass, but the fact that it fails in an all-tests run and passes by itself makes your resource leak theory seem pretty likely. 

Maybe, but not necessarily. It could be that we have tons of ports that have been opened, but can't be recycled yet.
Default ulimit (1024) is a value suited for desktop, not server.

> Some casual lsofing while running tests does show a lot of pipes allocated.

Sure:
* every test class starts one or two embedded Jetty (hence server ports)
* every test method start one client (hence one client port per connection)

slandelle:
lsof accounts for all open files, not only sockets. What does netstat says?

marshallpierce:
Is there anything in particular you'd like to look at in netstat output? I don't see anything other than a hundred TIME_WAIT sockets, which since they've already been closed are not associated with the opening process. In my lsof'ing I'm grepping for this (with all other java processes closed):

```
lsof | grep '^java' | grep pipe | wc -l
```

Or this:

```
lsof | grep '^java' | grep eventpoll | wc -l
```

I can't think of a good reason why that should output 121256 eventpoll fds, or 85741 pipes, and yet I've seen those numbers. Even if the sockets were persisting after being properly closed, they shouldn't be tied to that pid anymore.

marshallpierce:
I also found one of these in the test log:

```
2016-06-03 08:52:35,827 [Finalizer] ERROR org.asynchttpclient.DefaultAsyncHttpClient - AsyncHttpClient.close() hasn't been invoked, which may produce file descriptor leaks
```

So, we've got at least one leak.

marshallpierce:
I stuck a Thread.sleep(Long.MAX_VALUE) at the top of testRequestNonProxyHost to give me some time to investigate. After dumping the (live only) heap with jmap, some cursory `jhat` tells me that there are 800 live instances of `sun.nio.ch.EPollSelectorImpl`, the class whose ctor fails with the ISE. That seems wrong for a test. 

When I run just that test alone and take a live heap dump while it's sleeping at the beginning of the test, I see 8 instances, so it seems likely that the other 792 are leaked from previous tests.

slandelle:
> 2016-06-03 08:52:35,827 [Finalizer] ERROR org.asynchttpclient.DefaultAsyncHttpClient - AsyncHttpClient.close() hasn't been invoked, which may produce file descriptor leaks

This was probably a derived client no being closed: https://github.com/AsyncHttpClient/async-http-client/commit/f402385eec2789275b7f1ca79de2508b3bb50a98#diff-8bec36202922bc4718c8cabbdf131ea0L276.
False positive.

I did found tons of clients not being closed but in an extra module that's build downstream.

marshallpierce:
I've rebased this on top of latest master. Also, if travis will let you, would be interesting to add 'ulimit -H -n 4000' so we have a known config there. The build passes with 4000 for me. (The cutoff appears to be between 3500 and 4000.)

slandelle:
Done

marshallpierce:
Is there anything else you'd like me to improve on this, or are you happy with the current approach?

client/src/main/java/org/asynchttpclient/netty/ssl/DefaultSslEngineFactory.java
```diff
@@ -26,13 +26,11 @@
...
 
-    public DefaultSslEngineFactory(AsyncHttpClientConfig config) throws SSLException {
-        this.sslContext = getSslContext(config);
-    }
+    public DefaultSslEngineFactory() { }
```
slandelle: Please remove explicit default constructor.

slandelle:
Could you please address last comment and squash?

slandelle:
Thanks!

slandelle:
@marshallpierce I've just release 2.0.5

marshallpierce:
2.0.5 is working great with a custom subclass of DSEF; thanks!

slandelle:
Great! Thanks for your help!

PR #1182, opened 2016-06-07
Specify a version for javadoc plugin to appease Maven

marshallpierce:
I got tired of seeing the warning messages every build. This is the most recent version, and it only appears to be used in `netty-bp`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,6 +150,14 @@
 				</executions>
 			</plugin>
 		</plugins>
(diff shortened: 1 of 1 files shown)


slandelle:
Thanks! Those modules are just a copy of Netty ones that are only available for Netty 4.1 while we still use Netty 4.0. I just copied the poms without noticing the missing version.

PR #1187, opened 2016-06-09
Add  CompletableFuture usage example to Readme.md

fmcarvalho:
Include a simple use of AsyncHttpClient with CompletableFuture in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,24 @@ asyncHttpClient.prepareGet("http://www.example.com/").execute(new AsyncCompletio
 
 (this will also fully read `Response` in memory before calling `onCompleted`)
 
+Alternatively you may use continuations (through Java 8 class `CompletableFuture<T>`) to accomplish asynchronous (non-blocking) solution. The equivalent continuation approach to the previous example is:
+
+```java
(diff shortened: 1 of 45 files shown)


slandelle:
Hi,

Thanks for contributing.
However, I think it would be way better to contribute examples into a dedicated module inside this repository (that's [what Netty does](https://github.com/netty/netty/tree/4.1/example) for example) than to have a link pointing to your personal repo. Do you agree?

Cheers

fmcarvalho:
Sorry for my late response. I agree and I will create that module. Following the naming conventions from [Netty example](https://github.com/netty/netty/tree/4.1/example) I am thinking to put this usage demo into a class `Continuations` in a package `org.asynchttpclient.example.continuations`. Is this ok? 

I am not sure about using the name `Continuations` or `Promise` or any other?

slandelle:
Why not completable? It would be explicit that it's about CompletableFutures and not ReactiveStreams or something else.

fmcarvalho:
Do you mean `asynchttpclient.example.completable.CompletableFutures`?

slandelle:
yep

Issue #1184, opened 2016-06-09
A means to keep connections alive, even if fewer connections would suffice


tomfitzhenry:
### Usecase
I make 500req/s to the same server, whose response times average 5ms. This only needs 3 connections. This 500req/s occasionally bursts to 1000req/s, but as I only have 3 connections in my connection pool, I then have to establish a bunch of connections (let's say 17), at which point the latency suffers the cost of a TCP/SSL handshake.

The rate then drops to the baseline 500req/s, but next time the 1000req/s burst occurs, I'll be ready right?! I have 20 connections in my pool, waiting for me to need them.

No, I won't be ready. By the time the burst occurs, these extra connections have timed out, because they've not been used, and these connections are subject to TCP and HTTP time outs (both beyond my control).

It would be nice to have a way to keep these connections active, so they don't idle out, ready for when I next see a burst of 1000req/s.

### A possible implementation
A simple implementation would be to lease connections on a FIFO basis, rather than a LIFO basis (async-http-client's current behaviour). With a FIFO policy, all connections would be kept active, so they would never idle out.

async-http-client's LIFO policy comes from DefaultChannelPool using a Deque, and its pollFirst and offerFirst methods: lines 250 and 269 of https://github.com/AsyncHttpClient/async-http-client/blob/6627c53a67af3ad0c48dc3e7ef96d7d153899ff0/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java .

Other HTTP clients:
* RxJava uses a Queue (FIFO): https://github.com/ReactiveX/RxJava/blob/2a8d6c75205b13cd9bef2b7f7d47723e1c86454e/src/main/java/rx/internal/util/ObjectPool.java#L28
* Vert.x uses a Queue (FIFO): https://github.com/eclipse/vert.x/blob/dbdcb3c785e9d8ff70dcce6a2e148cd3da6d15f6/src/main/java/io/vertx/core/http/impl/Http1xPool.java#L48
* Apache HttpAsyncClient uses a LinkedList, in LIFO mode: lines 83 and 129 of https://github.com/apache/httpcore/blob/4.4.x/httpcore-nio/src/main/java/org/apache/http/nio/pool/RouteSpecificPool.java#L83

slandelle:
The benefit of using a FIFO instead of a LIFO really depends on your use case. 
Actually, we originally used a FIFO.
[...]

PR #1188, opened 2016-06-13
Allow user to specify whether pool is LIFO or FIFO

tomfitzhenry:
Issue: #1184

What do you think about whether this should be configurable via ahc.properties? I initially did this, but I wondered whether there might be implementations of ChannelPool for which LIFO vs FIFO didn't make sense, and by putting LIFO vs FIFO at properties file level, that might be a leaky abstraction.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
@@ -21,10 +21,7 @@ import io.netty.util.Timeout;
 import io.netty.util.Timer;
 import io.netty.util.TimerTask;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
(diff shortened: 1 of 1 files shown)


slandelle:
> What do you think about whether this should be configurable via ahc.properties?

Usage would be very specific/limited as it could only tune the default `DefaultChannelPool`.
I think this version is sufficient for a first version.

Let's see if users need more (meaning that they would need to be aware of this feature, meaning documentation, which is this project's weakness).

WDYT?

bbc-tomfitzhenry:
I added this wiki page: https://github.com/AsyncHttpClient/async-http-client/wiki/Connection-pooling

Suggestions welcome!

slandelle:
@bbc-tomfitzhenry Awesome, thanks!

Some comments:
* You actually have a `AsyncHttpClientConfig`, not a builder, as you call `build`
* You could use the helpers from `org.asynchttpclient.Dsl` (use a static import)
* Beware that if you're having your own timer, you're in charge of shutting it down

tomfitzhenry:
Fixed docs to use `org.asynchttpclient.Dsl`. Didn't know about that!

slandelle:
Thanks!

> Didn't know about that!

New in AHC2. When I'm telling you doc is lacking ;-)

PR #1208, opened 2016-07-12
Update Grizzly to latest version and add test case for HTTPS close co…

afelisatti:
…nnection case


afelisatti:
@slandelle Could you take a look at this? I think it'd be great if the 1.9.39 release includes this upgrade.

afelisatti:
Thanks!

Issue #1212, opened 2016-07-17
[AHC 2] ByteBuffer Response wrong position 


marenzo:
Calling `Response#getResponseBodyAsByteBuffer` will return a new ByteBuffer but the current position will be the end of the buffer, rather than the start, causing to rewind manually the buffer.
The allocated byte buffer should have the position set to zero.

Problematic code is found at `NettyResponse#getResponseBodyAsByteBuffer`.

Just a side note: previous client returned the byte buffer with position zero.

slandelle:
Closed by #1213

PR #1213, opened 2016-07-17
Rewinding NettyResponse's BodyPart ByteBuffer

marenzo:
Related to #1212

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
+++ b/client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
@@ -179,6 +179,7 @@ public class NettyResponse implements Response {
         for (HttpResponseBodyPart part : bodyParts)
             target.put(part.getBodyPartBytes());
 
+        target.flip();
         return target;
     }
 


client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
```diff
@@ -179,6 +179,7 @@ public ByteBuffer getResponseBodyAsByteBuffer() {
         for (HttpResponseBodyPart part : bodyParts)
             target.put(part.getBodyPartBytes());
 
+        target.rewind();
```
slandelle: Why not flip?

marenzo: does it matter? I don't mind to change to flip, but it's doesn't do anything special in this context - it's updates the state of limit to be the same as already defined (which is the buffer length), while rewind just sets back to the index to the beginning.

slandelle: Well, `flip` is the idiomatic way for switching from write to read and vice versa, which is exactly what we need here.

marenzo: done.

slandelle:
Thanks!

PR #1222, opened 2016-07-28
Upgrade Netty 3.10.6.Final

johnou:


johnou:
@slandelle thanks!

slandelle:
Thanks Johno!

PR #1227, opened 2016-08-11
Cleanup IDE warnings

doom369:
Motivation : 

IDE highlights a lot of warnings, code hard to read with them.

Modifications : 

Removed unnecessary explicit type arguments, `final` for private methods, replaces `for` with `for each`.

Result : 

Cleaner code.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/cookie/CookieEncoder.java
+++ b/client/src/main/java/org/asynchttpclient/cookie/CookieEncoder.java
@@ -63,8 +63,7 @@ public final class CookieEncoder {
         } else {
             Cookie[] cookiesSorted = cookies.toArray(new Cookie[cookies.size()]);
             Arrays.sort(cookiesSorted, COOKIE_COMPARATOR);
-            for (int i = 0; i < cookiesSorted.length; i++) {
-                Cookie cookie = cookiesSorted[i];
+            for (Cookie cookie : cookiesSorted) {
                 if (cookie != null) {
                     add(sb, cookie.getName(), cookie.getValue(), cookie.isWrap());
                 }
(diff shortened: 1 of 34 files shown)


client/src/main/java/org/asynchttpclient/cookie/CookieEncoder.java
```diff
@@ -63,8 +63,7 @@ public static String encode(Collection<Cookie> cookies) {
         } else {
             Cookie[] cookiesSorted = cookies.toArray(new Cookie[cookies.size()]);
             Arrays.sort(cookiesSorted, COOKIE_COMPARATOR);
-            for (int i = 0; i < cookiesSorted.length; i++) {
-                Cookie cookie = cookiesSorted[i];
+            for (Cookie cookie : cookiesSorted) {
```
slandelle: Actually, I would tend to disagree here: the original form saves an Iterator. One could argue JIT escape analysis might be able to save it, but maybe not. Netty does the same thing.

slandelle:
LGTM except for one small nit. Could you please revert this one single change?

client/src/main/java/org/asynchttpclient/cookie/CookieEncoder.java
doom369: @slandelle what iterator do you mean? This is array, `for each` in arrays don't use iterator. Compiler extract this loop into your version, It has nothing to do with JIT. I can show you bytecode if you want :).

slandelle: Oh, right!

slandelle:
Thanks!

PR #1228, opened 2016-08-11
Cleanup : duplicated netty logic removed

doom369:
Motivation : 

Currently async http client creates own Epoll and Nio channel factories while netty has this logic inside Bootstrap class.

Modifications : 

Removed async http client ChannelFactories, now async http client relies on netty during factories initialization.

Result : 

Less duplicated code.


slandelle:
You missed the point of ChannelFactory: the netty-native-transport in an *optional* dependency, and it's only available on Linux. Here, your code would crash with a `ClassNotFoundException`.

PR #1248, opened 2016-09-12
Fix #1247: Problem with custom headers in ByteArrayPart

### LINKED ISSUE #1247: Problem with custom headers in ByteArrayPart
First, when adding a new header with addCustomHeader("name", "value"), one would expect "name: value" to be appended to the output but instead just "namevalue" is used. This forces the developer to arbitrarily add the separator inside the name or the value.

Second, some headers passed in the constructor aren't recognized in addCustomHeader. For example addCustomHeader("Content-Type", "text/plain") doesn't replace the corresponding header but adds another one. One could check for each of these special ones but programatically it's difficult to use.

Third, and most importantly, it's not possible to completely control the Content-Disposition header. It always tries to build it from 2 parts (dispositionType and name), and to get exactly a string one has to include everything in the first and set the second one to null.

These issues will be addressed in a PR.

alepulver:


src/main/java/com/ning/http/client/multipart/Part.java
```diff
@@ -12,13 +12,13 @@
...
 
-import static java.nio.charset.StandardCharsets.US_ASCII;
-
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.channels.WritableByteChannel;
 import java.nio.charset.Charset;
 
+import static java.nio.charset.StandardCharsets.US_ASCII;
+
```
slandelle: Could you please revert this? This change is unrelated to your fix and is probably due to different IDE settings.

src/main/java/com/ning/http/client/multipart/PartBase.java
```diff
@@ -23,6 +20,9 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import static com.ning.http.util.MiscUtils.isNonEmpty;
+import static java.nio.charset.StandardCharsets.US_ASCII;
+
```
slandelle: Please revert unrelated change

src/main/java/com/ning/http/client/multipart/PartBase.java
```diff
@@ -254,14 +277,46 @@ public void setDispositionType(String dispositionType) {
...
+    }
+
     public void addCustomHeader(String name, String value) {
         if (customHeaders == null) {
             customHeaders = new ArrayList<Param>(2);
         }
-        customHeaders.add(new Param(name, value));
+
+        // correctly handle special headers
+        if (name.toLowerCase().equals(CONTENT_TYPE.toLowerCase())) {
```
slandelle: Use equalsIgnoreCase instead of allocating 2 additional Strings.

src/main/java/com/ning/http/client/multipart/PartBase.java
```diff
@@ -254,14 +277,46 @@ public void setDispositionType(String dispositionType) {
...
+            setCustomContentType(value);
+        } else if (name.toLowerCase().equals(CONTENT_TRANSFER_ENCODING.toLowerCase())) {
```
slandelle: equalsIgnoreCase

src/main/java/com/ning/http/client/multipart/PartBase.java
```diff
@@ -254,14 +277,46 @@ public void setDispositionType(String dispositionType) {
...
+            transferEncoding = value;
+        } else if (name.toLowerCase().equals(CONTENT_ID.toLowerCase())) {
```
slandelle: equalsIgnoreCase

src/main/java/com/ning/http/client/multipart/PartBase.java
```diff
@@ -254,14 +277,46 @@ public void setDispositionType(String dispositionType) {
...
+            contentId = value;
+        } else if (name.toLowerCase().equals(CONTENT_DISPOSITION.toLowerCase())) {
```
slandelle: equalsIgnoreCase

src/main/java/com/ning/http/client/multipart/AbstractFilePart.java
```diff
@@ -66,7 +66,7 @@ public AbstractFilePart(String name, String contentType, Charset charset, String
 
     protected void visitDispositionHeader(PartVisitor visitor) throws IOException {
         super.visitDispositionHeader(visitor);
-        if (fileName != null) {
+        if (!hasCustomContentDisposition() && fileName != null) {
```
wissamalwakeel: why not use StringUtils.isNotBlank(fileName) instead of "fileName != null" to make sure that you don't get an empty string as well.

PR #1250, opened 2016-09-18
Enable asynchronous cancellation of AsyncHttpSingle

twz123:
AsyncHttpSingle is currently not forwarding an unsubscription to AsyncHttpClient, it just aborts request processing when an AsyncHandler callback method is invoked. To actually eagerly cancel a request on unsubscription, use the request future to actually forward the cancellation to AsyncHttpClient.


slandelle:
What is `AsyncHttpSingle`?

slandelle:
Thanks!

PR #1260, opened 2016-10-01
Demote log level on trivial 401 intercept

milliburn:
The `Unauthorized401Interceptor` calls the noisy `LOGGER.info()` where trivial 401 responses are involved, i.e. no cause for the interceptor to kick in.

Demotes that particular log message to `debug`, keeping the remaining messages at `info` to retain diagnostic value.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/handler/intercept/ProxyUnauthorized407Interceptor.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/intercept/ProxyUnauthorized407Interceptor.java
@@ -68,7 +68,7 @@ public class ProxyUnauthorized407Interceptor {
         Realm proxyRealm = future.getProxyRealm();
 
         if (proxyRealm == null) {
-            LOGGER.info("Can't handle 407 as there's no proxyRealm");
+            LOGGER.debug("Can't handle 407 as there's no proxyRealm");
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


slandelle:
> noisy LOGGER.info()

This is a mere opinion.
Some people could say that experiencing a 401 with no Realm configured is a misusage and should be logged in error.
Please elaborate with proper arguments.

milliburn:
- The [Realm](https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/java/org/asynchttpclient/Realm.java) supports only a subset of possible authentication protocols.
- In these unsupported cases the appropriate credentials would be passed through direct manipulation of the request.
- This particular log line applies whenever the RequestBuilder (or client) has not been _intentionally_ configured with a Realm. Intentional cases remain unaffected in terms of diagnostic output.
- For example, Realm does not cover the widely used [OAuth2 Bearer Token](https://tools.ietf.org/html/rfc6750). If the credentials in use have been revoked, the server is instructed to [respond 401](https://tools.ietf.org/html/rfc6750#page-9), which is not a challenge response.

slandelle:
If I get it right, you mostly complain on this log using info level because you implement quite standard auth schemes on your own that are not currently supported out of the box. If that's the case, why not contribute them instead?

milliburn:
> [...] you implement quite standard auth schemes [...] If that's the case, why not contribute them instead?

More auth protocols[1] will exist than the Realm can reasonably support. 

Even in cases where the Realm does happen to support a particular protocol, the request-building code might choose to implement the protocol manually; for example, to work around some intricacies of the target service[2], or simply to ensure complete control over the content.

[1] Or variations thereof: for example, OAuth2 services often consume a query string parameter rather than header.
[2] Specs are often loosely interpreted; for example, a service might omit an implicit "Bearer" portion.

slandelle:
@milliburn Could you also please address `ProxyUnauthorized407Interceptor`?

milliburn:
Pushed and squashed.

PR #1262, opened 2016-10-04
Fixes #1261 Grizzly Provider HttpTransactionContext.getAsyncHandler()…

### LINKED ISSUE #1261: Grizzly Provider HttpTransactionContext.getAsyncHandler() does not handle situation where future is null.
In various scenarios the future in HttpTransactionContext is null. While HttpTransactionContext exposes a getter for AsyncHandler, if does not handle the case where the future is null.

dfeist:
… does not handle situation where future is null.


dfeist:
@oleksiys Any change this can be merged in so it's in next release of 1.9.x?

Issue #613, opened 2014-07-11
Listen to WebSocket write failures


johnoliver:
As discussed on the Google Group, there is currently no way of detecting if a write to a WebSocket fails or not. Both Grizzly and Netty return Futures from their write operations that could be exposed in the API to allow the client code to listen for failed writes.

slandelle:
Renaming as the issue is not specific to WebSockets.

dguggemos:
Any news on this issue? Is it planned to be implemented? This would not only be useful to react on failures but also to get some feedback if the web socket accepts more messages. Currently messages are queued until we run out of memory when publishing at high rate and we have no chance to know if a message is really on the wire.

slandelle:
No, nothing planned atm. Contribs welcome.

PR #1276, opened 2016-10-13
#613 provide listener for send/stream methods

dguggemos:
Added methods with additional listener that gets notified if respective send/stream methods succeeds/fails.


slandelle:
@dguggemos Nice! Sadly, I can't merge it yet as this is API breaking (`WebSocket`). Will do when I'll switch master to 2.1, hopefully by the end of the year.

slandelle:
@dguggemos Thanks!

PR #1282, opened 2016-10-19
Add built-in support for InputStream body with defined Content-Length

Spikhalskiy:
Current state:
Hi, now if you pass ```InputStream``` as a body source - you always get request with chunked encoding.

Motivation:
Provide built-in ability to send requests with static ```Content-Length``` if we have fully in-memory ```InputStream``` (like ```ByteArrayInputStream```) + length.

Changes:
Added ```streamContentLength``` field to ```RequestBuilder```s and ```Request``` interface. This field could be used to support requests with ```Content-Length``` for any dynamic lengthed bodies in the future. Added optional field ```contentLength``` to ```NettyInputStreamBody```.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
@@ -103,7 +103,8 @@ public final class NettyRequestFactory {
                 nettyBody = new NettyFileBody(fileBodyGenerator.getFile(), fileBodyGenerator.getRegionSeek(), fileBodyGenerator.getRegionLength(), config);
 
             } else if (request.getBodyGenerator() instanceof InputStreamBodyGenerator) {
-                nettyBody = new NettyInputStreamBody(InputStreamBodyGenerator.class.cast(request.getBodyGenerator()).getInputStream());
+                InputStreamBodyGenerator inStreamGenerator = InputStreamBodyGenerator.class.cast(request.getBodyGenerator());
+                nettyBody = new NettyInputStreamBody(inStreamGenerator.getInputStream(), inStreamGenerator.getContentLength());
 
             } else if (request.getBodyGenerator() instanceof ReactiveStreamsBodyGenerator) {
                 ReactiveStreamsBodyGenerator reactiveStreamsBodyGenerator = (ReactiveStreamsBodyGenerator)request.getBodyGenerator();
--- a/client/src/main/java/org/asynchttpclient/netty/request/body/NettyInputStreamBody.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/body/NettyInputStreamBody.java
@@ -33,9 +33,15 @@ public class NettyInputStreamBody implements NettyBody {
     private static final Logger LOGGER = LoggerFactory.getLogger(NettyInputStreamBody.class);
 
     private final InputStream inputStream;
+    private final long contentLength;
 
     public NettyInputStreamBody(InputStream inputStream) {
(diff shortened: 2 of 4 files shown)


slandelle:
Hi

Sorry, but that's not something can can be merged into 2.0 (`Request` API breaking).

Then, I'm not super convinced by the use case:
* an all-in-memory `InputStream` with known size is not a "real" stream, eg a `ByteArrayInputStream` can easily provide a byte array.
* `InputStream` is a poor player in `AsyncHttpClient` in the sense that it's a blocking API. If you're using a real network based `InputStream`, you might end up blocking an IO thread, hence stalling all the channels locked on this thread.
* You can use a `BodyGenerator`
* If we're to break APIs, I suspect it would be better to completely reconsider in a future version how bodies are being passed, eg introduce an `Entity` abstraction

WDYT?

Spikhalskiy:
Hi,

> an all-in-memory InputStream with known size is not a "real" stream, eg a ByteArrayInputStream can easily provide a byte array.

Spikhalskiy:
@slandelle I commited this partial revert as a second commit in this branch. Should be without breaking changes. Let me know if it's fine - I will squash it.

slandelle:
> Right, they could, but they will allocate memory in most cases for it. And it could be avoided by current approach.

No, that's the same number of copies: most contents have to be copied into NIO buffers, while byte arrays can be directly wrapped.
I don't think your change brings any performance improvement.

What it indeed does is your original motivation: send a `Content-Length` header instead of a `Transfer-Encoding: chunked` one. Why not, but why is that important to you?

> I commited this partial revert as a second commit in this branch. Should be without breaking changes.

Could you please revert to the original imports ordering? Your PR is off because of that: 
capture d ecran 2016-10-20 a 00 02 14

Spikhalskiy:
> No, that's the same number of copies: most contents have to be copied into NIO buffers, while byte arrays can be directly wrapped.

I hope that this buffers would be reused from Netty pools. It's important for my task. Am I not right? So, instead of creating byte[], wrapping it - would be just a copy.

> What it indeed does is your original motivation: send a Content-Length header instead of a Transfer-Encoding: chunked one. Why not, but why is that important to you?

Motivation is pretty simple, server doesn't support chunked encoding. I solved it by custom BodyGenerator, but think it would be right to have ability to avoid chunked encoding in client framework.

slandelle:
> I hope that this buffers would be reused from Netty pools. It's important for my task. Am I not right? So, instead of creating byte[], wrapping it - would be just a copy.

Only if you enable the pooled allocator (which is the default in Netty 4.1 but not in 4.0). But then, if you're using a `ByteArrayInputStream`, you're already allocating a byte array here. Except if you're recycling it, but then why not directly use a ByteBuf? In the end, it all boils down to "why use an in-memory InputStream?!".

> Motivation is pretty simple, server doesn't support chunked encoding.

Sad server :(

slandelle:
Merged, thanks!

Spikhalskiy:
> Only if you enable the pooled allocator (which is the default in Netty 4.1 but not in 4.0). But then, if you're using a ByteArrayInputStream, you're already allocating a byte array here. Except if you're recycling it, but then why not directly use a ByteBuf? In the end, it all boils down to "why use an in-memory InputStream?!".

Heh. You are absolutely right, but in my case, for example, I get this InputStream from another library, so I have InputStream, I don't know exactly it's type and I have length of this stream :)

> Sad server :(

True! But very common in my industry, a lot of hand-writing stuff to make it more efficient. And not all standard features supported as a result common situation.

Thanks for the quick turnaround!

PR #1283, opened 2016-10-20
Expose ioThreadsCount config property

Spikhalskiy:
Hi,

It would be good to have an ability to set a static number of ioThreads in configuration.
Now this setup could be done thru ```setEventLoopGroop()``` by providing custom eventLoopGroup instance, but all code and logic of configuration/creation of native epool or nio event group becomes external. Exposing method to configure threads count makes configuration more flexible and helps to avoid writing eventLoopGroup instantiation code explicitly.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
+++ b/client/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
@@ -288,6 +288,8 @@ public interface AsyncHttpClientConfig {
 
     ByteBufAllocator getAllocator();
 
+    int getIoThreadsCount();
+
     interface AdditionalChannelInitializer {
 
         void initChannel(Channel channel) throws Exception;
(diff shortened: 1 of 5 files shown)


slandelle:
Hi,

> It would be good to have an ability to set a static number of ioThreads in configuration.

There's a demand for this about every 2 years.
Every time, the user was just trying to test the impact of trying to tune this parameter, and in the end, his tests failed to find use cases where it would interesting to change the default value (but could do the opposite: uses case where changing the default harmed performance).

Do you have such use case that would prove this change indeed brings more value?

Spikhalskiy:
@slandelle We are trying to keep the total amount of threads in our system <= number of cores. This makes it very easy to understand which subsystem is overloaded, which should be tuned, etc.
If you put 2 threads for a client and see that this threads are running 100% of time - mostly possible that here is a bottleneck and situation changed here, because previously this setup worked fine. If you have 24 threads for 12 cores and they are runnable for 60% of the time - it means nothing, because there is no resources separation.
This helps to isolate the processes of different subsystems from affecting each other.
So, we want to put for a http client not more that 3 cores to don't affect other threads for example. If it would require more cores as a result of some changes (more endpoints to call, whatever) - we would think about how to spread the CPU load, add one more core to this system or it's just a general capacity problem.

So, setting number of processes is not making some performance improvements, but helps to easily pin client to not more than XX% of CPU time and manage this setup/investigate which subsystem caused load problem in case of traffic volume/structure changes.

PR #1286, opened 2016-10-26
Future

stepancheg:
Two patches improving futures.

* simplify `AbstractListenableFuture.executionList` initialization
* Allow `null` executor in `ListenableFuture`

### DIFF
--- a/client/src/main/java/org/asynchttpclient/future/AbstractListenableFuture.java
+++ b/client/src/main/java/org/asynchttpclient/future/AbstractListenableFuture.java
@@ -29,6 +29,7 @@
 package org.asynchttpclient.future;
 
 import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
 import org.asynchttpclient.ListenableFuture;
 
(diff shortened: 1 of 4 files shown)


client/src/main/java/org/asynchttpclient/future/AbstractListenableFuture.java
```diff
@@ -43,41 +44,44 @@
...
-        ExecutionList localExecutionList = executionList;
-        if (localExecutionList == null) {
-            synchronized (this) {
-                localExecutionList = executionList;
-                if (localExecutionList == null) {
-                    localExecutionList = new ExecutionList();
-                    executionList = localExecutionList;
-                    executionListInitialized = true;
-                }
+        for (;;) {
```
slandelle: Are you sure we need this?

slandelle:
Nice catch, thanks! I do have a question though.

client/src/main/java/org/asynchttpclient/future/AbstractListenableFuture.java
stepancheg: Not sure what you meant. Loop is not needed indeed. I've updated the PR.

slandelle: That was exactly my thought. :)

stepancheg:
Wait!

I made a mistake in `addListener`:

```
    @Override
    public ListenableFuture<V> addListener(Runnable listener, Executor exec) {
        if (executionList == null) {
            ExecutionList.executeListener(listener, exec); // <-- HERE
            return this;
        }

        executionList().add(listener, exec);
        return this;
    }
```

Revert it please.

stepancheg:
Going to submit alternative patch.

stepancheg:
Revert for the first part is #1288, better patch for the first part is #1287.

slandelle:
@stepancheg I deleted the commit

PR #1287, opened 2016-10-26
Replace ExecutionList with linked list of RunnableExecutionPair

stepancheg:
Patch does CAS loop and XCHG instead of synchronized which should
be faster.

Patch also reduces per-future memory consumption.


stepancheg:
Should revert previous version of the patch (can be done by applying #1288).

slandelle:
> Patch does CAS loop and XCHG instead of synchronized which should be faster.

Do you have some numbers please? This code is forked from Guava, so before introducing code that will have to be maintained here, I'd rather be sure it's worth it.

stepancheg:
Well, it is easy to create a benchmark that proves anything.

I've created a simple benchmark:

https://gist.github.com/stepancheg/2f3236cb5428f215eb2d92e0fafed4f2

It prints:

```
19158
18589
19825
18905
19007
17067
18801
16763
```

on current master, and prints 

```
8200
6686
6798
6620
7832
7202
7713
6623
6706
```

on patched version. Slower is better. It is important for benchmark that `addListener` and `runListeners` are executed on different threads, otherwise JVM uses lock elision and original version is slightly faster.

However, the most important reason for me is better memory utilization in patched version: there is no `ExecutionList` object per future.

slandelle:
Awesome work, thanks!

PR #1289, opened 2016-10-26
Allow null executor in ListenableFuture

stepancheg:
It allows execution of future callback in completion thread, and
makes `ListenableFuture` closer to `CompletableFuture` (which allows
`null` executor).

It also saves a little memory in `toCompletableFuture` implementation.


slandelle:
Thanks!

PR #1291, opened 2016-10-26
Replace AtomicBoolean with AtomicFieldUpdater

stepancheg:
Avoid unnecessary heap allocations.

Currently in my app I have 100K `AtomicBoolean` objects with 50K more created per second in async-http-client futures. Probably it is not a big deal, but it is better not to allocate anyway.


slandelle:
I really can't remember why I did replace AtomicInteger fields with FieldUpdaters and not the AtomicBooleans too?!
Thanks a bunch!

PS: I'll rename some getters to isSomething.

PR #1294, opened 2016-10-28
Rewrite NettyResponseFuture using CompletableFuture

stepancheg:
`CompletableFuture` seems to be as expensive as `CountDownLatch`, however, when user asks for `toCompletableFuture`, `NettyResponseFuture` now returns already allocated `CompletableFuture` object without `Runnable` registration in `AbstractListenableFuture`.

Also, I believe that `CompletableFuture` is well optimized.

Tests pass, although it is hard to tell if anything is broken, because `NettyResponseFuture` implementation is hard.


slandelle:
Looking good! Please let me review carefully later today.

client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
```diff
@@ -47,24 +45,18 @@
  * 
  * @param <V> the result type
  */
-public final class NettyResponseFuture<V> extends AbstractListenableFuture<V> {
+public final class NettyResponseFuture<V> implements ListenableFuture<V> {
```
slandelle: So `AbstractListenableFuture` and `RunnableExecutorPair` can be removed, as they are replaced with `CompletableFuture#whenCompleteAsync`, right?

stepancheg: Yes. Should I remove it?

slandelle: Yes, please

stepancheg:
Update PR with removed `AbstractListenableFuture` and `RunnableExecutorPair`.

slandelle:
Thanks a lot!

PR #1295, opened 2016-10-31
Add some useful API methods

Spikhalskiy:
Motivation and proposals:

1) Method ```setHeaders(Map<String, Collection<String>> headers)``` is general, but in most cases applications use headers with only one value, so it's good to have a simple method to set them without converting ```Map<String, String>``` to ```Map<String, Collections.singletonList<String>>```.
2) Now if you already have Cookies in some entities from another API's it's a bit annoying to convert them to ```org.asynchttpclient.Cookie``` manually. Expose two helper factory methods set for Netty's and standard ```javax.servlet``` entities.

These methods were implemented as static utils for async-http-client in my project, I think it would be handy to have them out of the box.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
+++ b/client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
@@ -15,13 +15,22 @@
  */
 package org.asynchttpclient;
 
-import static org.asynchttpclient.util.HttpUtils.*;
-import static org.asynchttpclient.util.MiscUtils.isNonEmpty;
 import io.netty.handler.codec.http.DefaultHttpHeaders;
 import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.resolver.DefaultNameResolver;
 import io.netty.resolver.NameResolver;
 import io.netty.util.concurrent.ImmediateEventExecutor;
+import org.asynchttpclient.channel.ChannelPoolPartitioning;
+import org.asynchttpclient.cookie.Cookie;
+import org.asynchttpclient.proxy.ProxyServer;
+import org.asynchttpclient.request.body.generator.BodyGenerator;
+import org.asynchttpclient.request.body.generator.ReactiveStreamsBodyGenerator;
+import org.asynchttpclient.request.body.multipart.Part;
+import org.asynchttpclient.uri.Uri;
+import org.asynchttpclient.util.UriEncoder;
+import org.reactivestreams.Publisher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.InputStream;
@@ -34,17 +43,9 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
-import org.asynchttpclient.channel.ChannelPoolPartitioning;
-import org.asynchttpclient.cookie.Cookie;
-import org.asynchttpclient.proxy.ProxyServer;
-import org.asynchttpclient.request.body.generator.BodyGenerator;
-import org.asynchttpclient.request.body.generator.ReactiveStreamsBodyGenerator;
-import org.asynchttpclient.request.body.multipart.Part;
-import org.asynchttpclient.uri.Uri;
-import org.asynchttpclient.util.UriEncoder;
-import org.reactivestreams.Publisher;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static org.asynchttpclient.util.HttpUtils.parseCharset;
+import static org.asynchttpclient.util.HttpUtils.validateSupportedScheme;
+import static org.asynchttpclient.util.MiscUtils.isNonEmpty;
 
 /**
  * Builder for {@link Request}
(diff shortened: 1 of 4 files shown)


slandelle:
Hi,

There's too many different things in this PR. It would be better to isolate them.

I'm OK with the first one (once again, I'm so sad/disappointed by Java's type erasure...). See comments.

Regarding the second one, I'm not fond of it. First, it would introduce a required dependency to the Servlet API for very limited value. And then, it would mean having different ways of doing the same thing. And what about `java.net.HttpCookie` (that's shipped with JDK) or `io.netty.handler.codec.http.cookie.Cookie` that ships with Netty?
I would be fine with bridges/factories that only require an optional dependency, though.

client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
```diff
@@ -34,17 +43,9 @@
...
-import org.asynchttpclient.request.body.generator.ReactiveStreamsBodyGenerator;
-import org.asynchttpclient.request.body.multipart.Part;
-import org.asynchttpclient.uri.Uri;
-import org.asynchttpclient.util.UriEncoder;
-import org.reactivestreams.Publisher;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static org.asynchttpclient.util.HttpUtils.parseCharset;
+import static org.asynchttpclient.util.HttpUtils.validateSupportedScheme;
+import static org.asynchttpclient.util.MiscUtils.isNonEmpty;
```
slandelle: Please revert imports ordering: unrelated.

client/src/main/java/org/asynchttpclient/cookie/Cookie.java
```diff
@@ -12,16 +12,65 @@
  */
 package org.asynchttpclient.cookie;
 
-import static org.asynchttpclient.cookie.CookieUtil.*;
+import static org.asynchttpclient.cookie.CookieUtil.validateCookieAttribute;
+import static org.asynchttpclient.cookie.CookieUtil.validateCookieName;
+import static org.asynchttpclient.cookie.CookieUtil.validateCookieValue;
```
slandelle: please revert

client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
```diff
@@ -194,14 +195,17 @@ public T setHeaders(HttpHeaders headers) {
...
-                this.headers.add(headerName, entry.getValue());
-            }
+            headers.forEach((name, values) -> this.headers.add(name, values));
         }
         return asDerivedType();
     }
 
+    public T setSingleHeaders(Map<String, String> headers) {
+        this.headers.clear();
+        headers.forEach((name, value) -> this.headers.add(name, value));
```
slandelle: missing null check, like in setHeaders

client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
```diff
@@ -194,14 +195,17 @@ public T setHeaders(HttpHeaders headers) {
...
-                String headerName = entry.getKey();
-                this.headers.add(headerName, entry.getValue());
-            }
+            headers.forEach((name, values) -> this.headers.add(name, values));
         }
         return asDerivedType();
     }
 
+    public T setSingleHeaders(Map<String, String> headers) {
+        this.headers.clear();
```
slandelle: add a `clearHeaders` method?

client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
```diff
@@ -194,14 +195,17 @@ public T setHeaders(HttpHeaders headers) {
...
-            for (Map.Entry<String, Collection<String>> entry : headers.entrySet()) {
-                String headerName = entry.getKey();
-                this.headers.add(headerName, entry.getValue());
-            }
+            headers.forEach((name, values) -> this.headers.add(name, values));
         }
         return asDerivedType();
     }
 
+    public T setSingleHeaders(Map<String, String> headers) {
```
slandelle: missing Javadoc

pom.xml
```diff
@@ -251,6 +251,12 @@
 				<version>${netty.version}</version>
 				<optional>true</optional>
 			</dependency>
+			<dependency>
+				<groupId>javax.servlet</groupId>
+				<artifactId>javax.servlet-api</artifactId>
+				<version>${javax.servlet-api.version}</version>
+				<optional>true</optional>
```
slandelle: Are you sure this dep is really optional? I feel like the `Cookie` class would fail to load and crash.

Spikhalskiy: No, just import doesn't mean anything. Cookie class would be loaded (and fail) if somebody actually would use these new methods.

Spikhalskiy:
@slandelle +1 for dropping cookie in favour of Netty's Cookie.
> I would be OK with isolated bridges that only require an optional dependency, though.
Current code should be absolutely fine with optional dependency (so, if nobody will add javax.servlet and would not use new conversion methods - there would be no ClassDefNotFound), but I would double check.

slandelle:
Closing this.
Let's aim at dropping AHC's own `Cookie` impl for AHC 2.1. I just want to make sure I contributed everything upstream and we don't lose anything in the move.

PR #1296, opened 2016-11-01
Expose additional set headers method for headers with single and multiple values

Spikhalskiy:
This has been separated from PR https://github.com/AsyncHttpClient/async-http-client/pull/1295

Motivation:

Expose additional methods to make it more handy and flexible to setup single/multi-value headers using ```RequestBuilder```.


slandelle:
Thanks!

PR #1317, opened 2016-12-15
Remove usage of Netty Atomic*FieldUpdater in favor of JDKs

johnou:
Motivation:

In later Java8 versions our Atomic*FieldUpdater are slower then the JDK implementations so we should not use ours anymore. Even worse the JDK implementations provide for example an optimized version of addAndGet(...) using intrinsics which makes it a lot faster for this use-case.

Modifications:

- Remove methods that return Netty Atomic*FieldUpdaters.
- Use the JDK implementations everywhere.

Result:

Faster code.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
+++ b/client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
@@ -14,7 +14,6 @@
 package org.asynchttpclient.netty;
 
 import static org.asynchttpclient.util.DateUtils.unpreciseMillisTime;
-import static io.netty.util.internal.PlatformDependent.*;
 import io.netty.channel.Channel;
 
 import java.util.concurrent.CancellationException;
@@ -49,8 +48,10 @@ public final class NettyResponseFuture<V> implements ListenableFuture<V> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NettyResponseFuture.class);
 
-    private static final AtomicIntegerFieldUpdater<NettyResponseFuture<?>> REDIRECT_COUNT_UPDATER = newAtomicIntegerFieldUpdater(NettyResponseFuture.class, "redirectCount");
-    private static final AtomicIntegerFieldUpdater<NettyResponseFuture<?>> CURRENT_RETRY_UPDATER = newAtomicIntegerFieldUpdater(NettyResponseFuture.class, "currentRetry");
+    @SuppressWarnings("rawtypes")
+    private static final AtomicIntegerFieldUpdater<NettyResponseFuture> REDIRECT_COUNT_UPDATER = AtomicIntegerFieldUpdater.newUpdater(NettyResponseFuture.class, "redirectCount");
+    @SuppressWarnings("rawtypes")
+    private static final AtomicIntegerFieldUpdater<NettyResponseFuture> CURRENT_RETRY_UPDATER = AtomicIntegerFieldUpdater.newUpdater(NettyResponseFuture.class, "currentRetry");
 
     private final long start = unpreciseMillisTime();
     private final ChannelPoolPartitioning connectionPoolPartitioning;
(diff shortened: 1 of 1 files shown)


johnou:
@slandelle heads up that without this async-http-client will be broken with the next release of Netty, see https://github.com/netty/netty/pull/6130

johnou:
Where is the 2.0 branch? 😞

slandelle:
> heads up that without this async-http-client will be broken with the next release of Netty, see netty/netty#6130

Yeah, I was aware of that, but I was waiting for 4.1.7 to be released :)

> Where is the 2.0 branch? 

Not created yet, as I haven't backported anything yet since upgrading master to AHC2.1/Netty4.1. Will do.

client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
```diff
@@ -49,8 +48,8 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NettyResponseFuture.class);
 
-    private static final AtomicIntegerFieldUpdater<NettyResponseFuture<?>> REDIRECT_COUNT_UPDATER = newAtomicIntegerFieldUpdater(NettyResponseFuture.class, "redirectCount");
-    private static final AtomicIntegerFieldUpdater<NettyResponseFuture<?>> CURRENT_RETRY_UPDATER = newAtomicIntegerFieldUpdater(NettyResponseFuture.class, "currentRetry");
+    private static final AtomicIntegerFieldUpdater<NettyResponseFuture> REDIRECT_COUNT_UPDATER = AtomicIntegerFieldUpdater.newUpdater(NettyResponseFuture.class, "redirectCount");
+    private static final AtomicIntegerFieldUpdater<NettyResponseFuture> CURRENT_RETRY_UPDATER = AtomicIntegerFieldUpdater.newUpdater(NettyResponseFuture.class, "currentRetry");
```
slandelle: Why remove the `<?>`? Without them, you get tons of compiler warnings.

johnou:
> Yeah, I was aware of that, but I was waiting for 4.1.7 to be released :)

Ow, unfortunately I am on a tight release schedule and I just cut an internal build of Netty from 4.0 branch and our application blew up unexpectedly..

```noformat
java.lang.NoClassDefFoundError: Could not initialize class org.asynchttpclient.netty.NettyResponseFuture
	at org.asynchttpclient.netty.request.NettyRequestSender.newNettyResponseFuture(NettyRequestSender.java:311) ~[async-http-client-2.0.24.jar:?]
	at org.asynchttpclient.netty.request.NettyRequestSender.newNettyRequestAndResponseFuture(NettyRequestSender.java:193) ~[async-http-client-2.0.24.jar:?]
```

slandelle:
2.0 branch published

johnou:
Thanks, I will cut an internal build of async-http-client too.

client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
johnou: Error:(51, 137) java: incompatible types: inferred type does not conform to equality constraint(s)
    inferred: org.asynchttpclient.netty.NettyResponseFuture<?>
    equality constraints(s): org.asynchttpclient.netty.NettyResponseFuture<?>,org.asynchttpclient.netty.NettyResponseFuture

johnou: Netty takes Class<? super T> while JDK takes Class<U>

johnou: and for what it's worth I didn't see any compiler warnings after removing the wildcard, maybe you could confirm with your IDE too.

slandelle: Please add `@SuppressWarnings("rawtypes")` annotations :)

johnou: I can add it but are you absolutely sure it's necessary? my IDE and compiler are happy with it.

johnou: Added.

slandelle:
Thanks!

johnou:
Cheers!

slandelle:
cherry-picked on 2.0 as 6f53cda71219dca8ed975d53784fd9aad1b0a837

PR #1318, opened 2016-12-22
Make read timeout overridable on a per-request basis

twz123:
Simple straight forward implementation, just like it's done for request timeouts.


slandelle:
Thanks!

twz123:
Thanks for merging! Any chance that this also lands in 2.0?

slandelle:
Sorry, but this change breaks the `Request` interface, so I'm afraid it would break downstream projects, such as Play!.

PR #1341, opened 2017-01-28
Change "LATEST_VERSION" to "LATEST"

gonzalezjo:
The Maven dependency with "LATEST_VERSION" doesn't appear to work in Maven.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ First, in order to add it to your Maven project, simply download from Maven cent
 <dependency>
 	<groupId>org.asynchttpclient</groupId>
 	<artifactId>async-http-client</artifactId>
-	<version>LATEST_VERSION</version>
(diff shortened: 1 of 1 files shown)


slandelle:
The meaning of `LATEST_VERSION` here is just a placeholder, to be replaced by the latest version mentioned above in the README.

The `LATEST` special version marker that you're suggesting has actually actually been [dropped in maven 3](https://cwiki.apache.org/confluence/display/MAVEN/Maven+3.x+Compatibility+Notes#Maven3.xCompatibilityNotes-PluginMetaversionResolution).

But thanks anyway :)

gonzalezjo:
No, thank you! Up until now, I had no idea that maven 3 was even a thing.

slandelle:
:)

PR #1342, opened 2017-01-29
fixed an integer overflow bug

gculik:
when writing large files > 2147483648 bytes, the overflow bug due to the use of an integer instead of a long, causes an exception to be thrown by the caller

### DIFF
--- a/src/main/java/com/ning/http/client/multipart/FilePart.java
+++ b/src/main/java/com/ning/http/client/multipart/FilePart.java
@@ -101,7 +101,7 @@ public class FilePart extends AbstractFilePart {
 
         handler.start();
 
-        int length = 0;
+        long length = 0;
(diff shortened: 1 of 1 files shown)


slandelle:
Sorry but AHC1 has reached EOL. Please upgrade to AHC2.

slandelle:
Note that this issue has already been fixed in AHC2 for a long time.

gculik:
Is AHC2 a drop-in replacement for AHC1?  Because I am not using AHC1 directly but rather it is used by another library, and I do not currently have bandwidth to refactor that library if AHC2 is not backwardly compatible.

slandelle:
No, it's not a drop-in replacement. APIs have changed a bit and more importantly, so have groupId and root package.

PR #1345, opened 2017-02-04
Fix NPE at getClass().getClassLoader()

isopov:
I've tried to play with async-http-client in jcstress, but AFAIU it uses bootstrap classloader and creating async-http-client fails with NPE.


slandelle:
Hi,

I agree with the issue, but your fix is broken. We can't live without a loaded config.

A proper fix would be to try ContextClassLoader, Class' ClassLoader and SystemClassLoader until we find a non null one, and then use it to call `getResourceAsStream`.

Could you please fix properly?

Cheers

slandelle:
@isopov Could you please check if the fix in 6573e24cdda24ca0a13c7be316d37a7f92368b01 works for you?

isopov:
Sorry, I was on vacation - that fix works well for me, thanks!

slandelle:
Great, thanks for your feedback!

PR #1354, opened 2017-02-12
Update progress during upload

### LINKED ISSUE #728: How can I get progress rate when uploading big file? Get it when uploading not completed!
I am trying to implement async file upload with progress ,get progress rate when I uploading a big file in order to show it to user..

I tried the method suggested in the docs. Using transfer listener. http://sonatype.github.com/async-http-client/transfer-listener.html

I implemented onBytesSent of TransferListener interface (just as test):

``` java
public void onBytesSent(ByteBuffer byteBuffer) {

     System.out.println("Total bytes sent - ");
     System.out.println(byteBuffer.capacity());
}
```

Then in another thread(because I don't want to block the app) I tried to do the following:

``` java
TransferCompletionHandler tl = new TransferCompletionHandler();
tl.addTransferListener(listener);

asyncHttpClient.preparePut(getFullUrl(fileWithPath))
      .setBody(new BodyGenerator() {
           public Body createBody() throws IOException {
                return new FileBodyWithOffset(file, offset);
           }
       })
       .addHeader(CONTENT_RANGE, new ContentRange(offset, localSize).toString())
       .execute(handler).get();
```

Everything is fine. File is uploaded correctly and very fast. <b>But the issue is - I am getting messages from onBytesSent in TransferListener only AFTER the upload is finished.<b> For exmaple the upload is completed in 10 minutes. And during that 10 minutes I get nothing. And only after that everything is printed on the console.

I can't figure out what is wrong with this code. I just tried to follow the docs.

I tried to execute the above code in the main thread and it didn't work either.

Maybe it is a wrong way to implement upload progress listener using this client?

odiszapc:
To resolve #728

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/request/WriteProgressListener.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/WriteProgressListener.java
@@ -45,7 +45,9 @@ public class WriteProgressListener extends WriteListener implements ChannelProgr
             if (total < 0) {
                 total = expectedTotal;
             }
-            progressAsyncHandler.onContentWriteProgress(progress - lastLastProgress, progress, total);
+            if (progress != lastLastProgress) {
+                progressAsyncHandler.onContentWriteProgress(progress - lastLastProgress, progress, total);
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


slandelle:
Thanks! Could you please update your commit message to match this [template](http://netty.io/wiki/writing-a-commit-message.html)?

odiszapc:
Done. Please check

slandelle:
Awesome, thanks a lot!

PR #1359, opened 2017-02-19
Less logging for DefaultChannelPool

doom369:
After recent update to `2.1.0-alpha2` my logs were spammed with message:

`DEBUG- Closed 0 connections out of 0 in 0 ms`

I decreased level of logging to `trace` as debug is used often on QA envirnoments and every second prints makes logs really huge on long run.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
@@ -229,9 +229,9 @@ public final class DefaultChannelPool implements ChannelPool {
                 }
             }
 
-            if (LOGGER.isDebugEnabled()) {
+            if (LOGGER.isTraceEnabled()) {
                 long duration = unpreciseMillisTime() - start;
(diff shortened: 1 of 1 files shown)


slandelle:
> After recent update to 2.1.0-alpha2:

Well, this log has been with DEBUG level for [a very long time](https://github.com/AsyncHttpClient/async-http-client/blob/async-http-client-1.9.0/src/main/java/com/ning/http/client/providers/netty/channel/pool/DefaultChannelPool.java#L218).

> my logs were spammed with message

The default cleaner period is 1s ([1000ms](https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/resources/ahc-default.properties#L6)). I wouldn't call that spam.

> I decreased level of logging to trace as debug

Logging level is an endless debate. If I were to clean up logs, I might use TRACE for very verbose logs, that's not really the case here.
As there's a simple workaround if you don't want those logs (increase logging level for this class only), I'd rather change things only if you demonstrate the value.

doom369:
Hm. You're right. This is old code. I was certain that it started with update, however it could be changes in my logging. Sorry.

slandelle:
Thanks anyway :)

Issue #1369, opened 2017-03-21
ListenableFuture#addListener with null executor throws NPE


joedj:
ListenableFuture#addListener javadoc says:

    * Executor can be <code>null</code>, in that case executor will be executed
    * in the thread where completion happens.

However, this doesn't work with the `NettyResponseFuture` implementation:

    java.lang.NullPointerException
	at java.util.concurrent.CompletableFuture.screenExecutor(CompletableFuture.java:415)
	at java.util.concurrent.CompletableFuture.whenCompleteAsync(CompletableFuture.java:2142)
	at org.asynchttpclient.netty.NettyResponseFuture.addListener(NettyResponseFuture.java:257)

I'm happy to submit a pull request for this issue, if you can indicate how you want it fixed (i.e. update the documentation to reflect current behavour, or change behaviour to match documentation).

slandelle:
Hi @joedj 

Thanks for reporting! I guess it would be better to try to implement what the doc says. But if it turns out to be difficult, let's update the doc.
And sure, PR welcome :)

PR #1371, opened 2017-03-27
#1369 Support null executor in NettyResponseFuture#addListener

joedj:
Resolves issue #1369 

This is basically the same implementation as Guava's `MoreExecutors#directExecutor`.


slandelle:
Thanks!

PR #1372, opened 2017-03-28
Simplify semaphore logic in ChannelManager

stepancheg:
* avoid using `java.util.concurrent.Semaphore` which is blocking,
  and its blocking part is not used, which is confusing.
* Use `NonBlockingSemaphoreInfinite` to avoid nulls and ifs
  when connection count is not limited.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
@@ -45,7 +45,6 @@ import java.net.InetSocketAddress;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.Semaphore;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
@@ -109,9 +108,9 @@ public class ChannelManager {
     private final ChannelPool channelPool;
     private final ChannelGroup openChannels;
     private final boolean maxTotalConnectionsEnabled;
-    private final Semaphore freeChannels;
+    private final NonBlockingSemaphoreLike freeChannels;
     private final boolean maxConnectionsPerHostEnabled;
-    private final ConcurrentHashMap<Object, Semaphore> freeChannelsPerHost = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<Object, NonBlockingSemaphore> freeChannelsPerHost = new ConcurrentHashMap<>();
 
     private AsyncHttpClientHandler wsHandler;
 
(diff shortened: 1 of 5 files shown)


client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -336,15 +338,19 @@ public boolean removeAll(Channel connection) {
...
 
     private boolean tryAcquireGlobal() {
-        return !maxTotalConnectionsEnabled || freeChannels.tryAcquire();
+        return freeChannels.tryAcquire();
     }
 
-    private Semaphore getFreeConnectionsForHost(Object partitionKey) {
-        return freeChannelsPerHost.computeIfAbsent(partitionKey, pk -> new Semaphore(config.getMaxConnectionsPerHost()));
+    private NonBlockingSemaphoreLike getFreeConnectionsForHost(Object partitionKey) {
+        if (maxConnectionsPerHostEnabled) {
```
slandelle: ternary operator

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -139,18 +138,23 @@ public ChannelManager(final AsyncHttpClientConfig config, Timer nettyTimer) {
         maxTotalConnectionsEnabled = config.getMaxConnections() > 0;
         maxConnectionsPerHostEnabled = config.getMaxConnectionsPerHost() > 0;
 
+        if (maxTotalConnectionsEnabled) {
```
slandelle: ternary operator

client/src/main/java/org/asynchttpclient/netty/channel/NonBlockingSemaphore.java
```diff
@@ -0,0 +1,35 @@
+package org.asynchttpclient.netty.channel;
```
slandelle: Please add package header, granting copyright to "The AsyncHttpClient" project, just like in `ChannelManager` but with inception year 2017.

client/src/main/java/org/asynchttpclient/netty/channel/NonBlockingSemaphoreInfinite.java
```diff
@@ -0,0 +1,21 @@
+package org.asynchttpclient.netty.channel;
```
slandelle: Please add missing file header

client/src/main/java/org/asynchttpclient/netty/channel/NonBlockingSemaphoreLike.java
```diff
@@ -0,0 +1,12 @@
+package org.asynchttpclient.netty.channel;
```
slandelle: Please add missing file header

client/src/test/java/org/asynchttpclient/netty/channel/NonBlockingSemaphoreTest.java
```diff
@@ -0,0 +1,63 @@
+package org.asynchttpclient.netty.channel;
```
slandelle: Please add missing file header

client/src/main/java/org/asynchttpclient/netty/channel/NonBlockingSemaphoreInfinite.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Non-blocking semaphore-like object with infinite permits.
+ *
+ * So try-acquire always succeeds.
+ *
+ * @author Stepan Koltsov
+ */
+enum NonBlockingSemaphoreInfinite implements NonBlockingSemaphoreLike {
```
slandelle: `NoopNonBlockingSemaphore`?

client/src/main/java/org/asynchttpclient/netty/channel/NonBlockingSemaphoreInfinite.java
```diff
@@ -0,0 +1,21 @@
...
+    I;
```
slandelle: `INSTANCE`?

slandelle:
@stepancheg Thanks! Could you please address comments?

client/src/main/java/org/asynchttpclient/netty/channel/NonBlockingSemaphoreInfinite.java
stepancheg: No-op does not make it clear that `tryAcquire` returns true. `Infinite` make it look like a semaphore with infinite number of permits, so `tryAcquire` always returns true.

I put `Infinite` part at the end of the class name so classes were together in the file list in the IDE.

So I think names are good, but I can change both name and placement if you still think it is better.

stepancheg:
Addressed all the issues except name of semaphore class, see comments above.

slandelle:
Great job, thanks!

PR #1373, opened 2017-03-28
Store partition key in Channel attribute instead of global hash map

stepancheg:


slandelle:
@stepancheg Thanks a lot!

PR #1378, opened 2017-03-28
Unnecessary AtomicBoolean in inactive token attribute

stepancheg:


slandelle:
Oh, I didn't realize `Attribute` had atomic CAS, thanks!

PR #1379, opened 2017-03-28
Tie connection counter to the future/channel

stepancheg:
Instead of manually tracking where connection counter should be
released, just do it when future becomes done or connection becomes
closed.

I believe, code is safer against leaks this way.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
+++ b/client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
@@ -16,6 +16,7 @@ package org.asynchttpclient.netty;
 import static org.asynchttpclient.util.DateUtils.unpreciseMillisTime;
 import io.netty.channel.Channel;
 
+import java.io.IOException;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
@@ -31,8 +32,10 @@ import org.asynchttpclient.ListenableFuture;
 import org.asynchttpclient.Realm;
 import org.asynchttpclient.Request;
 import org.asynchttpclient.channel.ChannelPoolPartitioning;
+import org.asynchttpclient.netty.channel.ChannelManager;
 import org.asynchttpclient.netty.channel.ChannelState;
 import org.asynchttpclient.netty.channel.Channels;
+import org.asynchttpclient.netty.channel.ConnectionSemaphore;
 import org.asynchttpclient.netty.request.NettyRequest;
 import org.asynchttpclient.netty.timeout.TimeoutsHolder;
 import org.asynchttpclient.proxy.ProxyServer;
@@ -56,6 +59,7 @@ public final class NettyResponseFuture<V> implements ListenableFuture<V> {
 
     private final long start = unpreciseMillisTime();
     private final ChannelPoolPartitioning connectionPoolPartitioning;
+    private final ConnectionSemaphore connectionSemaphore;
     private final ProxyServer proxyServer;
     private final int maxRetry;
     private final CompletableFuture<V> future = new CompletableFuture<>();
@@ -73,6 +77,8 @@ public final class NettyResponseFuture<V> implements ListenableFuture<V> {
     private volatile int onThrowableCalled = 0;
     @SuppressWarnings("unused")
     private volatile TimeoutsHolder timeoutsHolder;
+    // partition key, when != null used to release lock in ChannelManager
+    private volatile Object partitionKeyLock;
 
     @SuppressWarnings("rawtypes")
     private static final AtomicIntegerFieldUpdater<NettyResponseFuture> isDoneField = AtomicIntegerFieldUpdater.newUpdater(NettyResponseFuture.class, "isDone");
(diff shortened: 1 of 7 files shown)


slandelle:
@stepancheg Great piece of work, thanks! Did it fix your issue? Could you please answer/address comments?

stepancheg:
> Did it fix your issue?

Don't know yet. Probably there was not issue in latest alpha. We used alpha1 before.

So, now we are running our program with alpha9.

> Could you please answer/address comments?

Sorry, what comments?

BTW, I've rebased the patch against master.

stepancheg:
No, seems like alpha9 is also leaking:



And funny thing is that on another server it seems to leak in different direction: number of permits + number of open connection exceeds configured max connections.

Now going to try patched version.

stepancheg:
Updated patch to remove now unused `partitionKey` channel attribute.

stepancheg:
Patched version seem not to leak counter, but it hard to say with 100% confidence after just an hour of running.

client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
```diff
@@ -454,6 +482,29 @@ public Object getPartitionKey() {
...
+        Object partitionKey = getPartitionKey();
+        channelManager.acquireChannelLock(partitionKey);
+        Object prevKey = partitionKeyLockField.getAndSet(this, partitionKey);
+        if (prevKey != null) {
+            // self-check
+
+            channelManager.releaseChannelLock(prevKey);
+            releasePartitionKeyLock();
+
+            throw new AssertionError("unreachable");
```
stepancheg: ```throw new IllegalStateException("Trying to acquire partition lock concurrently. Please report.");``` "Twice" is not proper word, because function is allowed to be called twice.

client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
```diff
@@ -56,6 +58,7 @@
 
     private final long start = unpreciseMillisTime();
     private final ChannelPoolPartitioning connectionPoolPartitioning;
+    private final ChannelManager channelManager;
```
stepancheg: Something like `ConnectionSemaphore`. Implementing it now.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -265,16 +265,10 @@ private Channel getOpenChannel(NettyResponseFuture<?> future, Request request, P
 
         Object partitionKey = future.getPartitionKey();
 
-        // we disable channelPreemption when performing next requests
-        final boolean acquireChannelLock = !performingNextRequest;
```
stepancheg: Yes, I overlooked it, and I agree, that it did something strange.

stepancheg:
Updated the patch with you requests.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -93,30 +92,27 @@
...
-    private final IOException tooManyConnectionsPerHost;
 
     private final ChannelPool channelPool;
     private final ChannelGroup openChannels;
-    private final boolean maxTotalConnectionsEnabled;
-    private final NonBlockingSemaphoreLike freeChannels;
-    private final boolean maxConnectionsPerHostEnabled;
-    private final ConcurrentHashMap<Object, NonBlockingSemaphore> freeChannelsPerHost = new ConcurrentHashMap<>();
+
+    private final ConnectionSemaphore connectionSemaphore;
```
slandelle: This shouldn't be here :)

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -93,30 +92,27 @@
...
 
     private AsyncHttpClientHandler wsHandler;
 
     public ChannelManager(final AsyncHttpClientConfig config, Timer nettyTimer) {
 
         this.config = config;
+
+        connectionSemaphore = new ConnectionSemaphore(config);
```
slandelle: same as above

slandelle:
@stepancheg How did your tests went? Is the semaphore leak fixed? If so, I'm a bit frustrated that we don't know how the leak happened :) Are we sure that was just the semaphores leaking and not the connections too?
There's just some dead code to be removed, and then I can merge when you want.

Thanks a lot!

stepancheg:
> How did your tests went? Is the semaphore leak fixed?

Well, counters seems to be correct for last 8 hours. It means that either problem is fixed or system/network load changed.

And at the same time I see incorrect semaphore values on different servers (with ligher load) on alpha9, on these servers probably semaphore is released extra time.



> I'm a bit frustrated that we don't know how the leak happened

I think there were 10s of places where leak could happen. Because it is hard to manually manager locks/ref counters. The idea of the patch is to write code in a way that you don't have to check and recheck every line for corner cases and exception safety.

> Are we sure that was just the semaphores leaking and not the connections too?

I think so. We've checked previosly with `lsof`, there were expected number of open connections.

>  There's just some dead code to be removed, and then I can merge when you want.

OK, give me a couple of minutes.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
slandelle: The original author's intent was to not require an extra semaphore when being redirected to another host. IMHO, too specific to be worth supporting.

stepancheg:
Updated a patch, removed dead code.

slandelle:
Great! I'll merge, try to backport on 2.0 too, and will release both 2.1-alpha and 2.0.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
stepancheg: > The original author's intent was to not require an extra semaphore when being redirected to another host.

And that is actually wrong IMHO. For example, because of that behaviour you can run off your operating system file descriptor limit which could be configured to be almost equal to connection count limit.

stepancheg:
I'd avoid backporting to 2.0, it's too dangerous IMO. I'd prefer if you released it as alpha, and people (in particular, our project) could perform more tests with this change. And even if this patch fixes issues, some behaiour could be changed, and users could rely on old behavior.

slandelle:
I've just released 2.1.0-alpha10, it should be up on central in 20 min.
I've backported your work in a branch: https://github.com/AsyncHttpClient/async-http-client/tree/2.0-fix1377
Once the fix is confirmed and stable, I still think we should release it. I don't think there's any behavior change.

Thanks a lot!

PR #1389, opened 2017-04-11
fixed a typo in the readme.

ijklr:
The java example code in the readme wouldn't run with the typo.


slandelle:
Sorry, but this code is fine. You need the Dsl.* static import, just like in the sample.

ijklr:
Ok. sorry about that!

PR #1391, opened 2017-04-18
Introduce AHC RxJava 2.x extras

twz123:
* Main interface is `RxHttpClient`, modelled after `AsyncHttpSingle` from RxJava 1.x extras
* Use `Maybe` reactive base type instead of `Single`, since RxJava 2 won't allow emission of `null` values any longer
* Update to RxJava/ReactiveStreams terminology (i.e. "unsubscribe" became "dispose")

### DIFF
--- a/extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/DisposedException.java
+++ b/extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/DisposedException.java
@@ -19,12 +19,9 @@ import java.util.concurrent.CancellationException;
  * Indicates that the HTTP request has been disposed asynchronously via RxJava.
  */
 public class DisposedException extends CancellationException {
-    private static final long serialVersionUID = -2913992117415097629L;
+    private static final long serialVersionUID = -5885577182105850384L;
 
-    public DisposedException() {
-    }
-
-    public DisposedException(final Throwable cause) {
-        initCause(cause);
+    public DisposedException(String message) {
+        super(message);
     }
 }
--- a/extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/DisposedException.java
+++ b/extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/DisposedException.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2017 AsyncHttpClient Project. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at
+ *     http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+package org.asynchttpclient.extras.rxjava2;
+
+import java.util.concurrent.CancellationException;
+
+/**
+ * Indicates that the HTTP request has been disposed asynchronously via RxJava.
+ */
+public class DisposedException extends CancellationException {
+    private static final long serialVersionUID = -2913992117415097629L;
+
+    public DisposedException() {
+    }
+
+    public DisposedException(final Throwable cause) {
+        initCause(cause);
+    }
+}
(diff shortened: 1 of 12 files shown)


extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/DisposedException.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Indicates that the HTTP request has been disposed asynchronously via RxJava.
+ */
+public class DisposedException extends CancellationException {
+    private static final long serialVersionUID = -2913992117415097629L;
+
+    public DisposedException() {
+    }
+
+    public DisposedException(final Throwable cause) {
```
slandelle: looks unused to me

extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/RxHttpClient.java
```diff
@@ -0,0 +1,85 @@
...
+     *            supplies the desired {@code AsyncHandler} instances that are used to produce results
+     *
+     * @return a {@code Maybe} that executes {@code request} upon subscription and that emits the results produced by
+     *         the supplied handers
+     *
+     * @throws NullPointerException
+     *             if at least one of the parameters is {@code null}
+     */
+    <T> Maybe<T> prepare(Request request, Supplier<? extends AsyncHandler<T>> handlerSupplier);
+
```
slandelle: nit: kill empty line

extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/maybe/AbstractMaybeAsyncHandlerBridge.java
```diff
@@ -0,0 +1,170 @@
...
+    @Override
+    public final Void onCompleted() {
+        if (delegateTerminated.getAndSet(true)) {
+            return null;
+        }
+
+        final T result;
+        try {
+            result = delegate().onCompleted();
+        } catch (final Throwable t) {
```
slandelle: Is it really OK to catch Throwable instead of only Exception?

extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/maybe/AbstractMaybeAsyncHandlerBridge.java
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * Called to indicate that request processing is to be aborted because the linked Rx stream has been disposed. If
+     * the {@link #delegate() delegate} didn't already receive a terminal event,
+     * {@code AsyncHandler#onThrowable(Throwable) onThrowable} will be called with a {@link DisposedException}.
+     *
+     * @return always {@link State#ABORT}
+     */
+    protected final AsyncHandler.State disposed() {
+        if (!delegateTerminated.getAndSet(true)) {
+            delegate().onThrowable(new DisposedException());
```
slandelle: `DisposedException` could be turned into a singleton/constant without stacktrace.

extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/DisposedException.java
twz123: Yep, just like in the RxJava1 extras. It's just a convenience thingy when someone wants to use/extend this stuff. I'll remove it if you prefer...?

extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/maybe/AbstractMaybeAsyncHandlerBridge.java
twz123: That's an RxJava pattern found in various places there, e.g. in [one of the the map operator implementations](https://github.com/ReactiveX/RxJava/blob/v2.0.8/src/main/java/io/reactivex/internal/operators/single/SingleMap.java#L57). If it's some "fatal" exception from RxJava's point of view, it will be rethrown via [`io.reactivex.exceptions.Exceptions.throwIfFatal(…)`](https://github.com/ReactiveX/RxJava/blob/v2.0.8/src/main/java/io/reactivex/exceptions/Exceptions.java#L64) in `emitOnError`. Everything else will be reported via `onError`.

twz123:
Done. I opted to do lazy init for the cached `DisposedException` in order to capture at least one meaningful `StackTraceElement`.

extras/rxjava2/src/main/java/org/asynchttpclient/extras/rxjava2/maybe/AbstractMaybeAsyncHandlerBridge.java
slandelle: Good to know, thanks

slandelle:
Great work, thanks a lot!

PR #1398, opened 2017-04-20
Handle onTrailingHeadersReceived in AHC extras for RxJava 1 and 2

twz123:
Override default implementation and forward it to delegates, add tests.


slandelle:
Woops, thanks!

PR #1399, opened 2017-04-20
Bump up default User-Agent HTTP header value

twz123:
Change from AHC/2.0 to AHC/2.1


slandelle:
Thanks! Actually, with the large amount of changes, I'm considering targeting 3.0 :)

PR #1414, opened 2017-05-19
Update README.md

parmarnikunjkumar:
It looks confusing to me if its getting started and getting sends me to a different link and started to something else. With this changes its clear what sends me where.


slandelle:
Thanks!
But now, you no longer have a valid sentence.
Maybe turn into a bullet list?

TomGranot:
This is really old and not relevant anymore. Closing.

Issue #989, opened 2015-10-06
Switch to GET on 301 redirect too


slandelle:
Wrong as per RFC2616, but that's what browsers do...

Issue #1380, opened 2017-03-29
Spurious ReactiveStreamsTest#testConnectionDoesNotGetClosed failure


slandelle:
Reproducer: `org.asynchttpclient.reactivestreams.ReactiveStreamsTest#main` (might have to run several times):

```
testConnectionDoesNotGetClosed(org.asynchttpclient.reactivestreams.ReactiveStreamsTest)  Time elapsed: 0.342 sec  <<< FAILURE!
java.lang.AssertionError:  expected [[B@95f1422] but found [[B@31b7112d]
	at org.asynchttpclient.reactivestreams.ReactiveStreamsTest.testConnectionDoesNotGetClosed(ReactiveStreamsTest.java:66)
```

farmdawgnation:
I checked the most recent build from Travis.

I found this buried in the logs.

```
Results :



Failed tests: 

  BasicHttpsTest.multipleSequentialPostRequestsOverHttps:95->lambda$multipleSequentialPostRequestsOverHttps$5:96->lambda$null$4:105 expected [hello there] but found [<html>

<head>

<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>

<title>Error 500 No handler enqueued</title>

</head>

<body><h2>HTTP ERROR 500</h2>

<p>Problem accessing /foo/bar. Reason:

<pre>    No handler enqueued</pre></p><hr><a href="http://eclipse.org/jetty">Powered by Jetty:// 9.4.6.v20170531</a><hr/>



</body>

</html>

]

  ReactiveStreamsTest.testConnectionDoesNotGetClosed:256 Invalid response byte at position 12000 expected [111] but found [45]
```

I'm not sure, but it looks like there's more than one issue going on here...

slandelle:
I'm pretty sure I got the `BasicHttpsTest.multipleSequentialPostRequestsOverHttps` failure pinned, see #1434

But sadly, it's completely unrelated to the `ReactiveStreamsTest.testConnectionDoesNotGetClosed` one :(

slandelle:
I can get `ReactiveStreamsTest#main` to fail almost every time on my machine:

```
Request DefaultHttpRequest(decodeResult: success, version: HTTP/1.1)
PUT /foo/test HTTP/1.1
X-content-length: 265495
X-content-md5: sBFA+zEHvFHWTKspPX86Gg==
transfer-encoding: chunked
host: localhost:62655
accept: */*
user-agent: AHC/2.1

[...]

PR #1421, opened 2017-06-09
Fix broken Head302Test

ssoloff:
I was originally going to report a bug that my HEAD requests were being switched to GET upon a 301 redirect.  Then I came across #989 and learned that this behavior is by design. :smile:

However, before I discovered that issue, I was playing around with `Head302Test` because it seemed to confirm my impression that the HEAD method should be preserved upon a redirect.  It was at this time that I discovered this test was falsely passing.

I first updated the test to demonstrate that it actually fails.  Then I modified it to conform to the behavior described in #989 so that it now passes.  This PR captures both of these changes.

### DIFF
--- a/client/src/test/java/org/asynchttpclient/Head302Test.java
+++ b/client/src/test/java/org/asynchttpclient/Head302Test.java
@@ -16,6 +16,7 @@
 package org.asynchttpclient;
 
 import static org.asynchttpclient.Dsl.*;
+import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.fail;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


ssoloff:
The Travis build failure appears to be related to #1380:

```
ReactiveStreamsTest.testConnectionDoesNotGetClosed:95 Invalid response byte at position 12000 expected [111] but found [45]
```

slandelle:
Thanks a lot!

> The Travis build failure appears to be related to #1380:

Yep. It takes hundreds of cycles on my machine to reproduce. I guess I would have to run on a small Docker container so I can investigate, and I don't have time for that atm, even though this one gets me nuts... The issue could be anywhere: JDK, Netty, AHC, netty-reactive-streams, Jetty, rx-java...

PR #1422, opened 2017-06-09
Fixes #1420

### LINKED ISSUE #1420: Implement a getter for the AsyncHttpClientConfig on AsyncHttpClient
The DefaultAsyncHttpClient overloaded constructor can take in a AsyncHttpClientConfig as an argument. However the interface, nor the implementation exposes a way to get the AsyncHttpClientConfig from that instance. This becomes an issue if one wants to re-create a new instance with the same config. Without the getter, this becomes difficult.

We would like to be able to do something like this

```
if (asyncHttpClient.isClosed())
{
      newAsyncHttpClient = new DefaultAsyncHttpClient(asyncHttpClient.getConfig()); // The getConfig() is the missing method that makes this not possible.
}
```

atehrani:
@slandelle Please review :)

### DIFF
--- a/client/src/main/java/org/asynchttpclient/AsyncHttpClient.java
+++ b/client/src/main/java/org/asynchttpclient/AsyncHttpClient.java
@@ -281,4 +281,10 @@ public interface AsyncHttpClient extends Closeable {
      * @param predicate the predicate
      */
     void flushChannelPoolPartitions(Predicate<Object> predicate);
+    
+    /**
+     * Return the config associated to this client.
+     * @return the config associated to this client.
+     */
+    AsyncHttpClientConfig getConfig();
 }
--- a/client/src/main/java/org/asynchttpclient/AsyncHttpClient.java
+++ b/client/src/main/java/org/asynchttpclient/AsyncHttpClient.java
@@ -281,7 +281,7 @@ public interface AsyncHttpClient extends Closeable {
      * @param predicate the predicate
      */
     void flushChannelPoolPartitions(Predicate<Object> predicate);
-    
+
     /**
      * Return the config associated to this client.
      * @return the config associated to this client.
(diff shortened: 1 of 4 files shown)


client/src/main/java/org/asynchttpclient/AsyncHttpClient.java
```diff
@@ -281,4 +281,10 @@
      * @param predicate the predicate
      */
     void flushChannelPoolPartitions(Predicate<Object> predicate);
+
```
slandelle: Please format and remove spaces

client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
```diff
@@ -276,4 +276,9 @@ protected BoundRequestBuilder requestBuilder(String method, String url) {
     protected BoundRequestBuilder requestBuilder(Request prototype) {
         return new BoundRequestBuilder(this, prototype).setSignatureCalculator(signatureCalculator);
     }
+
```
slandelle: Please format and remove spaces

extras/registry/src/test/java/org/asynchttpclient/extras/registry/TestAsyncHttpClient.java
```diff
@@ -134,4 +134,9 @@ public ClientStats getClientStats() {
     public void flushChannelPoolPartitions(Predicate<Object> predicate) {
         throw new UnsupportedOperationException();
     }
+
```
slandelle: Please format and remove spaces

extras/registry/src/test/java/org/asynchttpclient/extras/registry/BadAsyncHttpClient.java
```diff
@@ -138,4 +138,9 @@ public ClientStats getClientStats() {
     public void flushChannelPoolPartitions(Predicate<Object> predicate) {
         throw new UnsupportedOperationException();
     }
+
```
slandelle: Please format and remove spaces

atehrani:
@slandelle What is your formatting preference? Note that the above notes are pieces of code that I did not touch.

slandelle:
> Note that the above notes are pieces of code that I did not touch.

Of course you did. Those are all the places you added new methods. When you hit enter to jump line, your IDE doesn't create an empty line, but one with as many white spaces as the previous cursor position. Those are the ones I ask you to remove to I don't have to reformat after merging your PR.

atehrani:
@slandelle My bad! 😭  Darn non-printable characters. The extra spaces should no longer be there.

slandelle:
Thanks :)

PR #1425, opened 2017-06-13
Allow SSLEngineFactory to be customized at the request level

tafkacn:
Someone was looking for such functionality a long time ago: https://groups.google.com/forum/#!topic/asynchttpclient/7vxcOxeerjo, and I have worked on projects where such functionality can be useful so I have take this upon myself to add support for customizing the SSLEngineFactory on a per request basis. The approach that I have taken is intended to be non-intrusive (existing code logic remain the same if SSLEngineFactory is not specified at the request level) but suggestions on how to better organize the injection of the request-specific SSLEngineFactory are welcome.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/DefaultRequest.java
+++ b/client/src/main/java/org/asynchttpclient/DefaultRequest.java
@@ -61,6 +61,7 @@ public class DefaultRequest implements Request {
     private final Charset charset;
     private final ChannelPoolPartitioning channelPoolPartitioning;
     private final NameResolver<InetAddress> nameResolver;
+    private final SslEngineFactory sslEngineFactory;
     // lazily loaded
     private List<Param> queryParams;
 
@@ -88,7 +89,8 @@ public class DefaultRequest implements Request {
             long rangeOffset,//
             Charset charset,//
             ChannelPoolPartitioning channelPoolPartitioning,//
-            NameResolver<InetAddress> nameResolver) {
+            NameResolver<InetAddress> nameResolver,
+            SslEngineFactory sslEngineFactory) {
         this.method = method;
         this.uri = uri;
         this.address = address;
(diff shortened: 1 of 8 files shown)


slandelle:
> I have worked on projects where such functionality can be useful

Could you please elaborate? Before introducing new features, I want to be convinced about the use case and that the proposed implementation is the right way to support it.

Note that what you've achieved doesn't work when connection pooling is enabled.

tafkacn:
Hi @slandelle, thanks for your feedback on the pull request. The scenario that I encountered before has to do with getting security tokens from an external service (with a limited number of hosts) on behalf of our clients with TLS client certificates tied to specific clients. This is similar to the scenario that Mikko brought up in the linked message.

In hindsight, I was going about adding request-level SslEngineFactory customization in a rather naive manner, and you are absolutely right that the customization will not work when an HTTPS connection is reused with keep-alive as TLS will not be renegotiated, otherwise it would nullify most of the benefit of reusing https connections. 

The original reason why we didn't want to create a different AsyncHttpClient for each TLS client certificate was the following HashWheelTimer warning we got when we create too many AsyncHttpClient instances:

org.jboss.netty.util.internal.SharedResourceMisuseDetector
WARNING: You are creating too many HashedWheelTimer instances.  HashedWheelTimer is a shared resource that must be reused across the application, so that only a few instances are created.

But looking at the DefaultAsyncHttpClient code a bit more carefully, it would appear that we can now inject a shared instance of HashWheelTimer across multiple DefaultAsyncHttpClient instances through the config so it is entirely possible to maintain a cache/pool of relatively low-overhead DefaultAsyncHttpClient instances keyed by the TLS client certificates sharing the same HashWheelTimer. As such, I am going to go ahead and close this pull request.

slandelle:
> it would appear that we can now inject a shared instance of HashWheelTimer

Don't forget the `EventLoop` too!

Indeed, I was suspecting what you actually want was the ability to spawn multiple client instances but share the underlying resources, hence my question about actual the use case.

PR #1433, opened 2017-07-03
Trying to figure out what's up with the tests

farmdawgnation:
Some work on #1380 to try and track down the cause. I've so far been unable to reproduce the failure on my local machine. Going to see what Travis does this time.


slandelle:
Thanks

Issue #1412, opened 2017-05-17
Unable to close client properly


MiErnst:
Calling close on a client instance with default config does not wait for all resources to be closed. Wait means that `ChannelManager.close()` starts an async process (`eventLoopGroup.shutdownGracefully...`) but does not wait for the future to be completed. 

Because it does not wait, the `close` call completes before the async task is completed and my application continues to shutdown while the async task is still running. This results in an `IllegalStateException` and lots of `NoClassDefFoundError`s in my application. The `IllegalStateException` is caused because the web application which integrates the ahc is already terminated. The `NoClassDefFoundError`s are caused because the netty threads are still running but the classloader is not able to load the classes anymore. Because there are lots of threads, there are also lots of this `NoClassDefFoundError`s.

I miss either a wait till all futures are completed or a close method which provides a future as return value so I can wait for the close. 

Current workaround is to call `Thread.sleep` after calling close but this is not a deterministic solution.

```
09-May-2017 11:23:39.493 INFO [globalEventExecutor-1-2] org.apache.catalina.loader.WebappClassLoaderBase.checkStateForResourceLoading Illegal access: this web application instance has been stopped already. Could not load [io.netty.channel.group.ChannelMatchers]. The following stack trace is thrown for debugging purposes as well as to attempt to terminate the thread which caused the illegal access.
 java.lang.IllegalStateException: Illegal access: this web application instance has been stopped already. Could not load [io.netty.channel.group.ChannelMatchers]. The following stack trace is thrown for debugging purposes as well as to attempt to terminate the thread which caused the illegal access.
	at org.apache.catalina.loader.WebappClassLoaderBase.checkStateForResourceLoading(WebappClassLoaderBase.java:1305)
	at org.apache.catalina.loader.WebappClassLoaderBase.checkStateForClassLoading(WebappClassLoaderBase.java:1293)
	at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1158)
	at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1119)
	at io.netty.channel.group.DefaultChannelGroup.close(DefaultChannelGroup.java:208)
	at org.asynchttpclient.netty.channel.ChannelManager.doClose(ChannelManager.java:357)
	at org.asynchttpclient.netty.channel.ChannelManager.lambda$close$1(ChannelManager.java:364)
[...]

PR #1435, opened 2017-07-05
#1412 Wait for async close tasks before close completes

MiErnst:
#1412 Implemented a count down latch and called some API methods to close resources of netty.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/AsyncHttpClientState.java
+++ b/client/src/main/java/org/asynchttpclient/AsyncHttpClientState.java
@@ -23,7 +23,7 @@ public class AsyncHttpClientState {
         this.closed = closed;
     }
     
-    public boolean isClosed() {
+    public boolean isClosedOrClosingIsTriggered() {
         return closed.get();
     }
 }
--- a/client/src/main/java/org/asynchttpclient/AsyncHttpClientState.java
+++ b/client/src/main/java/org/asynchttpclient/AsyncHttpClientState.java
@@ -17,13 +17,13 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 public class AsyncHttpClientState {
 
-    private final AtomicBoolean closed;
+    private final AtomicBoolean closeTriggered;
     
-    public AsyncHttpClientState(AtomicBoolean closed) {
-        this.closed = closed;
+    public AsyncHttpClientState(AtomicBoolean closeTriggered) {
+        this.closeTriggered = closeTriggered;
     }
     
-    public boolean isClosedOrClosingIsTriggered() {
-        return closed.get();
+    public boolean isCloseTriggered() {
+        return closeTriggered.get();
     }
 }
--- a/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
+++ b/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
@@ -118,7 +118,7 @@ public class DefaultAsyncHttpClient implements AsyncHttpClient {
             
             //see https://github.com/netty/netty/issues/2084#issuecomment-44822314
             try {
-                ThreadDeathWatcher.awaitInactivity(5, TimeUnit.SECONDS);
+                ThreadDeathWatcher.awaitInactivity(config.getShutdownTimeout(), TimeUnit.MILLISECONDS);
             } catch(InterruptedException t) {
                 // Ignore
             }
--- a/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
+++ b/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
@@ -21,8 +21,12 @@ import io.netty.channel.EventLoopGroup;
 import io.netty.util.HashedWheelTimer;
 import io.netty.util.ThreadDeathWatcher;
 import io.netty.util.Timer;
+import io.netty.util.concurrent.GlobalEventExecutor;
 
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Predicate;
 
@@ -102,29 +106,58 @@ public class DefaultAsyncHttpClient implements AsyncHttpClient {
 
     @Override
     public void close() {
+        closeInternal(false);
+    }
+    
+    public void closeAndAwaitInactivity() {
+        closeInternal(true);
+    }
+    
+    private void closeInternal(boolean awaitInactivity) {
         if (closeTriggered.compareAndSet(false, true)) {
-            try {
-                channelManager.close();
-            } catch (Throwable t) {
-                LOGGER.warn("Unexpected error on ChannelManager close", t);
-            }
-            if (allowStopNettyTimer) {
-                try {
-                    nettyTimer.stop();
-                } catch (Throwable t) {
-                    LOGGER.warn("Unexpected error on HashedWheelTimer close", t);
+            CompletableFuture<Void> handledCloseFuture = channelManager.close().whenComplete((v, t) -> {
+                if(t != null) {
+                    LOGGER.warn("Unexpected error on ChannelManager close", t);
                 }
+                if (allowStopNettyTimer) {
+                    try {
+                        nettyTimer.stop();
+                    } catch (Throwable th) {
+                        LOGGER.warn("Unexpected error on HashedWheelTimer close", th);
+                    }
+                }
+            });
+
+            if(awaitInactivity) {
+                handledCloseFuture = handledCloseFuture.thenCombine(awaitInactivity(), (v1,v2) -> null) ;
             }
-            
-            //see https://github.com/netty/netty/issues/2084#issuecomment-44822314
+
             try {
-                ThreadDeathWatcher.awaitInactivity(config.getShutdownTimeout(), TimeUnit.MILLISECONDS);
-            } catch(InterruptedException t) {
-                // Ignore
+                handledCloseFuture.get(config.getShutdownTimeout(), TimeUnit.MILLISECONDS);
+            } catch (InterruptedException | TimeoutException t) {
+                LOGGER.warn("Unexpected error on AsyncHttpClient close", t);
+            } catch (ExecutionException e) {
+                // already handled and could be ignored
             }
-            
             closed.compareAndSet(false, true);
         }
+    }
+
+    private CompletableFuture<Void> awaitInactivity() {
+        //see https://github.com/netty/netty/issues/2084#issuecomment-44822314
+        CompletableFuture<Void> wait1 = CompletableFuture.runAsync(() -> {
+        try {
+            GlobalEventExecutor.INSTANCE.awaitInactivity(config.getShutdownTimeout(), TimeUnit.MILLISECONDS);
+        } catch(InterruptedException t) {
+            // Ignore
+        }});
+        CompletableFuture<Void> wait2 = CompletableFuture.runAsync(() -> {
+        try {
+            ThreadDeathWatcher.awaitInactivity(config.getShutdownTimeout(), TimeUnit.MILLISECONDS);
+        } catch(InterruptedException t) {
+            // Ignore
+        }});  
+        return wait1.thenCombine(wait2, (v1,v2) -> null);
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
```diff
@@ -112,6 +115,15 @@ public void close() {
                     LOGGER.warn("Unexpected error on HashedWheelTimer close", t);
                 }
             }
+            
+            //see https://github.com/netty/netty/issues/2084#issuecomment-44822314
+            try {
+                ThreadDeathWatcher.awaitInactivity(5, TimeUnit.SECONDS);
```
slandelle: Shouldn't the await activity be aligned with `AsyncHttpClientConfig#getShutdownTimeout` or `AsyncHttpClientConfig#getShutdownQuietPeriod`?

See https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java#L309

slandelle (changes requested):
Thanks a lot!
Could you please have a look at my question?

client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
MiErnst: Yes, I guess aligning it with `AsyncHttpClientConfig#getShutdownTimeout` makes absolutely sense. The quietPeriod seems to have different meaning and I guess it is inappropriate at this place.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -300,8 +305,17 @@ public boolean removeAll(Channel connection) {
     }
 
     private void doClose() {
-        openChannels.close();
+        openChannels.close().addListener(future -> closeLatch.countDown());
         channelPool.destroy();
+        
+        //see https://github.com/netty/netty/issues/2084#issuecomment-44822314
+        try {
+            GlobalEventExecutor.INSTANCE.awaitInactivity(5, TimeUnit.SECONDS);
```
MiErnst: Same here

client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
slandelle: 👍

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -300,8 +305,17 @@ public boolean removeAll(Channel connection) {
...
+        } catch(InterruptedException t) {
+            // Ignore
+        }
+        
+        closeLatch.countDown();
```
slandelle: In finally block?

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -310,6 +324,13 @@ public void close() {
                     .addListener(future -> doClose());
         } else
             doClose();
+        
+        try {
+            closeLatch.await();
```
slandelle: No timeout for this await?

slandelle (changes requested):

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
MiErnst: yes, timeout could be reused!

client/src/main/java/org/asynchttpclient/AsyncHttpClientState.java
```diff
@@ -23,7 +23,7 @@ public AsyncHttpClientState(AtomicBoolean closed) {
         this.closed = closed;
     }
     
-    public boolean isClosed() {
+    public boolean isClosedOrClosingIsTriggered() {
         return closed.get();
```
slandelle: I guess it's best to rename `closed` into `closeTriggered`.

slandelle (changes requested):

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -597,7 +597,7 @@ public void replayRequest(final NettyResponseFuture<?> future, FilterContext fc,
     }
 
     public boolean isClosed() {
```
slandelle: Rename method too.

slandelle (changes requested):

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
MiErnst: Yes could be done to cover the corner cases.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
MiErnst: Done and attached another commit to the pull request!

twz123:
I'm just wondering... Will this interfere with other parts of an application that also use Netty? So let's say I close an AHC instance but have some other Netty-based component, e.g. a server component - or just another AHC instance - that is still in use. To me, it sounds like those global helper methods would then just run into a timeout.

slandelle:
Yeah, actually I think @twz123 is right and this would also break for people who spawn and shutdown AHC instances at runtime without stopping the application :(

MiErnst:
@twz123 thanks for the check but I can't see any interference to other instances of AHC or netty. If there is still activity on the threads the `awaitInactivity` calls triggered in the `close` method of the AHC client should run into the timeout and that's it. The close returns successfully without any exception. Maybe the close then needs a while till it returns but it should not affect other instances or did I miss something?

If we could not clarify this today we have to discuss this another time because I'm out of office for four weeks now (starting tomorrow). Sorry for this inconvenience.

twz123:
Hey @MiErnst,

> should not affect other instances

Okay, that's true. Maybe I should have written in passive voice "is affected by other parts of the app that use Netty" instead of "interferes with". My caveat is exactly what you wrote:

> close then needs a while till it returns

So the blocking behavior of `close` only works as expected when no other parts of the app are currently using Netty. This is certainly not a complete show-stopper but boils down to "just wait some amount of time and just assume that all resources have been closed concurrently by then" for those cases where Netty is still in use.

From my point of view, there are several possibilities now:

1. "Live with it!" :smile: Then I'd add a prominent note to the Javadoc of `close` that explains that behavior.
2. Have a second, blocking version of close that does exactly what you've implemented. Maybe with some `boolean blocking` parameter, or returning some `Future` that can be waited on by the caller if desired. So people can choose according to their use case.
3. Find a way that only waits for resources to be closed that actually belong to the AHC instance being closed. I have the wary feeling that this isn't possible, since those global helper methods are obviously there for a reason.
4. Don't call the static Netty methods inside `close` at all. Since they're static, one could just call them manually if desired, after the close call on the AHC instance itself.

client/src/main/java/org/asynchttpclient/AsyncHttpClientState.java
```diff
@@ -17,13 +17,13 @@
...
     
-    public AsyncHttpClientState(AtomicBoolean closed) {
-        this.closed = closed;
+    public AsyncHttpClientState(AtomicBoolean closeTriggered) {
+        this.closeTriggered = closeTriggered;
     }
     
-    public boolean isClosed() {
-        return closed.get();
+    public boolean isCloseTriggered() {
```
johnou: What's wrong with the old name?

client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
```diff
@@ -112,6 +115,15 @@ public void close() {
                     LOGGER.warn("Unexpected error on HashedWheelTimer close", t);
                 }
             }
+            
+            //see https://github.com/netty/netty/issues/2084#issuecomment-44822314
+            try {
+                ThreadDeathWatcher.awaitInactivity(config.getShutdownTimeout(), TimeUnit.MILLISECONDS);
+            } catch(InterruptedException t) {
+                // Ignore
```
johnou: Re-assert interrupted status

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -300,16 +305,30 @@ public boolean removeAll(Channel connection) {
...
     private void doClose() {
-        openChannels.close();
+        openChannels.close().addListener(future -> closeLatch.countDown());
         channelPool.destroy();
+        
+        //see https://github.com/netty/netty/issues/2084#issuecomment-44822314
+        try {
+            GlobalEventExecutor.INSTANCE.awaitInactivity(config.getShutdownTimeout(), TimeUnit.MILLISECONDS);
+        } catch(InterruptedException t) {
+            // Ignore
```
johnou: Re-assert interrupted status

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -300,16 +305,30 @@ public boolean removeAll(Channel connection) {
...
             eventLoopGroup.shutdownGracefully(config.getShutdownQuietPeriod(), config.getShutdownTimeout(), TimeUnit.MILLISECONDS)//
                     .addListener(future -> doClose());
-        } else {
+        } else
             doClose();
+
+        try {
+            closeLatch.await(config.getShutdownTimeout(), TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+           // Ignore
```
johnou: Re-assert interrupted status

zhan-ge:
Hi guys, is there a plan to release these commits?
In my scenario I have to close the specified client during runtime, therefore, I urgently need this. Thanks a lot!

hepin1989:
@zhan-ge There was changes request,so won't be any soon.and AHC has its own release cycle.

slandelle:
Frankly, I'm not fond of this implementation.
People using Netty for multiple components are probably at least 50% of AHC users, eg:
* server side HTTP layer
* whatever NoSQL database client (Cassandra, MongoDB...)
* new AWS async client
* multiple AHC instances

So having a mechanism based on this `GlobalEventExecutor.INSTANCE` singleton seems broken to me.

Why not keeping track of flying requests? Like having an `AsyncHttpClient` that would be a decorator, that would increment an AtomicInteger when executing requests, would register a Listener to decrement on completion, and that could switch to closing state (reject new requests, await for current flying requests and then close).

MiErnst:
Hey together,

sorry for the circumstances that I wasn’t able to work on this pull request. 

@twz123 
I like your comment and I would like to answer to your mentioned options:
1.	“Live with it”, yes this is possible but I think your considerations are right. 
2.	I really like this idea. I added a commit to the pull request with a second method. I have rewritten the fix to use CompletableFutures and there is only one “get” call which waits till it times out or the close completes. I have tested it in our application and it always completes within one second when calling the new method. I don’t return the future because the only thing the client could do is to call the get method itself with the client config and it’s timeout settings in mind. From an API designer perspective, I think the return type of the new method should be void.
3.	Finding a way that only waits for resources to be closed that actually belong to the AHC instance is now the default behavior of calling the existing method close. The new method closeAndAwaitInactivity additionally waits for the shared resources.
4.	Yes, this is possible but I think that other AHC users could benefit from a fix and finding these hidden API methods was a pain. :-)

As you can see in the implementation I changed the return type of the ChannelManager’s close method. Although this method is public, the instance of the ChannelManager seems only to managed by the DefaultAsyncHttpClient and is not returned by any public method of this client, so I think this internal API change should not be a problem.

The current close method also waits for the ChannelManager to complete successfully. If the eventLoopGroup is going to shutdown gracefully, the close will wait till this method completes but I think this is OK because it’s a resource created by the AHC. To prevent the amount of TimeoutExceptions in the travis build, the system property org.asynchttpclient.shutdownTimeout in the surefire plugin configuration has to be increased to 1500 because the gracefull shutdown needs at least 1.2 seconds to complete. I don’t know if this affects the travis build time to much so it is currently not part of the push request.

I don’t think the currently failed test is related to this implementation. Can’t see what the cause is.
Don’t hesitate to provide feedback for the new implementation.

@slandelle
I can’t see how keeping track of flying requests might help to wait on this shared resources of netty. A decorator would only help if this decorator is implemented around the netty stack and AHC as well as all other applications which use netty would use this decorator. I don’t think this is a realistic fix. By the new implementation, all netty resources which are not shared between AHC clients, are closed by calling close or the new method. The new method additionally waits for the shared resources of netty. With the new method users can choose which close method they want to use.

transamericamoon:
Can you guys just change it to this and be done with it, I think by the time they are closing the pool waiting sync vs async does not matter that much:

```
 private void doClose() {
    openChannels.close().awaitUninterruptibly();
    channelPool.destroy();
  }
```

PR #1438, opened 2017-07-14
Support of 308 HTTP Status

u2032:
Motivation:
The status 308 is defined by RFC7538.
This RFC has currently the state Proposed Standard since 2 years, but the status code is already handle by all browsers (Chrome, Firefox, Edge, Safari, …).

Changes:
HTTP Status 308 is added to constants
HTTP Status 308 is added on well-known statuses for redirection
When 308, according to the RFC, we are not allowed to switch to get and we keep body


slandelle:
Thanks!

PR #1442, opened 2017-07-31
Improved error messages if an invalid URL was passed

bahrmichael:
```
    @Test
    public void name() throws Exception {
        AsyncHttpClient asyncHttpClient = sut.getAsyncHttpClient();
        asyncHttpClient.prepareDelete("url");
    }
```

currently leads to

```
java.lang.NullPointerException: scheme

	at org.asynchttpclient.util.Assertions.assertNotNull(Assertions.java:23)
	at org.asynchttpclient.uri.Uri.<init>(Uri.java:64)
	at org.asynchttpclient.uri.Uri.create(Uri.java:39)
	at org.asynchttpclient.uri.Uri.create(Uri.java:32)
	at org.asynchttpclient.RequestBuilderBase.setUrl(RequestBuilderBase.java:148)
	at org.asynchttpclient.DefaultAsyncHttpClient.requestBuilder(DefaultAsyncHttpClient.java:259)
	at org.asynchttpclient.DefaultAsyncHttpClient.prepareDelete(DefaultAsyncHttpClient.java:157)
```

This may be confusing at first, so I changed the value passed from the Uri constructor to assertNotNull to be more descriptive.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/uri/Uri.java
+++ b/client/src/main/java/org/asynchttpclient/uri/Uri.java
@@ -61,7 +61,7 @@ public class Uri {
             String path,//
             String query) {
 
-        this.scheme = assertNotNull(scheme, "The scheme could not be resolved. Please provide a valid URL including the protocol.");
+        this.scheme = assertNotNull(scheme, "The scheme could not be resolved. Please provide a valid URL.");
         this.userInfo = userInfo;
         this.host = assertNotNull(host, "The host could not be resolved. Please provide a valid host.");
         this.port = port;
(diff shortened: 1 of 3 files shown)


slandelle:
Hi @bahrmichael 

Thanks, but I think the current checks are perfectly valid as they currently are and the ones you would like should actually be in the `create` static method as they are related to the parsing result.

Could you please update your PR?

Thanks!


those new/different messages are not in the right place.
They are related to the parsing, not the constructor, so they should be thrown

bahrmichael:
I will.

slandelle:
Thanks!

bahrmichael:
Thanks for reacting so fast! First open source PR merged 🎉

slandelle:
Congrats! Hopefully more will follow and you'll get used to it ;-)

Please note the follow up in #1445

PR #1447, opened 2017-08-02
Add setWebSocket0 extension method

thabach:
Add an additional template method for the `setWebSocket` method in `WebSocketUpdateHandler`.


slandelle:
Fixed up in previous commit

PR #1458, opened 2017-09-15
Immediately close channel when subscribers cancel

HKrausAxon:
This commit includes two changes:
- More aggressively mark the ResponseFuture as done
  to avoid threading issues that still read the old
  state
- Directly invoke channel closing, since there is
  no guarantee of another channelRead occuring (e.g.
  if the stream was hard closed)

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
@@ -42,16 +42,15 @@ public class StreamedResponsePublisher extends HandlerPublisher<HttpResponseBody
     protected void cancelled() {
         logger.debug("Subscriber cancelled, ignoring the rest of the body");
 
-        // The subscriber cancelled early, we need to drain the remaining elements from the stream
-        channelManager.drainChannelAndOffer(channel, future);
-        channel.pipeline().remove(StreamedResponsePublisher.class);
-
         try {
             future.done();
         } catch (Exception t) {
             // Never propagate exception once we know we are done.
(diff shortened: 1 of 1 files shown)


HKrausAxon:
Huh, this seems very strange - the test passed for me locally. Looking into it.

client/src/test/java/org/asynchttpclient/reactivestreams/ReactiveStreamsTest.java
```diff
@@ -304,22 +304,27 @@ public void streamedResponseTest() throws Throwable {
 
     @Test(groups = "standalone")
     public void cancelStreamedResponseTest() throws Throwable {
-        try (AsyncHttpClient c = asyncHttpClient()) {
-
-            // Cancel immediately
-            c.preparePost(getTargetUrl()).setBody(LARGE_IMAGE_BYTES).execute(new CancellingStreamedAsyncProvider(0)).get();
+        // Cancel immediately
+        createClientAndCancelAfterN(0);
```
slandelle: This test is wrong. You shouldn't be closing and recreating a client instance in every step.
The purpose of this test is to check how a client instance recovers from those cancellations.

HKrausAxon: Sorry, I think I misunderstood the intent of this test then and wanted to clean it up :)

HKrausAxon:
Addressed all comments by @slandelle (from both this as well as the 2.0 PR)

slandelle:
Cherry-picked as bade99201f116f1ab7f0df20cee3c188b370a512

PR #1459, opened 2017-09-15
Immediately close channel when subscribers cancel

HKrausAxon:
This commit includes two changes:
- More aggressively mark the ResponseFuture as done
  to avoid threading issues that still read the old
  state
- Directly invoke channel closing, since there is
  no guarantee of another channelRead occuring (e.g.
  if the stream was hard closed)

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
@@ -42,16 +42,15 @@ public class StreamedResponsePublisher extends HandlerPublisher<HttpResponseBody
     protected void cancelled() {
         logger.debug("Subscriber cancelled, ignoring the rest of the body");
 
-        // The subscriber cancelled early, we need to drain the remaining elements from the stream
-        channelManager.drainChannelAndOffer(channel, future);
-        channel.pipeline().remove(StreamedResponsePublisher.class);
-
         try {
             future.done();
         } catch (Exception t) {
             // Never propagate exception once we know we are done.
             logger.debug(t.getMessage(), t);
         }
+
+        // The subscriber cancelled early - this channel is dead and should be closed.
+        channelManager.closeChannel(channel);
     }
 
     NettyResponseFuture<?> future() {


client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
```diff
@@ -42,16 +42,18 @@ public StreamedResponsePublisher(EventExecutor executor, ChannelManager channelM
...
-        // The subscriber cancelled early, we need to drain the remaining elements from the stream
-        channelManager.drainChannelAndOffer(channel, future);
-        channel.pipeline().remove(StreamedResponsePublisher.class);
-
         try {
             future.done();
         } catch (Exception t) {
             // Never propagate exception once we know we are done.
-            logger.debug(t.getMessage(), t);
+            logger.error(t.getMessage(), t);
```
slandelle: Why raise logging level to error?

client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
```diff
@@ -42,16 +42,18 @@ public StreamedResponsePublisher(EventExecutor executor, ChannelManager channelM
...
         }
+
+        // The subscriber cancelled early - this channel is dead and should be closed.
+        channelManager.releaseChannelLock(future.getPartitionKey());
```
slandelle: Releasing the lock is already done on future.done() above. You're most likely double releasing here.

client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
```diff
@@ -42,16 +42,18 @@ public StreamedResponsePublisher(EventExecutor executor, ChannelManager channelM
...
+        channelManager.closeChannel(channel);
+
+        channel.pipeline().remove(StreamedResponsePublisher.class);
```
slandelle: Now that we close the channel, I think it's useless to remove the publisher from the pipeline. The channel will be garbage collected.

client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
HKrausAxon: It seemed like a pretty severe error to me - but I can reduce it back to debug.

client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
HKrausAxon: Will remove.

client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
HKrausAxon: Will remove.

HKrausAxon:
Addressed all comments by @slandelle (from this PR and master version)
--> Significantly minimized commit size :)

client/src/main/java/org/asynchttpclient/netty/handler/StreamedResponsePublisher.java
```diff
@@ -42,16 +42,15 @@ public StreamedResponsePublisher(EventExecutor executor, ChannelManager channelM
...
-
         try {
             future.done();
         } catch (Exception t) {
             // Never propagate exception once we know we are done.
             logger.debug(t.getMessage(), t);
         }
+
+        // The subscriber cancelled early - this channel is dead and should be closed.
+        channelManager.closeChannel(channel);
```
slandelle: Any reason why you moved this after completing the future?
I think it should be moved back where it was first.

HKrausAxon: Actually yes. There are other places that read the channel which check the future status.
In the previous version, it was possible that channelManager.closeChannel was invoked, the scheduler switches over to another thread, and now that other thread tries to read from the channel that is currently being closed (because its future.isDone check returned false)

Pushing the future.done() further up will avoid that potential threading issue.

slandelle:
Cherry-picked. Release 2.0.36 is under way.
Thanks!

PR #1466, opened 2017-10-02
Implemented SOCKS proxy support

Lesiuk:
Please provide feedback what you dont like or what would you change. As far as I was testing it works correctly but I'm not sure about my code style so I'm open for suggestions.


Lesiuk:
It seems that tests are failing because gatling.io is returning 301 Moved Permanently for every request on port 80. Am I right?

slandelle:
> It seems that tests are failing because gatling.io is returning 301 Moved Permanently for every request on port 80. Am I right?

Right, sorry for that (we switched today to https and plain http gets redirected).
Please let me fix the test suite. Could you please then rebase?
I'm pretty swamped atm, so I'm afraid I'll only be able to review next week. Is that OK for you?

Thanks a lot!

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -391,8 +387,49 @@ public SslHandler addSslHandler(ChannelPipeline pipeline, Uri uri, String virtua
...
 
-    public Bootstrap getBootstrap(Uri uri, ProxyServer proxy) {
-        return uri.isWebSocket() && proxy == null ? wsBootstrap : httpBootstrap;
+    public Promise<Bootstrap> getBootstrap(Request request, ProxyServer proxy) {
+        final Promise<Bootstrap> promise = ImmediateEventExecutor.INSTANCE.newPromise();
+
+        if (request.getUri().isWebSocket() && proxy == null) {
+            promise.setSuccess(wsBootstrap);
+        } else {
+            if (proxy != null && proxy.getProxyType().isSocks()) {
```
slandelle: Please collapse into `else if`

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -391,8 +387,49 @@ public SslHandler addSslHandler(ChannelPipeline pipeline, Uri uri, String virtua
...
+                request.getNameResolver().resolve(proxy.getHost()).addListener((Future<InetAddress> proxyDnsFuture) -> {
```
slandelle: s/proxyDnsFuture/whenProxyAddress

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -391,8 +387,49 @@ public SslHandler addSslHandler(ChannelPipeline pipeline, Uri uri, String virtua
...
+                            @Override
+                            public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
+                                handler.handlerAdded(ctx);
+                                super.handlerAdded(ctx);
+                            }
+
+                            @Override
+                            protected void initChannel(Channel channel) throws Exception {
+                                InetSocketAddress proxyAddress = new InetSocketAddress(proxyDnsFuture.get(), proxy.getPort());
+                                if (proxy.getProxyType() == ProxyType.SOCKS_V4) {
```
slandelle: `switch` instead of `if/else`

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -108,7 +97,8 @@ public NettyRequestSender(AsyncHttpClientConfig config,//
         ProxyServer proxyServer = getProxyServer(config, request);
 
         // websockets use connect tunnelling to work with proxies
-        if (proxyServer != null && (request.getUri().isSecured() || request.getUri().isWebSocket()) && !isConnectDone(request, future))
+        if (proxyServer != null && (request.getUri().isSecured() || request.getUri().isWebSocket()) && !isConnectDone(request, future) //
+                && !proxyServer.getProxyType().isSocks())
```
slandelle: for consistency, should be `proxyServer.getProxyType() == ProxyType.HTTP`

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -391,8 +387,49 @@ public SslHandler addSslHandler(ChannelPipeline pipeline, Uri uri, String virtua
...
+        final Promise<Bootstrap> promise = ImmediateEventExecutor.INSTANCE.newPromise();
+
+        if (request.getUri().isWebSocket() && proxy == null) {
+            promise.setSuccess(wsBootstrap);
+        } else {
+            if (proxy != null && proxy.getProxyType().isSocks()) {
+                request.getNameResolver().resolve(proxy.getHost()).addListener((Future<InetAddress> proxyDnsFuture) -> {
+                    if (proxyDnsFuture.isSuccess()) {
+                        Bootstrap bootstrap = httpBootstrap.clone();
+                        ChannelHandler handler = bootstrap.config().handler();
```
slandelle: Correct me if I'm wrong, but it looks to me that you're editing the shared `httpBootstrap`, so it would stop wrong for other requests that wouldn't have this proxy configured.
Shouldn't installing the socks proxy handler be done in `NettyConnectListener`?

client/src/main/java/org/asynchttpclient/netty/channel/NettyConnectListener.java
```diff
@@ -94,12 +106,7 @@ public void onSuccess(Channel channel, InetSocketAddress remoteAddress) {
             Object partitionKeyLock = future.takePartitionKeyLock();
 
             if (partitionKeyLock != null) {
-                channel.closeFuture().addListener(new GenericFutureListener<Future<? super Void>>() {
-                    @Override
-                    public void operationComplete(Future<? super Void> future) throws Exception {
-                        connectionSemaphore.releaseChannelLock(partitionKeyLock);
-                    }
-                });
+                channel.closeFuture().addListener(future -> connectionSemaphore.releaseChannelLock(partitionKeyLock));
```
slandelle: Agree, but not related to this feature. Please revert (I'll push it right away).

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -332,7 +326,7 @@ protected void onFailure(Throwable cause) {
         Uri uri = request.getUri();
         final Promise<List<InetSocketAddress>> promise = ImmediateEventExecutor.INSTANCE.newPromise();
 
-        if (proxy != null && !proxy.isIgnoredForHost(uri.getHost())) {
+        if (proxy != null && !proxy.isIgnoredForHost(uri.getHost()) && !proxy.getProxyType().isSocks()) {
```
slandelle: for consistency, should be proxyServer.getProxyType() == ProxyType.HTTP

client/src/test/java/org/asynchttpclient/testserver/SOCKSProxy.java
```diff
@@ -0,0 +1,192 @@
...
+
+import java.nio.ByteBuffer;
+import java.nio.channels.*;
+import java.io.*;
+import java.net.*;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.Set;
+
+public class SOCKSProxy {
```
slandelle: Please fix case to `SocksProxy`

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
slandelle: Actually, is this really needed? The point here is to honor proxy's ignoredHosts, doesn't seem related to proxy type.

client/src/main/java/org/asynchttpclient/channel/ChannelPoolPartitioning.java
```diff
@@ -23,12 +24,14 @@
         private final int proxyPort;
         private final boolean secured;
         private final String targetHostBaseUrl;
+        private final ProxyType proxyType;
 
-        public ProxyPartitionKey(String proxyHost, int proxyPort, boolean secured, String targetHostBaseUrl) {
+        public ProxyPartitionKey(String proxyHost, int proxyPort, boolean secured, String targetHostBaseUrl, ProxyType proxyType) {
```
slandelle: Is this really needed? I mean: would someone really have a SOCKS and an HTTP proxy on the same host?

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
```diff
@@ -220,7 +220,7 @@ public String toString() {
             sb.append("\t\t").append(header.getKey()).append(": ").append(header.getValue()).append("\n");
         }
         sb.append("\tbody=\n").append(getResponseBody()).append("\n")//
-                .append("}").toString();
+                .append("}");
```
slandelle: Nice catch! Not related to this change though. Will commit separately.

PR #1471, opened 2017-10-30
#1470: Microsoft has switched to HTTPS adapt PerRequestRelative302Test

### LINKED ISSUE #1470: PerRequestRelative302Test fails, since microsoft.com now uses HTTPS
Microsoft.com has been switched to HTTPS lately, so any redirect goes to 

```
https://www.microsoft.com:443
```

not 

```
http://www.microsoft.com:80
```

PerRequestRelative302Test therefore fails.

gerdriesselmann:


slandelle:
Thanks for contributing, but AHC 1 is dead, see #1470

PR #1472, opened 2017-10-30
Apache.org does not send TRACE Accept HTTP header any more

gerdriesselmann:
AsyncStreamHandlerTest expect the site www.apache.org to contain TRACE as Accept HTTP header. However this has been disabled on that site lately, so the test fails.

As a quick fix, just look for the remaining, but provided headers

- GET
- POST
- HEAD
- OPTIONS


slandelle:
Thanks for contributing, but AHC 1 is dead, see #1470

Issue #1235, opened 2016-08-23
Have a CookieJar


fangqiao:
Hi, 

I am using this lib for testing a custom REST server based on Netty 4.1. 4, the latest, using Scala.

I need to test that cookie has been set correctly at sever side and can be parsed correct on next request.

I used a single DefaultAsyncHttpClient object to submit the first round of request and it received server set cookies successfully.

```
val httpClient = new  DefaultAsyncHttpClient()

  val request : Request = (new RequestBuilder("POST")).
    setUrl("http://127.0.0.1:9090/login").
    setBody(json).
    build()

  val future : Future[Response] = httpClient.executeRequest(request)


  val response: Response = future.get()

  response.getStatusCode should equal (200)
  val cookies = response.getCookies 
```

The cookies from server were correctly returned.

Then I used the same DefaultAsyncHttpClient object to submit another request. I supposed the cookies be sent to server automatically. But the netty server was never able to get the cookies from the requests.

Then i tried in the next request by explicitly setting cookies.
      val request2: Request = (new RequestBuilder("GET")).
        setUrl("http://127.0.0.1:9090/developer").
        setCookies(cookies).
        build() 

This time the netty server can parse the cookies correctly.

I am wondering if it is by design that DefaultAsyncHttpClient will not send server side cookies in next round of request.

If so, I suggest to add the support to coming versions.

If not, what's the correct way to use this feature, which is very useful for my case.

Thanks

slandelle:
> I am wondering if it is by design that DefaultAsyncHttpClient will not send server side cookies in next round of request.

No, AHC doesn't provide a cookie jar.

> If so, I suggest to add the support to coming versions.

Contribs welcome

> If not, what's the correct way to use this feature, which is very useful for my case.

You have the responses where you can fetch the Set-Cookie header, fetch those and use them to craft the next requests.

fangqiao:
Thanks for the answers.

> You have the responses where you can fetch the Set-Cookie header, fetch those and use them to craft the next requests.

That is exactly what i am doing now. 

Maybe as interim solution, add a line to API doc or github homepage?

slandelle:
No contrib so far. Closing for now.

slandelle:
[...]

Issue #1094, opened 2016-02-16
Realm of replayed request is ignored


jufickel-b:
Hi,

with 2.0.0-RC8 there seems to be a problem regarding ``NettyRequestSender#newNettyRequestAndResponseFuture``. In a custom ``ResponseFilter`` I replay the request with a Realm set. However this realm is ignored. Setting the realm data manually as header works. Debugging showed that the afore mentioned method uses the realm of the original future instead of the realm of the (changed) request.

slandelle:
This bug is not a priority for me, as I don't use ResponseFilters in Gatling.
It would be great if you could contribute a fix.

Diagoras:
@slandelle So after mucking around with some failed tests, it looks like this is harder than I first expected. Both the ProxyUnauthorized407Interceptor and the Unauthorized401Interceptor modify the Realm and ProxyRealm on the returned future in all kinds of ways, which makes just substituting the new request's values not enough to preserve correct behavior.

This would be a lot easier if those two interceptors were ResponseFilters, which modified the new request instead of the old response. Do you see any major obstacles to converting the various interceptor classes in the Interceptors class into RequestFiliters/ResponseFilters/IOExceptionFilters? Or am I biting off more than I can chew/over-complicating this simple issue?

slandelle:
To be frank, I think that those RequestFilters/ResponseFilters/IOExceptionFilters are more or less broken.
They were introduced before I joined here, their test coverage is not sufficient in regards to the complexity they introduce. But as I don't use them, I don't give them any love, and don't plan to (way too swamped to spend time on things I don't use).
I'd like to get some serious feedback on those. I mean: what do they provide that can't be achieved with a Future/AsyncHandler pipeline? If they're just another way to achieve something, I might as well get rid of them in a future version.

> Do you see any major obstacles to converting the various interceptor classes in the Interceptors class into RequestFilters/ResponseFilters/IOExceptionFilters? 

* Breaking something that works :)
* Possibly increasing allocation and hurting performance, which would be a nay, considering I develop AHC for Gatling.

slandelle:
Then, the Realm and auth layer needs some serious refactoring too :(

Diagoras:
Hmm, a lot to think about here. Let's see:

> I'd like to get some serious feedback on those. I mean: what do they provide that can't be achieved with a Future/AsyncHandler pipeline? If they're just another way to achieve something, I might as well get rid of them in a future version.

[...]

Issue #426, opened 2013-11-27
IOExceptionFilters not Invoked on ConnectException


jlook7:
It appears that if a server is down, the IOExceptionFilter does nothing to attempt reconnecting to the server.  I also have also set request retries to Integer.MAX_VALUE.

2013-11-26 18:42:24,011 [New I/O boss #17] DEBUG NettyAsyncHttpProvider  Unexpected I/O exception on channel [id: 0xc09e06ee]
java.net.ConnectException: Connection refused: stunspun.corp.gq1.yahoo.com/10.88.182.226:4080
    at sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
    at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:735)
    at org.jboss.netty.channel.socket.nio.NioClientBoss.connect(NioClientBoss.java:150)
    at org.jboss.netty.channel.socket.nio.NioClientBoss.processSelectedKeys(NioClientBoss.java:105)
    at org.jboss.netty.channel.socket.nio.NioClientBoss.process(NioClientBoss.java:79)
    at org.jboss.netty.channel.socket.nio.AbstractNioSelector.run(AbstractNioSelector.java:312)
    at org.jboss.netty.channel.socket.nio.NioClientBoss.run(NioClientBoss.java:42)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    at java.lang.Thread.run(Thread.java:724)

Ideally, there would be a feature to enable exponential backoff on disconnects to maintain a connection between the client and server.

amirhmd:
hi there Im new to open source and would love to contribute to AsyncHttpClient. Can anyone help me get started? Or is AsyncHttpClient for people who are more experienced in open source?

tsteff:
Shouldn't this be a defect? the point of the filter is to be triggered on IOException.
I was hoping to use the IOExceptionFilter for closing and submitting spans for distributed tracing if using this client.

slandelle:
@amirhmd Best way is to dig into the code, get familiar with it, and get a stab at fixing an issue, possibly one you hit. I can help with Pull Request review.

[...]

PR #1491, opened 2018-01-07
Added handling of httpOnly and secure flags of cookies when redirecting

unoexperto:
Certain calls work incorrectly due to overridden cookies. 

For example GET request of 
http://www.thelancet.com/journals/lancet/article/PIIS1474-4422(17)30433-7
consists of eight redirects and all of them are HTTPS calls.

Response from 3rd redirect to `id.elsevier.com` returns `Set-Cookie` for `JSESSION` with `httpOnly` flag. But `JSESSION` is already set. This response overrides it which breaks subsequent call to `secure.jbs.elsevierhealth.com`.

As result client is being redirected to front page of the website.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
@@ -129,8 +129,10 @@ public class Redirect30xInterceptor {
 
                 for (String cookieStr : responseHeaders.getAll(SET_COOKIE)) {
                     Cookie c = ClientCookieDecoder.STRICT.decode(cookieStr);
-                    if (c != null)
-                        requestBuilder.addOrReplaceCookie(c);
+                    if (c != null) {
+                        if ((!c.isHttpOnly() && !c.isSecure()) || (c.isHttpOnly() && !newUri.isSecured()) || (c.isSecure() && newUri.isSecured()))
+                            requestBuilder.addOrReplaceCookie(c);
+                    }
                 }
 
                 boolean sameBase = isSameBase(request.getUri(), newUri);


slandelle:
Hi,

Thanks for trying to contribute. Sadly, your fix is wrong. The wrong JSESSIONID cookie is being sent because AHC doesn't currently have a cookie jar implementing [RFC6265](https://tools.ietf.org/html/rfc6265). httpOnly attribute has nothing to do with this (makes a cookie not reachable with Javascript).

CookieJar feature request was logged as #1235. I recently closed it due to lack of activity.
Feel free to give it a try!

unoexperto:
thank you for quick response, @slandelle ! Could you please clarify what would be different in handling my example in case of implementation of `CookieJar` ? Do you expect `httpOnly` version of cookie to survive redirects to HTTPS urls in case it ever eds up on HTTP resource ? But `CookieKey` in your implementation doesn't have httpOnly flag. Even if `CookieJar` used here it will override correct JSESSION. What do you think ?

slandelle:
@unoexperto Please take the time to read RFC6265, in particular the rules about domain and path matching. Also please properly read specification of `Secure` and  `HttpOnly` attributes. "HttpOnly" blocks non-HTTP ways, typically Javascript.

unoexperto:
@slandelle Perhaps I misinterpret the RFC. Consider following statement: `The HttpOnly attribute limits the scope of the cookie to HTTP requests.  In particular, the attribute instructs the user agent to omit the cookie when providing access to cookies via "non-HTTP" APIs (such as a web browser API that exposes cookies to scripts).`

In case redirection handler AHC is the agent. Thus it should not use `httpOnly` cookie when it opens non-HTTP url which is exactly what I do in my PR. How would you `CookieJar` handle this situation is your `CookieKey` can't distinguish between cookie with `httpOnly` flag and without ?

slandelle:
> Perhaps I misinterpret the RFC

Indeed

> In case of redirection handler AHC is the agent. Thus it should not use httpOnly=true cookie when it opens non-HTTP url

You get it wrong. A redirect is an HTTP request. A "non-HTTP" API is for example Javascript's `document.cookie`.

unoexperto:
> A redirect is an HTTP request. A "non-HTTP" API is for example Javascript's document.cookie.

It doesn't matter that it's HTTP request. From server's prospective it's an agent whether it's your code or browser with JS interpreter. It wants certain cookie to be present which is curently lost.

I'm ready to spend my personal time and implement cookie jar for this project but tell me exactly step-by-step what behavior for downloading http://www.thelancet.com/journals/lancet/article/PIIS1474-4422(17)30433-7  you consider correct handling.

Here is dump of correct and incorrect communication
https://transfer.sh/ema6C/lancet_valid_and_invalid_calls.chls
Session file can be opened with www.charlesproxy.com

Feel free to answer in French if it's more convenient for you.

slandelle:
Please, try reading the spec. HttpOnly is meant so to block cookie stealing with malicious Javascript.

AHC doesn't support fetching your page because you set 2 different cookies, both named JSESSIONID, on 2 different domains. As AHC doesn't currently have a CookieJar that implements domain matching, it gets it wrong.

If you want to contribute, there's everything you need in the Gatling implementation. You can also check the [test](https://github.com/gatling/gatling/blob/master/gatling-http/src/test/scala/io/gatling/http/cookie/CookieJarSpec.scala).

unoexperto:
@slandelle Actually now I see that you're right. I'll ping you when I'm done with integrating `CookieJar`.

slandelle:
Cool!

As I said, the main difference is an implementation for AHC would need to be mutable and threadsafe, typically thanks to a ConcurrentHashMap, on contrary to the Gatling's one whose thread safety is guaranteed by Gatling's model.

unoexperto:
@slandelle Got it. Could you create https://gitter.im room for any of your projects for easier communication in future ?

slandelle:
I don't plan to have chat rooms for my projects: AHC is my pet project, and I don't have time to provide free support, and Gatling is the means of making a living for my company, so we provide support for our customers.

unoexperto:
@slandelle Following is not final implementation but demonstration of approach I'd like to take. I looked at the AHC code and I think the least intrusive change would be to remove cookie manipulations from `Redirect30xInterceptor` and instead make sure `RequestFilter`-s are called before executing redirected requests.

Here is the change
https://github.com/unoexperto/async-http-client/commit/94e19b0e870fff88f17c8f03bc32f1fc899f13db

Then in my client code I I added one `RequestFilter` and `ResponseFilter` that catches and sets cookies like this (code is not thread-safe and just for purpose of demonstration)

```
object Test {
  var store = CookieJar.Empty
}

object SetRequestCookiesFilter extends RequestFilter {
  override def filter[T](ctx: FilterContext[T]): FilterContext[T] = {
    println(s"ReqFilter: ${ctx.getRequest.getUrl} [cookies: ${ctx.getRequest.getCookies}]")
    val items: List[Cookie] = Test.store.get(ctx.getRequest.getUri)
    if (items.nonEmpty) {
      val newRequest = new RequestBuilder(ctx.getRequest).setCookies(items.asJava).build()
      new FilterContextBuilder[T](ctx).request(newRequest).build()
    } else
      ctx
  }
}


object CatchResponseCookiesFilter extends ResponseFilter {
  override def filter[T](ctx: FilterContext[T]): FilterContext[T] = {
    println(s"RespFilter: ${ctx.getRequest.getUrl} [status: ${ctx.getResponseStatus}; headers: ${ctx.getResponseHeaders.entries().size()}]")

    val cookies = ctx.getResponseHeaders.getAll(HttpHeaderNames.SET_COOKIE).asScala.map(ClientCookieDecoder.STRICT.decode)
    if (cookies.nonEmpty)
      Test.store = Test.store.add(ctx.getRequest.getUri, cookies.toList)

    ctx
  }
}
```

Other things I may do:

1) Merge `RequestFilter` and `ResponseFilter` into single interface because they look alike;
2) For simplicity define `CookieJarLike` interface instead of implementing two above which would manage cookies

```
trait CookieJarLike {
  def storeCookies(requestUri: Uri, cookies: List[Cookie])
  def getCookies(requestUri: Uri): List[Cookie]
}
``` 

What do you think ?

slandelle:
> make sure RequestFilter-s are called before executing redirected requests.

There's a chance those Filters (Request, Response, IOExceptionFilter) will go away or will be redesigned in a future major release:

* I never use them personally (they were introduced way before I joined here)
* I never get contributions to help fixing bugs there (see #1094 and #426)
* They introduce allocation overhead

Atm, I'd rather not have implementation rely on something that might go away.

> Merge RequestFilter and ResponseFilter into single interface because they look alike;

Sorry, but that's not possible to break API now as I'm about to release 2.1.0 stable. 

> For simplicity define CookieJarLike interface instead of implementing two above which would manage cookies

Yep, just please code in Java here, not in Scala, so other people might be more inclined to join the discussion.

unoexperto:
@slandelle I see. So would you prefer more then if I instead add `CookieJarLike getCookieStorage()` to `AsyncHttpClientConfig` and call it from `Redirect30xInterceptor.exitAfterHandlingRedirect`, `DefaultAsyncHttpClient.executeRequest` and `ResponseFiltersInterceptor.exitAfterProcessingFilters` ?

slandelle:
Yep

unoexperto:
@slandelle All right. Here we go :)

https://github.com/unoexperto/async-http-client/commit/746ff0d15cb092fa82486260d78f343e75709cda

I decided not to invent the wheel so I took `java.net.InMemoryCookieStore` and adopted it to use `org.asynchttpclient.uri.Uri` and `io.netty.handler.codec.http.cookie.Cookie`.

I also think I should need to implement `Comparable<Uri>` in your `Uri`.

Let me know what you think.

slandelle:
> I took java.net.InMemoryCookieStore and adopted it to use org.asynchttpclient.uri.Uri and io.netty.handler.codec.http.cookie.Cookie

I'm sorry, but I don't feel comfortable with this.
`java.net.InMemoryCookieStore` has an Oracle copyright and is under GPLv2 license, so that's a big no.
I think translating Gatling's implementation to Java and use a ConcurrentHashMap should be pretty straightforward.

slandelle:
FYI, I'll be releasing AHC 2.1 stable once this lands.

unoexperto:
@slandelle No worries. I can rewrite it tomorrow. How do you feel about rest of changes and `CookieStore` ?

slandelle:
Regarding CookieStore, I'd rather keep the interface minimal first, ie only `add` and `get`, with the methods that we really needs internally, and then maybe extends when/if we get feedback.

Then, there's a problem with adding the cookies from the CookieStore into the request in `executeRequest`: it's performed after the `SignatureCalculator` with is performed when builder the request. But sadly that's something that can't be fixed atm. Will do when redesigning in next major release.

So, except for trimming CookieStore interface down, LGTM

PR #1495, opened 2018-01-11
Implemented cookie jar and its use in http redirection

unoexperto:
Motivation:

Existing implementation doesn't have proper support of cookie jar. It causes problem when requesting http URL that redirects to another location.

Modifications:

Implemented configurable CookieStore and default thread-safe implementation of it. Also modified redirection interceptor to use it.

Result:

Cookie store will be used for all http requests.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
+++ b/client/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
@@ -20,6 +20,7 @@ import java.util.function.Consumer;
 
 import org.asynchttpclient.channel.ChannelPool;
 import org.asynchttpclient.channel.KeepAliveStrategy;
+import org.asynchttpclient.cookie.CookieStore;
 import org.asynchttpclient.filter.IOExceptionFilter;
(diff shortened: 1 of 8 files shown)


slandelle:
Thanks a lot!

unoexperto:
@slandelle My pleasure! Let me know if you need help with another feature or something.

slandelle:
I've just merged another PR that introduces SOCKS proxies support.
I'll release 2.1.0 on Monday.

> Let me know if you need help with another feature or something.

Help is always welcome :) Doc, pending issues/features, whatever...

unoexperto:
@slandelle Just saw that you no longer can maintain the project. I'd hate this library to die. It was and is so useful to me. I can be maintainer if you're ok with it. I don't know how else to contact you but we should probably discuss details in some chat. Maybe in Telegram?

qinghuoptum:
This PR use the content of CookieStore to override the header("Cookie"). Instead of doing override, can we MEGER CookieStore to existing header("Cookie") if it exists.

PR #1504, opened 2018-01-24
Set timer thread name

zhurs:
Right now, if a timer is not supported, it's thread will be created with Executors.defaultThreadFactory, so name is indistinguishable, like pool-8-thread-1.
This patch sets name, based on threadPoolName with "-timer" suffix, so the purpose of the thread will be clear.


slandelle:
Thanks!

Issue #1503, opened 2018-01-23
Setter for connectionPoolCleanerPeriod


agapic:
Is there a particular reason that connectionPoolCleanerPeriod has no setter? If there's no reason, I can submit a PR for this, as I'd like to personally use it.

Currently, it's being set as 
`private int connectionPoolCleanerPeriod = defaultConnectionPoolCleanerPeriod();`

slandelle:
> If there's no reason, I can submit a PR for this, as I'd like to personally use it.

That's a mistake. Yes, please do.

agapic:
Sure. I'll do it.

slandelle:
Closed by #1506

PR #1506, opened 2018-01-24
#1503: Add setter for connectionPoolCleanerPeriod in AsyncHttpClient Builder

agapic:
Adding a setter for connectionPoolCleanerPeriod.

See issue here: #1503


slandelle:
Thanks!

PR #1508, opened 2018-01-29
Fixed #1507. Made socks proxy work with https.

### LINKED ISSUE #1507: Socks proxy does not work with https
Attempt to use SOCKS v4/v5 proxy when connecting to a secure (https) URL leads to java.nio.channels.ClosedChannelException. This is caused by the fact that SslHandler is added to the netty channel pipeline before Socks5ProxyHandler/Socks4ProxyHandler, while it must be added after.

daemas:
A quick (and possibly dirty) fix. This fix ensures the SslHandler is added to the netty channel pipeline AFTER the Socks4ProxyHandler/Socks5ProxyHandler. Without it any attempt to send a request to a https URL via a socks proxy ends with a java.nio.channels.ClosedChannelException.

I am not familiar with Java/maven/JUnit as I primarily write on scala/sbt/scalatest, so I cannot provide an automatic test for this, but I tried to test this manually from within my scala project - and it worked. Also I am not sure about HTTPS support over HTTP proxy (with HTTP CONNECT tunneling) as I do not have it available. This might require further investigation, tests and fixes. But socks proxy supports HTTPS with this fix for sure.

P.S. Sorry for my English, I'm not a native speaker ;)

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
@@ -359,7 +359,7 @@ public class ChannelManager {
     }
   }
 
-  public SslHandler addSslHandler(ChannelPipeline pipeline, Uri uri, String virtualHost) {
+  public SslHandler addSslHandler(ChannelPipeline pipeline, Uri uri, String virtualHost, boolean hasSocksProxyHandler) {
     String peerHost;
     int peerPort;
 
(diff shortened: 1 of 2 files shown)


slandelle:
Thanks! Could you edit your commit message to match [this template](https://netty.io/wiki/writing-a-commit-message.html), please?

slandelle:
BTW, in order to run tests, you just need to have maven installed and then run `mvn test`.

daemas:
Yep, I figured out how to run, I just do not have enough time to figure out how to write them properly under java/JUnit, as I am SBT/Scala guy :) Btw 2 or 3 tests fail on my machine for some reason, so I used `mvn install -DskipTests=true` to pull it to my local mvn repo and fetch it from there for my scala project. But I can see all tests pass under Travis/CI. This is kind of strange.

As per commit message, I'll try to do my best :) Can I just use the issue description in the Motivation section?

slandelle:
Thanks!

Issue #1238, opened 2016-08-26
Consider java.util.Base64 as a replacement of AHC class


slandelle:
Moreover, there's different types of Base64, we should check which one is supposed to be used and where.

PR #1516, opened 2018-02-09
#1238 replace AHC Base64 with java.util.Base64

doom369:
Fixes https://github.com/AsyncHttpClient/async-http-client/issues/1238


slandelle:
No way I'll ever merge that without benchmarks, my friend!
Also, there are different flavors of Base64 and I'm not even sure of which should be used where...

doom369:
@slandelle hah, let me prepare benchmark.

> Also, there are different flavors of Base64 and I'm not even sure of which should be used where...

Comment in removed Base64 says it RFC 2045. However, that is not true, as RFC 2045 uses \r\n. So used default that is RFC4648.

doom369:
@slandelle here you go:

```
Benchmark                          Mode  Cnt    Score    Error  Units
Base64PerfTest.AHC_BASE64_decode   avgt   20  102.123 ±  2.570  ns/op
Base64PerfTest.AHC_BASE64_encode   avgt   20  205.752 ± 11.434  ns/op
Base64PerfTest.JAVA_BASE64_decode  avgt   20  123.675 ± 18.390  ns/op
Base64PerfTest.JAVA_BASE64_encode  avgt   20   74.210 ±  3.350  ns/op
```

Benchmark:

```
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Thread)
@Fork(1)
@Warmup(iterations = 5, time = 1)
@Measurement(iterations = 20, time = 1)
public class Base64PerfTest {

    private String d = "TlRMTVNTUAACAAAAAAAAACgAAAABggAAU3J2Tm9uY2UAAAAAAAAAAA==";
    private byte[] e = Base64.getDecoder().decode(d);

    @Benchmark
    public byte[] AHC_BASE64_decode() {
        return Base64AHC.decode(d);
    }

    @Benchmark
    public byte[] JAVA_BASE64_decode() {
        return Base64.getDecoder().decode(d);
    }

    @Benchmark
    public String AHC_BASE64_encode() {
        return Base64AHC.encode(e);
    }

    @Benchmark
    public String JAVA_BASE64_encode() {
        return Base64.getEncoder().encodeToString(e);
    }
}
```

slandelle:
> Comment in removed Base64 says it RFC 2045. However, that is not true, as RFC 2045 uses \r\n.

Yeah, I suspected that Base64 AHC usage was a mess... :(
Usage should be revised on every use case: NTML, OAuth, Basic and WebSocket keys.

> JDK 9.0.1, VM 9.0.1+11
> I think this is very good result so AHC can move to JDK Base64.

I'll run this benchmark on JDK 8 too.

doom369:
> I'll run this benchmark on JDK 8 too.

Sure. Waiting for your results.

slandelle:
Benchmarks on my machine:

Java(TM) SE Runtime Environment (build 1.8.0_152-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.152-b16, mixed mode)

```
Benchmark                          Mode  Cnt    Score   Error  Units
Base64PerfTest.AHC_BASE64_decode   avgt   20   88,274 ± 1,054  ns/op
Base64PerfTest.AHC_BASE64_encode   avgt   20  188,539 ± 3,086  ns/op
Base64PerfTest.JAVA_BASE64_decode  avgt   20  182,451 ± 2,861  ns/op
Base64PerfTest.JAVA_BASE64_encode  avgt   20  114,778 ± 1,689  ns/op
```

Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)

```
Benchmark                          Mode  Cnt    Score    Error  Units
Base64PerfTest.AHC_BASE64_decode   avgt   20   82,235 ±  1,504  ns/op
Base64PerfTest.AHC_BASE64_encode   avgt   20  231,056 ± 24,291  ns/op
Base64PerfTest.JAVA_BASE64_decode  avgt   20  125,373 ±  2,797  ns/op
Base64PerfTest.JAVA_BASE64_encode  avgt   20   71,306 ±  0,650  ns/op
```

slandelle:
So it seems Java 9 got better Base64 encoding.

slandelle:
Anyway, let's do this, as I'm more confident in Java's implementation atm.

slandelle:
thanks!

doom369:
Yeah. That is good move, especially with new Java release circle and future performance improvements.

slandelle:
> Yeah. That is good move, especially with new Java release circle and future performance improvements.

Java 9 improved things and broke others, eg most parsers are doing worse because of the new String implementation.

Issue #1515, opened 2018-02-09
Add generic 'prepare(String method, String url)' method to AsyncHttpClient


doom369:
Right now I have next code:

```
    private BoundRequestBuilder buildRequestMethod(SupportedWebhookMethod method, String url) {
        switch (method) {
            case GET :
                return httpclient.prepareGet(url);
            case POST :
                return httpclient.preparePost(url);
            case PUT :
                return httpclient.preparePut(url);
            case DELETE :
                return httpclient.prepareDelete(url);
        }
    }
```

I would like to simply use:

```
    private BoundRequestBuilder buildRequestMethod(SupportedWebhookMethod method, String url) {
        return httpclient.prepare(method.toString(), url);
    }
```

Please let me know if you are ok with it. I'm ready to send PR.

slandelle:
Sure. I never realized it was missing as I use unbound requests in Gatling ;-)

PR welcome!

doom369:
@slandelle thanks!

PR #1518, opened 2018-02-09
#1515 added generic prepare(String method, String url) method to AsynsHttpClient

doom369:
Fixes https://github.com/AsyncHttpClient/async-http-client/issues/1515


slandelle:
I'll have to bump version to 2.4.0 for this one as it introduced a minor breaking API change.

slandelle:
thanks!

PR #1520, opened 2018-02-11
Fix missing read timeout in RequestBuilder

carlanton:
Currently it's not possible to set readTimeout on a `Request` when using `httpClient.prepareRequest`. This PR should fix that.

My current workaround is to use `httpClient.executeRequest` which doesn't seem to rebuild the request.


slandelle:
Nice catch, thanks!

PR #1522, opened 2018-02-17
Handle asynchronous calls to onSubscribe

rossabaker:
Publishers are not required to call `subscriber.onSubscribe` synchronously from `subscribe`.  When they don't, `delayedStart()` is called with a null `subscription`.  This modification allows `onSubscribe` to proceed as normal in case no subscription was yet available at the time of `delayedStart()`.

See 391b4d8e0eced22e56f390d4c5ee8a51c1c98d27 for context on why `delayedStart()` is necessary.

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -68,5 +68,10 @@
       <artifactId>rxjava</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.reactivestreams</groupId>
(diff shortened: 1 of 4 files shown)


rossabaker:
Is the build failure a known intermittent error?  I don't see an obvious link to this change, and it passes locally.

slandelle:
Thanks!

Will investigate the build failure, not related indeed.

Issue #1531, opened 2018-03-05
Retrofit module. Content-Type header is not passed


zeldigas:
Configuration:
```xml
        <dependency>
            <groupId>com.squareup.retrofit2</groupId>
            <artifactId>retrofit</artifactId>
            <version>2.3.0</version>
        </dependency>
        <dependency>
            <groupId>org.asynchttpclient</groupId>
            <artifactId>async-http-client</artifactId>
            <version>2.4.3</version>
        </dependency>
        <dependency>
            <groupId>org.asynchttpclient</groupId>
            <artifactId>async-http-client-extras-retrofit2</artifactId>
            <version>2.4.3</version>
        </dependency>
```
I found that content type header is not passed properly to AHC's request. The problem is that content-type is contained in Body object (retrofit object), not in headers list, but this field is ignored in `createRequest` method of org.asynchttpclient.extras.retrofit.AsyncHttpClientCall

slandelle:
@zeldigas Would you mind providing a PR? Extra modules are maintained by the community and I'm not a Retrofit user myself.

zeldigas:
Sure, I'll do it this week.

slandelle:
Cool, thanks!

zeldigas:
PR created, please check and let me know if I need improve it somehow

PR #1533, opened 2018-03-07
Content-Type fixes for reqest and response

zeldigas:
PR contains 2 fixes
1. Fix for content-type header in request (#1531)
2. Fix for content-type header nullability in response


slandelle:
Thanks!

Issue #1536, opened 2018-04-05
ahc-default.properties causes conflicts in shading


wsargent:
Play-WS will shade AHC's version of Netty to ensure that other Netty libraries can be used i.e. Netty 3.x can be used with AHC Netty 4.x.  

However, AHC points to /ahc-default.properties, which is hardcoded to point to the root resource.  This has caused issues when people include libraries that depend on older versions of AHC, such as Dispatch or Cassandra.

https://github.com/playframework/play-ws/issues/229

The ideal solution would be for AHC to do some introspection on the package and call up the current package, and then refer to `<package-name>/ahc-defaults.properties` -- that way it would be guaranteed local to whereever it was shaded.

slandelle:
PR? ;-)

wsargent:
@slandelle done

wsargent:
One thing to note -- I don't know how this is affected in OSGi environments, i.e. https://github.com/AsyncHttpClient/async-http-client/pull/1345

PR #1538, opened 2018-04-05
Use package relative resources

wsargent:
Fixes https://github.com/AsyncHttpClient/async-http-client/issues/1536

Use package relative resource loading with `class.getResourceAsStream`, which helps shading when there are multiple versions of AHC.


slandelle:
Thanks!

PR #1557, opened 2018-07-09
Allow to use channel manager provided via config

beargummy:
Sometimes it could be useful to provide custom ChannelManager and/or reuse it, like it is done with Timer.
Also, to support this feature, method `releaseChannelLock` is being reused in `DefaultChannelGroup#remove`, since logic is effectively the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,9 @@ AsyncHttpClientConfig cf = new DefaultAsyncHttpClientConfig.Builder()
 AsyncHttpClient c = new DefaultAsyncHttpClient(cf);
 ```
 
+It's possible to create shared resources (EventLoop, Timer, ChannelManager) beforehand and pass them to multiple client instances in the config. 
+You'll then be responsible for closing those shared resources.
+
 ## WebSocket
 
 Async Http Client also support WebSocket by simply doing:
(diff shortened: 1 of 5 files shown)


slandelle:
Sorry, but that's not possible. `ChannelManager` is an internal, not a public API, so it can can any time. Please explain your actual use case.

beargummy:
@slandelle actual use case is to be able to use custom channel(connections) limitations, not only global and per-host. This is partially achievable via custom `ChannelPoolPartitioning` with necessary logic, but it lacks custom limits. E.g. I want to limit per-host connections to some huge number of connections and also add different (lower) custom limit based on some request parameters.

slandelle:
Looks like you need something related to ConnectionSemaphore. Feel free to suggest API changes for your use case.

beargummy:
I'd like to, and I definitely will, but ConnectionSemaphore is part of 2.4.x version, and unfortunately I cannot migrate to this version due to open [netty's issue](https://github.com/netty/netty/issues/6706). That's why I want to provide custom ChannelManager.

slandelle:
IIRC, issue #6706 has actually always been there and is not related to a Netty upgrade.

PR #1563, opened 2018-07-18
Workaround for netty chunked upload issue

beargummy:
Motivation:
Netty doesn't provide any ability to handle early server response when uploading chunked file and instead, `IOException: Pipe closed` is thrown (see https://github.com/netty/netty/issues/6706 for details).

Changes:
NettyInputStreamBody swallows ChannelOutputShutdownException in case it is thrown from netty's ChunkedWriteHandler.

Result:
Multipart early response is available to be handled properly.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/future/StackTraceInspector.java
+++ b/client/src/main/java/org/asynchttpclient/netty/future/StackTraceInspector.java
@@ -15,6 +15,8 @@ package org.asynchttpclient.netty.future;
 import java.io.IOException;
 import java.nio.channels.ClosedChannelException;
 
+import io.netty.channel.socket.ChannelOutputShutdownException;
+
 public class StackTraceInspector {
 
   private static boolean exceptionInMethod(Throwable t, String className, String methodName) {
@@ -37,6 +39,11 @@ public class StackTraceInspector {
     return t instanceof ClosedChannelException
             || exceptionInMethod(t, "io.netty.handler.ssl.SslHandler", "disconnect")
             || (t.getCause() != null && recoverOnConnectCloseException(t.getCause()));
+  }
+
+  public static boolean recoverOnChunkedUploadFailed(Throwable t) {
+    return t instanceof ChannelOutputShutdownException
+          && exceptionInMethod(t, "io.netty.handler.stream.ChunkedWriteHandler", "flush");
   }
 
   public static boolean recoverOnReadOrWriteException(Throwable t) {
--- a/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
@@ -20,6 +20,7 @@ import io.netty.handler.stream.ChunkedWriteHandler;
 import org.asynchttpclient.AsyncHttpClientConfig;
 import org.asynchttpclient.netty.NettyResponseFuture;
 import org.asynchttpclient.netty.channel.ChannelManager;
+import org.asynchttpclient.netty.future.StackTraceInspector;
 import org.asynchttpclient.netty.request.WriteProgressListener;
 import org.asynchttpclient.request.body.Body;
 import org.asynchttpclient.request.body.RandomAccessBody;
@@ -77,9 +78,17 @@ public class NettyBodyBody implements NettyBody {
 
     channel.write(msg, channel.newProgressivePromise())
             .addListener(new WriteProgressListener(future, false, getContentLength()) {
+              @Override
               public void operationComplete(ChannelProgressiveFuture cf) {
                 closeSilently(body);
                 super.operationComplete(cf);
+              }
+
+              @Override
+              protected boolean abortOnThrowable(Channel channel, Throwable cause) {
+                // FIXME dirty hack until netty issue is not resolved, see https://github.com/netty/netty/issues/6706
+                return StackTraceInspector.recoverOnChunkedUploadFailed(cause)
+                      || super.abortOnThrowable(channel, cause);
               }
             });
     channel.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT, channel.voidPromise());
--- a/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
@@ -15,12 +15,12 @@ package org.asynchttpclient.netty.request.body;
 
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelProgressiveFuture;
+import io.netty.channel.socket.ChannelOutputShutdownException;
 import io.netty.handler.codec.http.LastHttpContent;
 import io.netty.handler.stream.ChunkedWriteHandler;
 import org.asynchttpclient.AsyncHttpClientConfig;
 import org.asynchttpclient.netty.NettyResponseFuture;
 import org.asynchttpclient.netty.channel.ChannelManager;
-import org.asynchttpclient.netty.future.StackTraceInspector;
 import org.asynchttpclient.netty.request.WriteProgressListener;
 import org.asynchttpclient.request.body.Body;
 import org.asynchttpclient.request.body.RandomAccessBody;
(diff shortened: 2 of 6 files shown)


client/src/main/java/org/asynchttpclient/netty/future/StackTraceInspector.java
```diff
@@ -39,6 +41,23 @@ public static boolean recoverOnNettyDisconnectException(Throwable t) {
...
   }
 
+  public static boolean recoverOnChunkedUploadFailed(Throwable t) {
+    if (t instanceof ChannelOutputShutdownException) {
+      try {
+        for (StackTraceElement element : t.getStackTrace()) {
+          String className = element.getClassName();
+          String methodName = element.getMethodName();
+          if (className.contains("ChunkedWriteHandler") && (methodName.equals("flush") || methodName.equals("write")))
+            System.exit(1);
```
slandelle: o_0

beargummy: lol... sorry

beargummy: anyway build will fail due to some issues:
> Dynamic Oracle JDK installations are currently unavailable due to a problem with our upstream provider. In the meantime, container-based builds come with JDKs preinstalled.

beargummy:
> The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $TRAVIS_APT_OPTS install oracle-java8-installer" failed and exited with 100 during .

not my day :D

slandelle:
@beargummy Any reason why you closed this PR? Is it broken?

beargummy:
I found one more case which also affected by this issue (BodyChunkedInput)

client/src/main/java/org/asynchttpclient/netty/future/StackTraceInspector.java
```diff
@@ -39,6 +41,22 @@ public static boolean recoverOnNettyDisconnectException(Throwable t) {
...
+  public static boolean recoverOnChunkedUploadFailed(Throwable t) {
+    if (t instanceof ChannelOutputShutdownException) {
+      try {
+        for (StackTraceElement element : t.getStackTrace()) {
+          String className = element.getClassName();
+          String methodName = element.getMethodName();
+          if (className.contains("ChunkedWriteHandler") && (methodName.equals("flush") || methodName.equals("write")))
+            return true;
+        }
+      } catch (Throwable ignore) {
```
slandelle: What's this try/catch for? Are you sure you want to catch Throwable???

client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
```diff
@@ -56,6 +59,14 @@ public void write(final Channel channel, NettyResponseFuture<?> future) {
     if (body instanceof RandomAccessBody && !ChannelManager.isSslHandlerConfigured(channel.pipeline()) && !config.isDisableZeroCopy()) {
       msg = new BodyFileRegion((RandomAccessBody) body);
 
+      channel.write(msg, channel.newProgressivePromise())
+            .addListener(new WriteProgressListener(future, false, getContentLength()) {
+              @Override
+              public void operationComplete(ChannelProgressiveFuture cf) {
+                closeSilently(body);
```
slandelle: Why do you need that? I would expect Netty to close the ChunkedInput.

client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
```diff
@@ -56,6 +59,14 @@ public void write(final Channel channel, NettyResponseFuture<?> future) {
     if (body instanceof RandomAccessBody && !ChannelManager.isSslHandlerConfigured(channel.pipeline()) && !config.isDisableZeroCopy()) {
       msg = new BodyFileRegion((RandomAccessBody) body);
```
slandelle: Is there any reason you only fixed for ChunkedInput and not for FileRegion?

client/src/main/java/org/asynchttpclient/netty/future/StackTraceInspector.java
```diff
@@ -39,6 +41,22 @@ public static boolean recoverOnNettyDisconnectException(Throwable t) {
...
+      }
+    }
+
+    return t.getCause() != null && recoverOnChunkedUploadFailed(t.getCause());
```
slandelle: Did you check which Exception you actually get? Do you really need to recurse?

client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
```diff
@@ -73,15 +84,25 @@ public void onError(Throwable t) {
...
                 closeSilently(body);
                 super.operationComplete(cf);
               }
+
+              @Override
+              protected boolean abortOnThrowable(Channel channel, Throwable cause) {
+                // FIXME dirty hack until netty issue is not resolved, see https://github.com/netty/netty/issues/6706
+                if (cause instanceof IOException && StackTraceInspector.recoverOnChunkedUploadFailed(cause))
+                  return false;
+                return super.abortOnThrowable(channel, cause);
```
slandelle: merge into one single boolean op

client/src/main/java/org/asynchttpclient/netty/request/body/NettyInputStreamBody.java
```diff
@@ -69,10 +70,19 @@ public void write(Channel channel, NettyResponseFuture<?> future) throws IOExcep
...
                 closeSilently(is);
                 super.operationComplete(cf);
               }
+
+              @Override
+              protected boolean abortOnThrowable(Channel channel, Throwable cause) {
+                // FIXME dirty hack until netty issue is not resolved, see https://github.com/netty/netty/issues/6706
+                if (cause instanceof IOException && StackTraceInspector.recoverOnChunkedUploadFailed(cause))
+                  return false;
+                return super.abortOnThrowable(channel, cause);
```
slandelle: merge into one single boolean op

client/src/main/java/org/asynchttpclient/netty/future/StackTraceInspector.java
beargummy: You're right, there is no need for recursive call.

client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
```diff
@@ -77,10 +78,18 @@ public void onError(Throwable t) {
...
+              @Override
               public void operationComplete(ChannelProgressiveFuture cf) {
                 closeSilently(body);
                 super.operationComplete(cf);
               }
+
+              @Override
+              protected boolean abortOnThrowable(Channel channel, Throwable cause) {
+                // FIXME dirty hack until netty issue is not resolved, see https://github.com/netty/netty/issues/6706
+                return StackTraceInspector.recoverOnChunkedUploadFailed(cause)
```
slandelle: Actually, do we really need stacktrace inspection? 
We already know that we're uploading a body. Is there really a way to get a `ChannelOutputShutdownException` whose stacktrace would pass though `ChunkedWriteHandler#flush` (note that there's still some work to do with FileRegion)?

Then, my concern is that I'm not sure wether a `ChannelOutputShutdownException` always means that we'll still be able to read nor if there's anything to read at all. I'm afraid if we don't have a response to read, we'll hang in half closed state forever.

client/src/test/java/org/asynchttpclient/test/EchoHandler.java
```diff
@@ -69,6 +69,13 @@ public void handle(String pathInContext, Request request, HttpServletRequest htt
         httpResponse.sendRedirect(httpRequest.getHeader("X-redirect"));
         return;
       }
+      if (headerName.startsWith("X-fail")) {
+        httpResponse.setStatus(HttpServletResponse.SC_EXPECTATION_FAILED);
+        httpResponse.getOutputStream().write("custom error message".getBytes());
```
slandelle: This response is invalid: as you set a body (and we want one as we want to test this use case), you need to set the Content-Length.

slandelle: `String#getBytes` is something horrible that should be dropped from Java. Please pass the Charset, eg `getBytes(StandardCharsets.US_ASCII/UTF_8)`

client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
beargummy: Yes, you're right, there is no need to check stack trace in this case.
Regarding FileRegion - it is not affected by this issue, since ChunkedWriteHandler handles it in totally different manner.

Issue #1565, opened 2018-07-30
CookieStore is designed incorrectly


tranchitam:
Hi, 

I am using AsyncHttpClient version 2.5.2 for my application. I notice that CookieStore will add the cookie when HttpResponse come https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java#L73-L82 and use cookie from CookieStore for redirect https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java#L129-L136 and execute new request https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java#L193-L207. 

When we send multiple requests with the same URI:
1. request 1 send to server
2. request 2 send to server
3. server returns response 2 with redirect cookie 2
4. server returns response 1 with redirect cookie 1

It's possible for the redirect request 2 will be sent with cookie 1 -> it's incorrect. Also, I don't want to add cookie to the request automatically when it's in cookie store. I can disable CookieStore, but doing that will make redirect won't work. 

I found this PR https://github.com/AsyncHttpClient/async-http-client/pull/1495 added cookie store. I hope we can revert it and don't use cookie store.

@slandelle @unoexperto Can you help to look at this issue?

slandelle:
Please explain the functionality you're trying to achieve.
Are you trying to crawl a website will multiple concurrent virtual users.

unoexperto:
@tranchitam I'm not sure I understand the requirement. You want http client to distinguish between different calls to same domain or do you want client to initialize cookie store only once ?

tranchitam:
@unoexperto I want http client to distinguish between different calls, the next request shouldn't add cookie from the previous request.

unoexperto:
@tranchitam But do you need to remember cookie from previous calls ? If not you can clear cookie store before making new call

```
asyncHttpClient.getConfig.getCookieStore.clear()
```

tranchitam:
@unoexperto it's not safe to use `asyncHttpClient.getConfig.getCookieStore.clear()` before making new call because it's possible after you clear, some http responses (with the same URI) come and cookie will be added again.

unoexperto:
[...]

PR #1566, opened 2018-07-30
Revert cookie store implementation

tranchitam:
Motivations:

- Cookie store works incorrectly when sending multiple requests having the same URI, the later cookie can override the sooner cookie -> not safe when using redirect.
- Don't want to set cookie back when sending new request.

Changes:

- CookieStore will be removed completely.
- Set cookie from response headers to redirect request

Results:

- Set cookie from response headers to redirect request
- Don't set cookie when sending new request.

Related issue:

https://github.com/AsyncHttpClient/async-http-client/issues/1565


slandelle:
No, you can't just remove a feature for the sole reason it doesn't perfectly fit your use case as is.

PR #1567, opened 2018-07-31
Feature/enable cookie store at request level

tranchitam:
**Motivations:**

- Enable the ability to pass CookieStore at request level -> cookies between different requests having the same URI will be independent and will not override each other.
- Still keep global CookieStore from config.

**Changes:**
- `Request` has `getCookieStore()` method.
- `RequestBuilder` has `setCookieStore()` method to set cookie store -> cookie store at request level is `null` by default.
- When http responses come -> add cookies to request cookie store if it's present, otherwise add cookies to global cookie store.
- When setting cookies for redirect request -> if request cookie store is present, get cookies from request cookie store and set them to redirect request. Otherwise, if global cookie store is present, get cookies from global cookie store and and them to redirect request.

**Results:**

- Still keep global cookie store.
- If request has its own cookie store -> cookies for different requests having the same URI will be independent and not override each other

**Related issue:**

#1565


client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
```diff
@@ -71,12 +71,19 @@ public boolean exitAfterIntercept(Channel channel,
 
     // This MUST BE called before Redirect30xInterceptor because latter assumes cookie store is already updated
     CookieStore cookieStore = config.getCookieStore();
```
slandelle: CookieStore cookieStore = request.getCookieStore() != null ? request.getCookieStore() : config.getCookieStore() ;

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
```diff
@@ -71,12 +71,19 @@ public boolean exitAfterIntercept(Channel channel,
...
-    if (cookieStore != null) {
+    CookieStore requestCookieStore = request.getCookieStore();
+    if (cookieStore != null || requestCookieStore != null) {
```
slandelle: Please restore

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
```diff
@@ -71,12 +71,19 @@ public boolean exitAfterIntercept(Channel channel,
...
         if (c != null) {
           // Set-Cookie header could be invalid/malformed
-          cookieStore.add(future.getCurrentRequest().getUri(), c);
+          if (cookieStore != null) {
+            cookieStore.add(future.getCurrentRequest().getUri(), c);
+          }
+
+          if (requestCookieStore != null) {
+            requestCookieStore.add(request.getUri(), c);
+          }
```
slandelle: Please restore

slandelle (changes requested):

slandelle:
@tranchitam Did you get a chance to have a look at my comments. The idea is to be able to override the global CookieStore with one scoped at request level.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
tranchitam: I already updated.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
tranchitam: Restored

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
tranchitam: Restored

tranchitam:
Hi @slandelle, I already updated the pull request to fix your comments. Please take a look. Thanks

mkurz:
@slandelle Can you have a look?

mkurz:
Also there is #1610 now, however seems liks it's just re-using the commits from this pull request here.

leomaiwego:
@slandelle Could you please take a look to this one?

tranchitam:
Hi @slandelle, @TomGranot have you got time to take a look into this pull request again? If there is no issue, we really want to merge this PR to master as soon as possible.

PR #1573, opened 2018-08-31
set useLaxCookieEncoder in DefaultAHCConfig.Builder constructor

SuminAndrew:
It seems that `useLaxCookieEncoder` is not set when creating a `DefaultAHCConfig.Builder` from an instance of `AsyncHttpClientConfig`


slandelle:
Thanks!

PR #1582, opened 2018-10-04
Improve SpnegoEngine to allow more login configuration options

nddipiazza:
# SpnegoEngine improvements

## Summary 

The current spnego implementation locks you into using a single kerberos login configuration, and the SpnegoEngine has no unit test coverage.

This fixes those things.

* Add a `SpnegoEngineTest` that uses apache kerby to test `SpnegoEngine`.
* Allow different JAAS login configurations as parameters to `SpnegoEngine`.
* Properly obtain SPN and don't hard code `HTTP@` spns.
* We should not need to specify a username/password in the `new Realm.Builder()`. It's optional for spnego, so we should not be asserting that it's not null.

## New parameters to the `SpnegoEngine` (parameters added to Realm/DSL where needed)
`servicePrincipalName` - new param* - to allow user to specify their own SPN.
`useCanonicalHostname` -new param* -  whether or not to obtain the canonical hostname or just use the raw host.
`customLoginConfig` - new param* - allow to specify your own login configuration as map<string, string>
`loginContextName` - new param* - allow to choose a specific login configuration from the login conf
`realmName` - send existing realm name to spnego engine to allow users to customize the `@realmName` appended to the SPN.
`username` and `password` - send existing principal and password to the spnego engine to allow username/password kerberos auth as well.

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -73,5 +73,11 @@
       <artifactId>reactive-streams-examples</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.kerby</groupId>
+      <artifactId>kerb-simplekdc</artifactId>
+      <version>${cxf.kerby.version}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -76,7 +76,6 @@
     <dependency>
       <groupId>org.apache.kerby</groupId>
       <artifactId>kerb-simplekdc</artifactId>
-      <version>${cxf.kerby.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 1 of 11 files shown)


nddipiazza:
@slandelle just wondering if you have a moment to check this out.

slandelle:
@nddipiazza Do you have any idea how this could be tested?

nddipiazza:
Hi @slandelle. The Apache CXF project has an example of a JUnit Kerberos ticket server that we could pull into the unit testing code. We could pull that in. it'd be a lot of work. otherwise docker can do it.

nddipiazza:
perhaps https://faustxvi.github.io/junit5-docker/ ?

nddipiazza:
Oh even better, I noticed this test module in apache cxf: https://github.com/apache/cxf/tree/master/systests/kerberos

It uses apache directory to create a real KDC and uses it accordingly. that would be very nice because no docker dependency

slandelle:
Hi @nddipiazza 
Sorry for the lag, I'm pretty swamped atm.
Yes, an embedded server would be awesome! Sorry that I can't help much, I'm not very familiar with Spnego... Do you think you could contribute a test suite based on a fork of CXF work?

nddipiazza:
@slandelle ready for review

slandelle:
Thanks a lot, looks great!

nddipiazza:
Thanks!

Issue #1568, opened 2018-07-31
retrofit2 extra > NullPointerException on handling "204 No-Content" http response


anton-fedorov:
Hello,

It seems that [AsyncHttpClientCall.java:255](https://github.com/AsyncHttpClient/async-http-client/blob/master/extras/retrofit2/src/main/java/org/asynchttpclient/extras/retrofit/AsyncHttpClientCall.java#L255) creates response body to map to OkHttp only if there was actually response in AsyncHttp call.

This unfortunately leads to NullPointerException if you try to parse 204 HTTP Response (No Content).

This happens due to the fact that Retrofit's [OkHttpCall.java#L202](https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/OkHttpCall.java#L202) tries access `rawBody` w/o any null checks.

Would it make sense to have `okHttpBody = ResponseBody.create(contentType, new byte[0])` in case if response was 204/205?

slandelle:
Probably. Contributions welcome :)

danesavot:
I am experiencing the same issue. Any workaround for a quick fix?

slandelle:
What about contributing like I suggested? This module is purely a contribution and I don't use Retrofit myself.

danesavot:
I will try to raise the PR.

slandelle:
I'm not familiar with retrofit and okhttp, but it really looks a retrofit bug to me.
I don't get why we should set a `okhttp3.ResponseBody` when response doesn't have a body, all the more as the ContentType constructor parameter is required, and we obviously don't have one here.

Please report to the retrofit project.

zeldigas:
@slandelle just a quick question - would you accept PR for async-http-client module or the only acceptable way for you would be fix of retrofit itself?

slandelle:
@zeldigas I'd like to have retrofit authors' opinion about this issue. Has anyone reached them and/or opened an issue there (I won't)?

zeldigas:
got it. let me check their javadocs in more details first. I agree that it's counter intuitive to have a body object when there is no body at all, but probably it's by design and we should adjust AHC adapter for their lib...

zeldigas:
Heres docs for `okhttp3.Response#body()` :

> Returns a non-null value if this response was passed to Callback.onResponse or returned from Call.execute(). Response bodies must be closed and may be consumed only once.
This always returns null on responses returned from cacheResponse, networkResponse, and priorResponse().

[...]

PR #1585, opened 2018-10-12
Add non-null ResponseBody for responses with empty body

OKoneva:
PR after discussion [NullPointer while processing response with no body using retrofit](https://github.com/AsyncHttpClient/async-http-client/issues/1568)


extras/retrofit2/src/main/java/org/asynchttpclient/extras/retrofit/AsyncHttpClientCall.java
```diff
@@ -254,6 +254,8 @@ private Response toOkhttpResponse(org.asynchttpclient.Response asyncHttpClientRe
               ? null : MediaType.parse(asyncHttpClientResponse.getContentType());
       val okHttpBody = ResponseBody.create(contentType, asyncHttpClientResponse.getResponseBodyAsBytes());
       rspBuilder.body(okHttpBody);
+    } else {
+      rspBuilder.body(ResponseBody.create(null, ""));
```
slandelle: Please make ResponseBody.create(null, "") a constant

slandelle:
Thanks!

Issue #1527, opened 2018-02-25
Enhancment Request - block while waiting for connection


maltalex:
At the moment, the client throws an exception if there are no free channels when a new request arrives. That makes sense in many scenarios, but in some cases, for instance when working with a pool of connections against some external API, it would be nice to have an option to put such requests in a queue or even block the requesting thread until a connection is available.

slandelle:
AHC is async at its core, there's no way it will ever have a "blocking" feature.

I personally have zero usage for this, so I'll never implement such queueing system myself.

Contributions welcome

PR #1586, opened 2018-10-16
 Added BlockingConnectionSemaphoreFactory 

maltalex:
In response to #1527 here's a simple implementation of a blocking `ConnectionSemaphoreFactory` using plain old Java semaphores.
In some cases where a limit on the number of connections is needed, it's preferable to wait for a free channel over throwing an exception. In such cases this factory can be set in the configuration.

I also think it might be a good idea to add a simpler blocking/non-blocking boolean flag in the configuration for this behavior. It would be simpler for users to find as opposed to `setConnectionSemaphoreFactory(new BlockingConnectionSemaphoreFactory()))`.


slandelle:
Actually, I've been concerned with such strategy for a long time.
Depending on how you are using AHC, you might be trying to send requests from the Netty IO threads: automatic retry on premature socket close, automatic redirect following, people sending requests form an AsyncHandler, etc.
Because of Netty threading model (sockets are assigned to a given thread), blocking such thread would block processing for all other sockets assigned on the same thread. I've seen people running into deadlocks this way.

At the very least, could you use tryAcquire(timeout) instead of infinite acquire?
Also, could you add some tests please?

maltalex:
That's not how I'm using AHC. But sure, it's a good idea to use `tryAquire` with sane defaults or configurable values instead of `aquire`, I'll make the change and add some documentation that says that mixing the blocking semaphore factory and sending requests from an `AsyncHander` are a bad combination. I can also add some tests.

Regarding redirects and similar functionality that's implemented within AHC itself - could you clarify the issue? From what I can tell, channel locks are released once the whole request completes (whether normally or abnormally). Requests following a redirect are sent by the netty thread while the the channel lock is still being held.

maltalex:
@slandelle I've made some changes. The blocking semaphores are now a configuration option. Channel acquisition goes through a `tryAcquire()` with a timeout. The default is zero.

PR #1587, opened 2018-10-19
Allow setting headers using a Map with a subclass of CharSequence as the key type

breun:
Currently this isn't supported:

```
Map<String, List<String>> headers = new HashMap<>();
headers.put("X-Forwarded-For", Colllections.singletonList("10.0.0.1"));

RequestBuilder requestBuilder = new RequestBuilder();
requestBuilder.setHeaders(headers);
```

My IDE reports: `Cannot resolve method 'setHeaders(java.util.Map<java.lang.String,java.util.List<java.lang.String>>)'`

This pull requests adds support for the above construct.


slandelle:
Thanks

PR #1588, opened 2018-10-19
Fix SpnegoEngine.getCompleteServicePrincipalName when servicePrincipalName is not specified.

nddipiazza:
Small fix related to previous PR https://github.com/AsyncHttpClient/async-http-client/pull/1582/files

By making my previous fix I broke scenario when principal name is not specified.

Added a simple unit test to prove this works better.


nddipiazza:
@slandelle did some testing on some real systems with my latest changes, found this one issue i caused.

slandelle:
Thanks!

Issue #857, opened 2015-04-15
Support multipart file part stream feeding


pablolagreca:
Currently AHC support sending a multipart file part but reading from File or a byte array. It should be possible to also send a part which content is read from an InputStream. 

For instance, I may be reading content from the network which I want to send as a multipart file part but I can't read the whole content into memory since te content is big.

slandelle:
@pablolagreca Will to contribute?

wsargent:
It's early, but if you look at the WS docs involving iteratees, there may
be something there

https://www.playframework.com/documentation/2.3.x/ScalaWS

On Thursday, April 16, 2015, Stephane Landelle <notifications@github.com>
wrote:

> @pablolagreca <https://github.com/pablolagreca> Will to contribute?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/AsyncHttpClient/async-http-client/issues/857#issuecomment-93718223>
> .
>


-- 
Will Sargent
Consultant, Professional Services
Typesafe <http://typesafe.com>, the company behind Play Framework
<http://www.playframework.com>, Akka <http://akka.io> and Scala
<http://www.scala-lang.org/>

wsargent:
Multiparty form handling is something that WS itself is weak on, so
contributions welcome :-)

On Thursday, April 16, 2015, Stephane Landelle <notifications@github.com>
wrote:

> @pablolagreca <https://github.com/pablolagreca> Will to contribute?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/AsyncHttpClient/async-http-client/issues/857#issuecomment-93718223>
> .
>


-- 
Will Sargent
Consultant, Professional Services
Typesafe <http://typesafe.com>, the company behind Play Framework
<http://www.playframework.com>, Akka <http://akka.io> and Scala
<http://www.scala-lang.org/>

slandelle:
Note that InputStream is a blocking API, so you might end up shooting yourself in the foot.
If the `InputStream` blocks, you'll block a Netty IO thread, meaning that not only the current channel/socket will be stalled, but also all the other ones associated with this thread.

slandelle:
Should be quite easy to implement since #1018 refactoring.

Anyone interested, please give it a try (I won't myself).

slandelle:
@pablolagreca Actually, what you want is not a `java.io.InputStream` which is a blocking crap.
[...]

Issue #1485, opened 2017-12-10
FileMultipartPart is never "Done" with empty file or channel close exception


yangye1947:
Hi, I'm using ahc 2.0.31.  Observed hanging AHC thread in two scenarios: 1. upload an empty file. 2. netty channel closed. At first, I thought these two cases were unrelated. But going through the code, found they seem caused by the same reason.

I used the default unpooled buffer allocator with UnpooledUnsafeDireByteBuf. In the code, on channel closed or file is empty, it will return -1, which makes sense.
```
public int setBytes(int index, ScatteringByteChannel in, int length) throws IOException {
        this.ensureAccessible();
        ByteBuffer tmpBuf = this.internalNioBuffer();
        tmpBuf.clear().position(index).limit(index + length);

        try {
            return in.read(tmpBuf);
        } catch (ClosedChannelException var6) {
            return -1;
        }
    }
```

But when file multipart part received -1, it decreased the position by -1. Then it compared the position with the length to see if the position reaches the end. With decrease by -1 every time, the position will never reach the end. And causing the thread looping forever...
```
    protected long transferContentTo(ByteBuf target) throws IOException {
        int transferred = target.writeBytes(this.channel, target.writableBytes());
        this.position += (long)transferred;
        if (this.position == this.length) {
            this.state = MultipartState.POST_CONTENT;
            this.channel.close();
        }

        return (long)transferred;
    }
```

This is easy to reproduce. You can simply upload with empty file. Let me know if I mis-understand anything. Thx!

slandelle:
I've added a test in `MultipartUploadTest` sending an empty file and wasn't able to reproduce.

```java
@Test(groups = "standalone")
public void sendEmptyFile() throws Exception {
    File file = getClasspathFile("empty.txt");
    try (AsyncHttpClient c = asyncHttpClient(config())) {
        Request r = post("http://localhost" + ":" + port1 + "/upload/bob")
                .addBodyPart(new FilePart("file", file, "text/plain", UTF_8))
                .build();

        Response res = c.executeRequest(r).get();
        assertEquals(res.getStatusCode(), 200);
    }
}
```

So yes, it would be great if you could provide a unit test to reproduce, please.

slandelle:
> I've added a test in MultipartUploadTest sending an empty file and wasn't able to reproduce.

OK, got it, the code you mention is only used on the non-zero-copy path.
Fixing

yangye1947:
[...]

PR #1593, opened 2018-10-26
#857 Support multipart files using InputStream from source file

samridh90:
**Motivation:**

- In certain situations it's useful for the user to be able to provide an `InputStream` to a file instead of the `File` object itself for a multipart file part.
- E.g.: Upload large files accessible over a network protocol such as SFTP to a remote server without copying the file to the local filesystem or loading it fully into memory

**Changes:**

- Introduce new `InputStreamPart`type for use in the `addBodyPart(...)` API
- Implement `InputStreamMultipartPart` to transfer `InputStream `to a Netty `ByteBuf` or `WritableByteChannel`
- _Unfortunately did not find a way to implement `protected long transferContentTo(WritableByteChannel target) throws IOException` for zero-copy uploads._
- Added new tests and updated existing tests to cover scenarios similar to `FilePart`

**Results:**

- It is now possible to provide an `InputStream` as a multipart body part

**Related Issue:**

[#857](https://github.com/AsyncHttpClient/async-http-client/issues/857)

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,8 +113,6 @@ This part can be of type:
 * `InputStreamPart`
 * `StringPart`
 
-**NOTE**: `InputStreamPart` does not support zero-copy transfers. Explicitly disable zero-copy using `config.setDisableZeroCopy(true)` when one of the body parts is an `InputStreamPart`.
-
 ### Dealing with Responses
 
 #### Blocking on the Future
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ Use the `addBodyPart` method to add a multipart part to the request.
 This part can be of type:
 * `ByteArrayPart`
 * `FilePart`
+* `InputStreamPart`
 * `StringPart`
 
 ### Dealing with Responses
--- a/README.md
+++ b/README.md
@@ -113,6 +113,8 @@ This part can be of type:
 * `InputStreamPart`
 * `StringPart`
 
+**NOTE**: `InputStreamPart` does not support zero-copy transfers. Explicitly disable zero-copy using `config.setDisableZeroCopy(true)` when one of the body parts is an `InputStreamPart`.
+
 ### Dealing with Responses
 
 #### Blocking on the Future
--- a/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
@@ -53,14 +53,8 @@ public class NettyBodyBody implements NettyBody {
   public void write(final Channel channel, NettyResponseFuture<?> future) {
 
     Object msg;
-    if (body instanceof RandomAccessBody && !ChannelManager.isSslHandlerConfigured(channel.pipeline()) && !config.isDisableZeroCopy()) {
-      long contentLength = getContentLength();
-      if (contentLength < 0) {
-        // contentLength unknown in advance, use chunked input
-        msg = new BodyChunkedInput(body);
-      } else {
-        msg = new BodyFileRegion((RandomAccessBody) body);
-      }
+    if (body instanceof RandomAccessBody && !ChannelManager.isSslHandlerConfigured(channel.pipeline()) && !config.isDisableZeroCopy() && getContentLength() > 0) {
+      msg = new BodyFileRegion((RandomAccessBody) body);
 
     } else {
       msg = new BodyChunkedInput(body);
--- a/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
@@ -54,13 +54,7 @@ public class NettyBodyBody implements NettyBody {
 
     Object msg;
     if (body instanceof RandomAccessBody && !ChannelManager.isSslHandlerConfigured(channel.pipeline()) && !config.isDisableZeroCopy()) {
-      long contentLength = getContentLength();
-      if (contentLength < 0) {
-        // contentLength unknown in advance, use chunked input
-        msg = new BodyChunkedInput(body);
-      } else {
-        msg = new BodyFileRegion((RandomAccessBody) body);
-      }
+      msg = new BodyFileRegion((RandomAccessBody) body);
 
     } else {
       msg = new BodyChunkedInput(body);
--- a/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
@@ -54,7 +54,13 @@ public class NettyBodyBody implements NettyBody {
 
     Object msg;
     if (body instanceof RandomAccessBody && !ChannelManager.isSslHandlerConfigured(channel.pipeline()) && !config.isDisableZeroCopy()) {
-      msg = new BodyFileRegion((RandomAccessBody) body);
+      long contentLength = getContentLength();
+      if (contentLength < 0) {
+        // contentLength unknown in advance, use chunked input
+        msg = new BodyChunkedInput(body);
+      } else {
+        msg = new BodyFileRegion((RandomAccessBody) body);
+      }
 
     } else {
       msg = new BodyChunkedInput(body);
--- a/client/src/main/java/org/asynchttpclient/request/body/multipart/InputStreamPart.java
+++ b/client/src/main/java/org/asynchttpclient/request/body/multipart/InputStreamPart.java
@@ -1,3 +1,16 @@
+/*
+ * Copyright (c) 2018 AsyncHttpClient Project. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at
+ *     http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
 package org.asynchttpclient.request.body.multipart;
 
 import java.io.InputStream;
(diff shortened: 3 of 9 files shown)


zbum:
👍

client/src/main/java/org/asynchttpclient/request/body/multipart/InputStreamPart.java
```diff
@@ -0,0 +1,49 @@
...
+import java.nio.charset.Charset;
+
+import static org.asynchttpclient.util.Assertions.assertNotNull;
+
+public class InputStreamPart extends FileLikePart {
+
+  private final InputStream inputStream;
+  private final long contentLength;
+
+  public InputStreamPart(String name, InputStream inputStream, long contentLength, String fileName) {
```
slandelle: small nit: could you please add a constructir without contenLength that sets the value as `-1`? There's a chance most users won't know the stream length beforehand.

client/src/test/java/org/asynchttpclient/request/body/multipart/MultipartUploadTest.java
```diff
@@ -142,6 +157,28 @@ public void sendEmptyFileZeroCopy() throws Exception {
     sendEmptyFile0(false);
   }
 
+  private void sendEmptyFileInputStream0(boolean disableZeroCopy) throws Exception {
+    File file = getClasspathFile("empty.txt");
```
slandelle: Could you please use a non empty file?

client/src/test/java/org/asynchttpclient/request/body/multipart/MultipartUploadTest.java
```diff
@@ -142,6 +157,28 @@ public void sendEmptyFileZeroCopy() throws Exception {
...
+    try (AsyncHttpClient c = asyncHttpClient(config().setDisableZeroCopy(disableZeroCopy))) {
+      InputStream inputStream = new BufferedInputStream(new FileInputStream(file));
+      Request r = post("http://localhost" + ":" + port1 + "/upload")
+              .addBodyPart(new InputStreamPart("file", inputStream, file.length(), file.getName(), "text/plain", UTF_8)).build();
```
slandelle: Could you please test with and without Content-Length prior knowledge?

client/src/test/java/org/asynchttpclient/request/body/multipart/MultipartUploadTest.java
samridh90: This helper method is explicitly to test empty file to ensure we're covering the case from issue https://github.com/AsyncHttpClient/async-http-client/issues/1485. I can add an additional test for non-empty file

slandelle: Yes please.

samridh90: Done.

samridh90:
I ran into some trouble with the Netty `HttpObjectEncoder` class when I tried to get zero-copy with -1 contentLength to work. Here's what happens:

- In `org/asynchttpclient/netty/request/body/NettyBodyBody.java#write(final Channel channel, NettyResponseFuture<?> future)` we set `msg = new BodyFileRegion((RandomAccessBody) body);`
- In `io/netty/netty-codec-http/4.1.30.Final/netty-codec-http-4.1.30.Final.jar!/io/netty/handler/codec/http/HttpObjectEncoder.class#encode(ChannelHandlerContext ctx, Object msg, List<Object> out)` this cause us to hit the following block
```java
            case 2:
                if (buf != null) {
                    out.add(buf);
                }

                this.encodeChunkedContent(ctx, msg, contentLength(msg), out);
                break;
```
- Taking a look at the `contentLength(msg)` method, this return -1 when `msg` matches `FileRegion`
```java
private static long contentLength(Object msg) {
        if (msg instanceof HttpContent) {
            return (long)((HttpContent)msg).content().readableBytes();
        } else if (msg instanceof ByteBuf) {
            return (long)((ByteBuf)msg).readableBytes();
        } else if (msg instanceof FileRegion) {
            return ((FileRegion)msg).count();
        } else {
            throw new IllegalStateException("unexpected message type: " + StringUtil.simpleClassName(msg));
        }
    }
```
- In `encodeChunkedContent(ChannelHandlerContext ctx, Object msg, long contentLength, List<Object> out)`:
```java
private void encodeChunkedContent(ChannelHandlerContext ctx, Object msg, long contentLength, List<Object> out) {
        ByteBuf buf;
        if (contentLength > 0L) {
            String lengthHex = Long.toHexString(contentLength);
            buf = ctx.alloc().buffer(lengthHex.length() + 2);
            buf.writeCharSequence(lengthHex, CharsetUtil.US_ASCII);
            ByteBufUtil.writeShortBE(buf, 3338);
            out.add(buf);
            out.add(encodeAndRetain(msg));
            out.add(CRLF_BUF.duplicate());
        }

        if (msg instanceof LastHttpContent) {
            HttpHeaders headers = ((LastHttpContent)msg).trailingHeaders();
            if (headers.isEmpty()) {
                out.add(ZERO_CRLF_CRLF_BUF.duplicate());
            } else {
                buf = ctx.alloc().buffer((int)this.trailersEncodedSizeAccumulator);
                ByteBufUtil.writeMediumBE(buf, 3149066);
                this.encodeHeaders(headers, buf);
                ByteBufUtil.writeShortBE(buf, 3338);
                this.trailersEncodedSizeAccumulator = 0.2F * (float)padSizeForAccumulation(buf.readableBytes()) + 0.8F * this.trailersEncodedSizeAccumulator;
                out.add(buf);
            }
        } else if (contentLength == 0L) {
            out.add(encodeAndRetain(msg));
        }
    }
```
- We basically miss all the blocks since the `encodeChunkedContent` method doesn't handle `contentLength < 0L`.

@slandelle : Not sure if my solution is acceptable, I'd be open to better ideas.

slandelle:
samridh90 I don't think it's worth trying to contribute chunked transfert encoding support for Netty's FileRegion. We can just disable [bypass zero-copy](https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java#L56) when content-length < 0.
InputStream isn't zero-copy anyway, such it would only be suboptimal when sending a multipart that mixes a local file and an InputStream.
Then, don't bother implementing zero-copy for InputStreamPart as it shouldn't be used.

WDYT?

samridh90:
@slandelle : Sounds good to me. I'll get rid of the zero-copy for InputStreamPart.

client/src/main/java/org/asynchttpclient/netty/request/body/NettyBodyBody.java
```diff
@@ -54,13 +54,7 @@ public void write(final Channel channel, NettyResponseFuture<?> future) {
 
     Object msg;
     if (body instanceof RandomAccessBody && !ChannelManager.isSslHandlerConfigured(channel.pipeline()) && !config.isDisableZeroCopy()) {
```
slandelle: There was some misunderstanding. In my previous comment, I pointed to this line. If you check the content-length here, you can force to the BodyChunkedInput branch if it's < 0. Hence, There's no need to advice the user to disable zero-copy globally, it can be completely transparent.

README.md
```diff
@@ -110,8 +110,11 @@ Use the `addBodyPart` method to add a multipart part to the request.
 This part can be of type:
 * `ByteArrayPart`
 * `FilePart`
+* `InputStreamPart`
 * `StringPart`
 
+**NOTE**: `InputStreamPart` does not support zero-copy transfers. Explicitly disable zero-copy using `config.setDisableZeroCopy(true)` when one of the body parts is an `InputStreamPart`.
```
slandelle: Please revert

samridh90: Thanks for clearing that up.
I've reverted the changes and added a check to BodyChunkedInput if contentLength  0

client/src/main/java/org/asynchttpclient/request/body/multipart/InputStreamPart.java
```diff
@@ -0,0 +1,53 @@
+package org.asynchttpclient.request.body.multipart;
```
slandelle: Could you please add the license file header:

```
/*
 * Copyright (c) 2018 AsyncHttpClient Project. All rights reserved.
 *
 * This program is licensed to you under the Apache License Version 2.0,
 * and you may not use this file except in compliance with the Apache License Version 2.0.
 * You may obtain a copy of the Apache License Version 2.0 at
 *     http://www.apache.org/licenses/LICENSE-2.0.
 *
 * Unless required by applicable law or agreed to in writing,
 * software distributed under the Apache License Version 2.0 is distributed on an
 * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
 */
```

client/src/main/java/org/asynchttpclient/request/body/multipart/part/InputStreamMultipartPart.java
```diff
@@ -0,0 +1,92 @@
+package org.asynchttpclient.request.body.multipart.part;
```
slandelle: missing file header

client/src/test/java/org/asynchttpclient/request/body/InputStreamPartLargeFileTest.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
```
slandelle: Please fix header: Sonatype is no longer involved in this project and new code should be donated to AHC.

slandelle:
On last nit and we'll be good :)

Thanks!

client/src/test/java/org/asynchttpclient/request/body/InputStreamPartLargeFileTest.java
samridh90: Fixed

client/src/main/java/org/asynchttpclient/request/body/multipart/InputStreamPart.java
samridh90: Added, thanks for pointing that out.

slandelle:
Merged. Great job, thanks!

slandelle:
2.6.0 should be available on maven central in 15 min.

PR #1596, opened 2018-11-13
Expose SSL Session of a connection to AsyncHandler.onTlsHandshakeSuccess

rolandomanrique:
The goal of this change is to expose underlying SSLSession object once handshake is successful. This seems like a feature that can be useful to anyone that needs to retrieve session and/or certificate information from the connection.

A quick search for the feature yielded this conversation in google groups from 2014 so it seems people have been trying to do this for a while:

https://groups.google.com/forum/#!msg/asynchttpclient/7ySVbKqntgc/_pycVfrwiVMJ

### DIFF
--- a/client/src/main/java/org/asynchttpclient/AsyncHandler.java
+++ b/client/src/main/java/org/asynchttpclient/AsyncHandler.java
@@ -19,6 +19,7 @@ import io.netty.channel.Channel;
 import io.netty.handler.codec.http.HttpHeaders;
 import org.asynchttpclient.netty.request.NettyRequest;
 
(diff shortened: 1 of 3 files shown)


rolandomanrique:
@slandelle any comments on this PR? Thanks.

slandelle:
Looks fine but will have to release a minor version as it's a minor breaking change.

rolandomanrique:
@slandelle would you like me to set the version in this PR? If so, should I go with `2.6.2-SNAPSHOT` or `2.7.0-SNAPSHOT`?

PR #1600, opened 2018-12-13
Update README.md

gsilvestrin:
Fixed artifact name to `async-http-client-extras-typesafe-config`, there was a missing dash


slandelle:
Thanks!

PR #1601, opened 2018-12-14
Fix bug in Url encoding of QueryParams

maesenka:
This fixes the bug that unreserved characters (see RFC-3986) are also percent-encoded.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/util/Utf8UrlEncoder.java
+++ b/client/src/main/java/org/asynchttpclient/util/Utf8UrlEncoder.java
@@ -135,7 +135,7 @@ public final class Utf8UrlEncoder {
   }
 
   public static StringBuilder encodeAndAppendQueryElement(StringBuilder sb, CharSequence input) {
-    return appendEncoded(sb, input, FORM_URL_ENCODED_SAFE_CHARS, false);
+    return encodeAndAppendPercentEncoded(sb, input);
   }
 
   public static StringBuilder encodeAndAppendFormElement(StringBuilder sb, CharSequence input) {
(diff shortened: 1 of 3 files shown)


slandelle:
RFC3986 only specifies valid chars inside a query.
From [RFC3986](https://www.ietf.org/rfc/rfc3986.txt) chapter 3.4:

> query       = *( pchar / "/" / "?" )

Then HTML spec specify how to encode (key, value) pairs from a form. This encoding is the most widely used query encoding.
From [HTML5 spec:](https://www.w3.org/TR/html52/sec-forms.html#mutate-action-url):

> Let query be the result of encoding the form data set using the application/x-www-form-urlencoded encoding algorithm, interpreted as a US-ASCII string.

The method you're suggesting to change, `encodeAndAppendQueryElement`, is used for encoding such (key, value) pairs, so it should go with the HTML spec, even though some characters might end up being encoded while they would be legit as per RFC3986.
You can find a summary of the algorithm on [Wikipedia](https://en.wikipedia.org/wiki/Query_string#URL_encoding).

maesenka:
A little context: I wrote an [AWS Signer](https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html) for AHC (see https://github.com/WegenenVerkeer/RxHttpClient). 

All tests in Amazon's test suite worked when using AHC 1.x, but one test failed when I used AHC 2.6.  The failed test involves the following request (note the '~') : `GET /?-._~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz=-._~0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz http/1.1`

Debugging let me to the method I proposed to change.

I understand your argument for rejecting this change. But is there some way then to distinguish the context of encoding form data and encoding the query-string?

slandelle:
I suspect you want to use `Utf8UrlEncoder#percentEncodeQueryElement`, or use your own encoder.

Issue #1091, opened 2016-02-04
OSGi bundle resolving problems


thjaeckle:
Hi. 
With the latest 2.0.0-RC8 there seem to be some problems with the OSGi manifests.
When trowing following bundles into into a Felix container I get 2 errors.

#### The Bundles
 (which are transitive dependencies of async-http-client):

* netty-transport-4.0.34.Final.jar
* netty-handler-4.0.34.Final.jar
* netty-common-4.0.34.Final.jar
* netty-codec-4.0.34.Final.jar
* netty-codec-http-4.0.34.Final.jar
* netty-buffer-4.0.34.Final.jar
* javassist-3.18.1-GA.jar
* netty-reactive-streams-1.0.2.jar

* async-http-client-2.0.0-RC8.jar
* netty-codec-dns-2.0.0-RC8.jar
* netty-resolver-dns-2.0.0-RC8.jar
* netty-resolver-2.0.0-RC8.jar

#### Error 1
```
2016-02-04 12:51:46,425 | WARN  | ? | org.apache.felix.fileinstall | Error while starting bundle: file:/D:/felix/osgi/deploy/netty-resolver-dns-2.0.0-RC8.jar
org.osgi.framework.BundleException: Uses constraint violation. Unable to resolve bundle revision org.asynchttpclient.netty-resolver-dns [105.0] because it exports package 'io.netty.channel' and is also exposed to it from bundle revision io.netty.transport [95.0] via the following dependency chain:

  org.asynchttpclient.netty-resolver-dns [105.0]
    import: (&(osgi.wiring.package=io.netty.handler.codec.dns)(version>=2.0.0)(!(version>=3.0.0)))
     |
    export: osgi.wiring.package=io.netty.handler.codec.dns; uses:=io.netty.channel.socket
  org.asynchttpclient.netty-codec-dns [102.0]
    import: (&(osgi.wiring.package=io.netty.channel.socket)(version>=4.0.0)(!(version>=5.0.0)))
     |
    export: osgi.wiring.package=io.netty.channel.socket; uses:=io.netty.channel
    export: osgi.wiring.package=io.netty.channel
```

#### Error 2
```
2016-02-04 12:51:46,430 | WARN  | ? | org.apache.felix.fileinstall | Error while starting bundle: file:/D:/felix/osgi/deploy/async-http-client-2.0.0-RC8.jar
[...]

PR #1605, opened 2019-01-14
reenable osgi support

jenskordowski:
The OSGi support was removed due to https://github.com/AsyncHttpClient/async-http-client/issues/1091

This is a revert of https://github.com/AsyncHttpClient/async-http-client/commit/ed92e2a1932d8709b28d892606e57f7fe60bf820

I tested with a current karaf distribution and the bundle wires providing it the necessary dependencies (netty et. al)

I assume the package uses constraint violation does not exist anymore.

Best regards
Jens

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,6 +151,28 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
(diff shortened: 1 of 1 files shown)


slandelle:
Thanks

PR #1618, opened 2019-02-28
Improved retrofit timeout handling.

bfg:
Retrofit `Call` now uses http client's configured timeout to compute
value for `timeout()` and blocking `execute()`  method execution timeout.

### DIFF
--- a/extras/retrofit2/src/main/java/org/asynchttpclient/extras/retrofit/AsyncHttpClientCall.java
+++ b/extras/retrofit2/src/main/java/org/asynchttpclient/extras/retrofit/AsyncHttpClientCall.java
@@ -40,14 +40,6 @@ import java.util.function.Supplier;
 @Builder(toBuilder = true)
 @Slf4j
 class AsyncHttpClientCall implements Cloneable, okhttp3.Call {
-  /**
-   * Default {@link #execute()} timeout in milliseconds (value: <b>{@value}</b>)
-   *
-   * @see #execute()
-   * @see #executeTimeoutMillis
-   */
-  public static final long DEFAULT_EXECUTE_TIMEOUT_MILLIS = 30_000;
-
   private static final ResponseBody EMPTY_BODY = ResponseBody.create(null, "");
 
   /**
@@ -63,12 +55,6 @@ class AsyncHttpClientCall implements Cloneable, okhttp3.Call {
    */
   @NonNull
   Supplier<AsyncHttpClient> httpClientSupplier;
-
-  /**
-   * {@link #execute()} response timeout in milliseconds.
-   */
-  @Builder.Default
-  long executeTimeoutMillis = DEFAULT_EXECUTE_TIMEOUT_MILLIS;
 
   /**
    * Retrofit request.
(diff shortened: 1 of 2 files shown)


bfg:
@slandelle, i'm a bit confused about AHC timeouts; as seen from [newly created test](https://github.com/AsyncHttpClient/async-http-client/pull/1618/files#diff-718e3401559905f9663f7e28b1517e23R345) it's possible to set read timeout to a higher value than request timeout. Which of those values are then being used by AHC instance? If request timeout is always enforced i'll change [retrofit timeout computation method](https://github.com/AsyncHttpClient/async-http-client/pull/1618/files#diff-d69183c7af5cfa64930ac287647f8b11R186).

slandelle:
If read timeout is longer than request timeout, it will be ignored.

bfg:
@slandelle i've just pushed c3585b1 with [updated timeout computation method](https://github.com/AsyncHttpClient/async-http-client/pull/1618/commits/c3585b1c063cf1ad77fdef58d2b581ed967fb5ac#diff-d69183c7af5cfa64930ac287647f8b11R186). Does it make sense to you? If yes, i'm going to squash commit, then you're free to merge.

slandelle:
No, that doesn't work. `read timeout` is about stalled response when nothing is received for a long time. Think large file download. You might want a 5s read timeout and a 2 min request timeout.

bfg:
So what would you suggest? `timeout()` retrofit `Call` method is used solely by okhttp native client, so it's irrelevant; But there's a possibility that some retrofit call adapters would call `execute()` in a blocking manner (example: [rxjava call adapter initialized with some scheduler](https://github.com/square/retrofit/tree/master/retrofit-adapters/rxjava2#usage)) where correct timeout is quite essential; I know that this is not super important, because the whole idea of using AHC with retrofit is to use async retrofitx rxjava call adapter, where timeout handling is done by AHC itself, but nevertheless i would like to improve experience also for users that for some reason initialize call adapters with some other scheduler.

In essence, what i would like to achieve in this PR is to retrieve correct timeout from actual configuration of http client that is going to execute this call; this is improvement, because previously, timeout has been semi-hardcoded in a field and was not configurable by the user.

slandelle:
Just map on request timeout.

bfg:
:+1: i've updated the code, squashed commits; now solely request timeout is used; you're free to merge. And yeah, this time i've tested snapshot with my service :-) Releasing new fix version would be awesome, but it's up to you.

slandelle:
@bfg I can't release because your PR breaks javadoc (which is required in order to release on maven central).
Run `mvn install javadoc:javadoc -DskipTests` in order to reproduce.
It looks like an issue with the latest lombok changes you've introduced.

Could you please fix?

bfg:
Sigh, sure, on it :-)

bfg:
@slandelle it builds successfully for me using maven `3.5`/`3.6`; which maven version are you using?

bfg:
Can you check [retrofit_javadoc_issue](https://github.com/bfg/async-http-client/tree/retrofit2_javadoc_issue) branch on my fork?

bfg:
I've managed to reproduce javadoc issue. Build works without issues on openjdk 8, but it fails on oracle jdk 8 :man_facepalming: 

```
Java version: 1.8.0_201, vendor: Oracle Corporation, runtime: /opt/java/jdk1.8.0_201/jre
```

I'll find a way to work around this issue, one way or another :-)

PR #1620, opened 2019-02-28
Added maven wrapper.

bfg:
Using maven wrapper helps at build stability. Changes:

* added maven wrapper `0.5.3` configured to use maven `3.6.0`
* updated travis config and scripts to use maven wrapper


slandelle:
No, I don't want to use maven wrapper. I already have all my local set up and maven config that works fine. And I already use maven 3.6.0.

bfg:
No problem, I'll investigate further until i solve javadoc issue issue.

slandelle:
Lombok will sadly always be a huge hack with brittle support from other tools...

PR #1622, opened 2019-03-07
expose netty channel to the KeepAliveStrategy

wuguangkuo:
expose netty channel to the KeepAliveStrategy, so that we can get current InetAddress used by current request. when the custom DNS cache is updated, we can set current request's keepalive to false, just like what we can do used the apache httpcomponents "ConnectionReuseStrategy".
Below is my pseudocode using the new KeepAliveStrategy：

```java
public boolean keepAlive(Channel channel, Request ahcRequest, HttpRequest request, HttpResponse response) {
        String hostname = ahcRequest.getUri().getHost();
        InetSocketAddress socketAddress = (InetSocketAddress) channel.remoteAddress();
        InetAddress currentAddress = socketAddress.getAddress();
        List<InetAddress> newAddressList = DnsCache.getAddress(hostname);
        if (currentAddress not in newAddressList) {
                //the "DnsCache" has been updated：the remote server is going to off line.
                return false;
        }
        return super.keepAlive(channel, ahcRequest, request, response);
    }
```


wuguangkuo:
@slandelle Hi, could you deal this pull request?

slandelle:
I'm not fond of exposing `Channel`. Could you please update your PR to expose the `InetSocketAddress` instead please?

wuguangkuo:
> I'm not fond of exposing `Channel`. Could you please update your PR to expose the `InetSocketAddress` instead please?

Ok, I encapsulate a class "RealConnection" as a holder of localAddress and remoteAddress.

slandelle:
I don’t get how local address is interesting for deciding on keeping the connection alive. Could you please explain ?

wuguangkuo:
> I don’t get how local address is interesting for deciding on keeping the connection alive. Could you please explain ?

In my case, the remote address is enough. But I think there may be someone need the local address to do something. In fact, I modified the code emulating "HttpInetConnection" in apache httpcomponents.

slandelle:
I really don't see a use case for local address here. Please just expose remote one.

wuguangkuo:
modified

PR #1623, opened 2019-03-12
web socket dose not work in socks5 proxy 

cowwen:
I am in trouble when using web socket in socks5 proxy.

Motivation:
Here are 2 issues when I debug in code
1, when send request in socks5, client dose not send upgrade request because using http bootstrap , the http bootstrap dose not have wsHandler.
2, Can't not use socks5 domain resolve when client need.
I want write unit test for that but need a socks5 test server, it need spend some time. So I pull change first, others can check the code before it.
 
Modifications:
1, Modified ChannelManager getBootstrap logic when using socks proxy. Web socket using in socks proxy will use wsBootstrap for upgrade protocol handshake.
2, Proxy config add a domain resolver for setting bootstrap resolve. Use Netty NoopAddressResolverGroup for domain resolve in socks5 server, if you dose not set for this, it will use JDK default strategy for resolving.
 
Result:
After I mofidied those, my program run success 

Forgive my poor English, it's not my native language...

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
@@ -344,7 +344,7 @@ public final class NettyRequestSender {
       InetSocketAddress unresolvedRemoteAddress = InetSocketAddress.createUnresolved(proxy.getHost(), port);
       scheduleRequestTimeout(future, unresolvedRemoteAddress);
       return RequestHostnameResolver.INSTANCE.resolve(request.getNameResolver(), unresolvedRemoteAddress, asyncHandler);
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -342,7 +344,16 @@ protected void onFailure(Throwable cause) {
       InetSocketAddress unresolvedRemoteAddress = InetSocketAddress.createUnresolved(proxy.getHost(), port);
       scheduleRequestTimeout(future, unresolvedRemoteAddress);
       return RequestHostnameResolver.INSTANCE.resolve(request.getNameResolver(), unresolvedRemoteAddress, asyncHandler);
-
+    } else if (proxy != null && proxy.isResolveDomain() && ProxyType.SOCKS_V5 == proxy.getProxyType()) {
```
slandelle: Could you please explain this change?

PR #1656, opened 2019-08-01
make NettyWebSocket.sendCloseFrame(int statusCode, String reasonText) actually respect its arguments

radai-rosenblatt:
as opposed to hard-coding 1000 and "normal closure"
Signed-off-by: radai-rosenblatt <radai.rosenblatt@gmail.com>


slandelle:
Thanks!

PR #1662, opened 2019-09-03
Add an option to toggle WebSocket mask flag

webfolderio:


slandelle:
Could you please elaborate on why you'd want this?
As [per spec](https://tools.ietf.org/html/rfc6455#section-10.3), masking is a **required** security measure, not an option:

> Clients MUST choose a new masking key for each frame, using an algorithm that cannot be predicted by end applications that provide data.

webfolderio:
I use async-http-client to connect Chrome DevTools protocol. DevTools protocol does not require to use mask. Without mask it performs better. My servers and clients are running on localhost. Because of this i am not highly concern with security.

slandelle:
> Without mask it performs better.

Could you please provide a link to a issue raised against Chrome DevTools explaining the issue and what's the position of the devs there? Look like an issue with Chrome DevTools to me.

> Because of this i am not highly concern with security.

You might not, but others might shoot themselves in the foot. Or be compromised. Or use AHC for malicious reasons.

PR #1664, opened 2019-09-27
Added Automatic-Module-Name

doom369:
Add `Automatic-Module-Name` to the manifest of builded jar and updated the `maven-jar-plugin` version. This is required to avoid warnings during the build of the systems that switched to the Java 9 modules.

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -9,6 +9,10 @@
   <name>Asynchronous Http Client</name>
   <description>The Async Http Client (AHC) classes.</description>
 
+  <properties>
+    <javaModuleName>org.asynchttpclient.client</javaModuleName>
+  </properties>
+
   <build>
     <plugins>
       <plugin>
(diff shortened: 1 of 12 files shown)


slandelle:
I'm not familiar with Java modules, but are you sure this is correct?
I would expect each artifact to have its own module name instead of hardcoding "org.asynchttpclient". How things are done in Netty:

* https://github.com/netty/netty/blob/4.1/pom.xml#L1292
* https://github.com/netty/netty/search?q=javaModuleName&unscoped_q=javaModuleName

doom369:
@slandelle you are right. I missed you have multiple modules.

doom369:
Pushed the fix. I can't check locally it is correct configuration right now (due to another non-related build errors).

slandelle:
Thanks!

doom369:
Thank you as well :)!

PR #1665, opened 2019-09-28
Support Netty kqueue transport

### LINKED ISSUE #1575: KQueueEventLoopGroup is ignored
Here it is
https://github.com/AsyncHttpClient/async-http-client/blob/a6d1cd2cd6cd027d9c183b023caeb77864f28691/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java#L138-L142

I just want to use native EventLoopGroup on MacOS.

astei:
Resolves #1575. I've tested this on macOS 10.14.6 and Ubuntu 18.04 LTS.

I did a bit more reshuffling than was strictly necessary so that it will be easier for AHC to support any native transports that Netty may introduce in the future.

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -57,6 +57,11 @@
     </dependency>
     <dependency>
       <groupId>io.netty</groupId>
+      <artifactId>netty-transport-native-kqueue</artifactId>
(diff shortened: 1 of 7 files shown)


slandelle:
Thanks!

PR #1669, opened 2019-10-15
Support for OpenSslEngine with no finalizer

johnou:
Motivation:

Support for Netty SslProvider.OPENSSL_REFCNT (OpenSSL-based implementation which does not have finalizers and instead implements ReferenceCounted).

Modification:

Add destroy method to SslEngineFactory to allow cleaning up reference counted SslContext.

Result:

Users can opt-in to a finalizer free OpenSslEngine and OpenSslContext.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/SslEngineFactory.java
+++ b/client/src/main/java/org/asynchttpclient/SslEngineFactory.java
@@ -39,4 +39,12 @@ public interface SslEngineFactory {
   default void init(AsyncHttpClientConfig config) throws SSLException {
     // no op
   }
+
+  /**
+   * Perform any necessary cleanup.
+   */
+  default void destroy() {
+    // no op
+  }
+
 }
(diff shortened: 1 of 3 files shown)


johnou:
Currently enabling openssl_refcnt using a custom DefaultSslEngineFactory..

```java
  public static class ReferenceCountedOpenSslEngineFactory extends DefaultSslEngineFactory {

    private SslProvider provider;

    @Override
    public void init(AsyncHttpClientConfig config) throws SSLException {
      this.provider = config.isUseOpenSsl() ? SslProvider.OPENSSL_REFCNT : SslProvider.JDK;
      super.init(config);
    }

    @Override
    protected SslContextBuilder configureSslContextBuilder(SslContextBuilder builder) {
      return builder.sslProvider(provider);
    }
  }
```

johnou:
As a side note I ran the ATC testsuite with `-Dio.netty.leakDetectionLevel=paranoid` which passed without any leak reports.

slandelle:
Thanks!

slandelle:
As a side issue, it would be nice to be able to enable from config instead of having to go with a custom SslEngineFactory.

johnou:
@slandelle I did consider that, however, OPENSSL_REFCNT is currently marked UnstableApi and wanted to avoid dependency issues with Netty.

slandelle:
Do you mean you're not sure OPENSSL_REFCNT is not production ready and you want to experiment first?

johnou:
I believe it is production ready, however it is still marked "unstable" in Netty here https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/SslProvider.java#L37

slandelle:
OK, let's leave as is for now.
Thanks!

Issue #1672, opened 2019-11-06
Client re-building a request prevents usage of custom requests


kringol:
Currently the client is creating a new RequestBuilder(prototype) on some cases like
- When there are cookies on the storage for that uri
- When the range offset is not 0

Problem is we are using some custom filters that rely on us using an extended implementation of Request.java which adds some extra settings which the filter consumes. Normally it works fine, and the filter checks if the request instance is of our type and extracts those. But due to the mentioned behavior we are loosing our custom implementation on the way when there are cookies for example. 

We can contribute with some changes to overcome if you are open to it.
Proposed solution: 
Instead of having a RequestBuilder(prototype) on the client, have something like request.builder(), which will do the same but we can override on our request implementation to use a builder that will take our extra settings into account.

slandelle:
Sure, please do send a PR.

kringol:
> Sure, please do send a PR.

Thanks @slandelle, I've submitted a PR and left some comments for you there as well

PR #1673, opened 2019-11-07
Added toBuilder methods to Request

kringol:
Changes for https://github.com/AsyncHttpClient/async-http-client/issues/1672

@slandelle These are the simple/minimum changes. 

But there is the fact that the constructors of RequestBuilder that take a prototype remain and could be used directly, breaking again the purpose of this change. 
But in order to overcome that a more impactful change (backwards incompatible) has to be made. But I would leave that decision to you or your opinion on it.
- Ideal: moving the RequestBuilder as an inner class of Request or DefaultRequest and restricting access to those methods 
- Not ideal: Making those constructors protected at least to reduce the chances of usage directly, but still poses a risk

Bonus question:
Do you prefer the implementation of toBuilder methods as default implementation on the interface or as it is now on the DefaultRequest.java?

### DIFF
--- a/client/src/main/java/org/asynchttpclient/Request.java
+++ b/client/src/main/java/org/asynchttpclient/Request.java
@@ -184,12 +184,8 @@ public interface Request {
   /**
    * @return a new request builder using this request as a prototype
    */
-  RequestBuilder toBuilder();
-
-  /**
-   * @param disableUrlEncoding whether to disable url encoding or not
-   * @param validateHeaders whether to enable header validation or not
-   * @return a new request builder using this request as a prototype
-   */
-  RequestBuilder toBuilder(boolean disableUrlEncoding, boolean validateHeaders);
+  @SuppressWarnings("deprecation")
+  default RequestBuilder toBuilder() {
+    return new RequestBuilder(this);
+  }
 }
--- a/client/src/main/java/org/asynchttpclient/Request.java
+++ b/client/src/main/java/org/asynchttpclient/Request.java
@@ -180,4 +180,16 @@ public interface Request {
    * @return the NameResolver to be used to resolve hostnams's IP
    */
   NameResolver<InetAddress> getNameResolver();
+
+  /**
+   * @return a new request builder using this request as a prototype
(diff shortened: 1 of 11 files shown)


slandelle:
We can’t introduce a breaking change on such core API atm.
The best we can do is flag as deprecated.

client/src/main/java/org/asynchttpclient/Request.java
```diff
@@ -180,4 +180,16 @@
    * @return the NameResolver to be used to resolve hostnams's IP
    */
   NameResolver<InetAddress> getNameResolver();
+
+  /**
+   * @return a new request builder using this request as a prototype
+   */
+  RequestBuilder toBuilder();
```
slandelle: This is a breaking change. Anyone with his own implementation like you would be forced to modify his code and recompile. Please make it a default method.

client/src/main/java/org/asynchttpclient/Request.java
```diff
@@ -180,4 +180,16 @@
...
+
+  /**
+   * @param disableUrlEncoding whether to disable url encoding or not
+   * @param validateHeaders whether to enable header validation or not
+   * @return a new request builder using this request as a prototype
```
slandelle: Unused. Please remove.

client/src/main/java/org/asynchttpclient/Request.java
kringol: Agreed. Moved as default method

client/src/main/java/org/asynchttpclient/Request.java
kringol: Removed

kringol:
> We can’t introduce a breaking change on such core API atm.
> The best we can do is flag as deprecated.

Flagged both constructors as deprecated for now then. And suppressed the deprecation warning on the interface default implementation for the time being. If that doesn't sound good, I can just remove the deprecation and hope people will not use it directly

slandelle:
Thanks!

PR #1675, opened 2019-11-13
Remove unused dependency

cesarsotovalero:
Hello, I noticed that dependency `netty-resolver-dns` is declared in module `async-http-client`. However, this dependency is not used and can be safely removed to make the library slimmer and the dependency tree less complex.


slandelle:
Thanks

Issue #1584, opened 2018-10-06
should config.getThreadFactory take a pool name or suffix?


dsilvasc:
Both `DefaultAsyncHttpClient` and `ChannelManager` call `config.getThreadFactory`, but when the result is null, `DefaultAsyncHttpClient` instantiates a `DefaultThreadFactory` with the pool name `config.getThreadPoolName() + "-timer"` while `ChannelManager` instantiates that same factory with the pool name `config.getThreadPoolName()`.

This means that when you don't set a thread factory in the client config, you'll end up with two different pool names, but if you do set a thread factory in the client config, you'll end up with the same pool name for both use cases.

Should `getThreadFactory` take a pool name (or even just a suffix) to allow differentiating its use cases in `DefaultAsyncHttpClient` and `ChannelManager`?

mchernyakov:
Hi @slandelle, is it still actual? I would like to take it.

slandelle:
Hi @mchernyakov 
Yes, feature request is still open, please feel free to contribute.

PR #1687, opened 2020-01-02
Channel thread factory #1584

mchernyakov:
Hi @slandelle, I offer PR with a new factory for Netty's Channels. I've also added a specific prefix for that group.

Was not sure about `config.getThreadPoolName() + "-timer"`, decided to keep it as we have it now.


mchernyakov:
@slandelle do you have any thoughts about this PR?

Issue #1686, opened 2019-12-27
Delay between RequestTimeout and ListenableFuture's completion


xiaofud:
Hi, I find the delay between RequestTimeout and the time corresponding future completes by a timeout exception is quite long. Below is my test.

version:  

```xml
<dependency>
	<groupId>org.asynchttpclient</groupId>
	<artifactId>async-http-client</artifactId>
	<version>2.10.4</version>
</dependency>
```

test code:

```java
@RequestMapping("/api/timeout_test")
public BaseApiResponse<Long> timeoutTest() {
    long timeout = 1000;
    try {
        Thread.sleep(timeout);
    } catch (InterruptedException e) {
        e.printStackTrace();
    }
    BaseApiResponse<Long> objectBaseApiResponse = new BaseApiResponse<>();
    objectBaseApiResponse.setData(timeout);
    return objectBaseApiResponse;
}

@RequestMapping("/api/timeout")
public BaseApiResponse<String> timeout() {

    DefaultAsyncHttpClientConfig.Builder builder = new DefaultAsyncHttpClientConfig.Builder();
    builder.setConnectTimeout(5);
    AsyncHttpClient client = asyncHttpClient(builder);
    int timeout = 20;
    String url = "http://localhost:8080/api/timeout_test";

    BaseApiResponse<String> response = new BaseApiResponse<>();
    RequestBuilder requestBuilder = new RequestBuilder();
    Request request = requestBuilder.setUrl(url).setRequestTimeout(timeout).build();

    long beforeFutureCreated = System.currentTimeMillis();
    ListenableFuture<Response> responseListenableFuture = client.executeRequest(request);
    long futureCreated = System.currentTimeMillis();

    responseListenableFuture.addListener(() -> {
        long futureComplete = System.currentTimeMillis();
        String msg = String.format("future creation: %s, future complete: %s", futureCreated - beforeFutureCreated, futureComplete - beforeFutureCreated);
        response.setMessage(msg);
    }, Runnable::run);
    try {
        responseListenableFuture.get();
    } catch (Exception e) {
        e.printStackTrace();
    }
    return response;
}
```

result:

I tried many times, and the future complete time is always much longer than 20ms. It doesn't matter if I use a pre-created AsyncHttpClient or create a new one on each request.

```json
{"code":0,"message":"future creation: 0, future complete: 103","data":null}
{"code":0,"message":"future creation: 0, future complete: 85","data":null}
[...]

PR #1705, opened 2020-03-12
#1686 configure accuracy in HashedTimerWheel

zagorulkinde:
Fix issue #1686


slandelle:
Could you please elaborate on the benefits of this PR? In particular how the defaults are not suited for you.

zagorulkinde:
Hi. In this PR i'd resolved the issue https://github.com/AsyncHttpClient/async-http-client/issues/1686 to decrease delay between RequestTimeout and ListenableFuture's completion. I want to get responses as fast as possible. Also i want to mention that default values now the same as it was before this PR.

zagorulkinde:
Hello @slandelle what do you think?

slandelle:
Why not. Could you please make it more simple and remove the TimeUnit and pass the duration as a number of millis?

zagorulkinde:
removed TimeUnit and passed duration as number of millis;

zagorulkinde:
> Why not. Could you please make it more simple and remove the TimeUnit and pass the duration as a number of millis?

Hey. Any feedback?

client/src/main/resources/org/asynchttpclient/config/ahc-default.properties
```diff
@@ -50,3 +50,5 @@ org.asynchttpclient.shutdownQuietPeriod=2000
 org.asynchttpclient.shutdownTimeout=15000
 org.asynchttpclient.useNativeTransport=false
 org.asynchttpclient.ioThreadsCount=0
+org.asynchttpclient.hashedWheelTimerTickDuration=100
+org.asynchttpclient.hashedWheelTimerSize=512
```
slandelle: nit: please add line feed

zagorulkinde: done

slandelle:
Thanks!

Issue #1580, opened 2018-09-28
CookieStore slows down with increasing number of cookies


joshovi:
In https://github.com/AsyncHttpClient/async-http-client/blob/365c1e6a582499f8aaec30d56a398e694033e999/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java#L160 the look up over cookies traverses all entries in the map. The side effect to clean all expired cookies traverses the list again. The cost for algorithm scales with the number of cookies.

In our project we had to disable the cookie store, i.e.
`new DefaultAsyncHttpClientConfig.Builder()
.setCookieStore(null)
...
.build()`  

Maybe you could rewrite the function using another data structure that facilitates the look up for candidate cookies and then make a lazy removal of the expired candidates, or remove expired cookies asynchronously and periodically.

sivaalli:
@slandelle Can I work on a fix for this?

slandelle:
@sivaalli sure, contributions welcome

PR #1708, opened 2020-03-19
Cookiejar optimization

sivaalli:
Closes: https://github.com/AsyncHttpClient/async-http-client/issues/1580

Changes:
1. Segmented map based on domain names. So instead of traversing all the domains it traverses the domains that are of interest.
2. Use NettyTimer to clean up the expired cookies asynchronously. The timer task that provides this functionality is CookieEvictionTask.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClientConfig.java
+++ b/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClientConfig.java
@@ -109,6 +109,7 @@ public class DefaultAsyncHttpClientConfig implements AsyncHttpClientConfig {
 
   // cookie store
   private final CookieStore cookieStore;
+  private final int expiredCookieEvictionDelay;
 
   // internals
   private final String threadPoolName;
@@ -192,6 +193,7 @@ public class DefaultAsyncHttpClientConfig implements AsyncHttpClientConfig {
 
                                        // cookie store
                                        CookieStore cookieStore,
+                                       int expiredCookieEvictionDelay,
 
                                        // tuning
                                        boolean tcpNoDelay,
@@ -283,6 +285,7 @@ public class DefaultAsyncHttpClientConfig implements AsyncHttpClientConfig {
 
     // cookie store
     this.cookieStore = cookieStore;
+    this.expiredCookieEvictionDelay = expiredCookieEvictionDelay;
 
     // tuning
     this.tcpNoDelay = tcpNoDelay;
@@ -558,6 +561,11 @@ public class DefaultAsyncHttpClientConfig implements AsyncHttpClientConfig {
     return cookieStore;
   }
 
+  @Override
+  public int expiredCookieEvictionDelay() {
+    return expiredCookieEvictionDelay;
+  }
+
   // tuning
   @Override
   public boolean isTcpNoDelay() {
@@ -746,6 +754,7 @@ public class DefaultAsyncHttpClientConfig implements AsyncHttpClientConfig {
 
     // cookie store
     private CookieStore cookieStore = new ThreadSafeCookieStore();
+    private int expiredCookieEvictionDelay = defaultExpiredCookieEvictionDelay();
 
     // tuning
     private boolean tcpNoDelay = defaultTcpNoDelay();
@@ -1146,6 +1155,11 @@ public class DefaultAsyncHttpClientConfig implements AsyncHttpClientConfig {
       return this;
     }
 
+    public Builder setExpiredCookieEvictionDelay(int expiredCookieEvictionDelay) {
+      this.expiredCookieEvictionDelay = expiredCookieEvictionDelay;
+      return this;
+    }
+
     // tuning
     public Builder setTcpNoDelay(boolean tcpNoDelay) {
       this.tcpNoDelay = tcpNoDelay;
@@ -1330,6 +1344,7 @@ public class DefaultAsyncHttpClientConfig implements AsyncHttpClientConfig {
               responseFilters.isEmpty() ? Collections.emptyList() : Collections.unmodifiableList(responseFilters),
               ioExceptionFilters.isEmpty() ? Collections.emptyList() : Collections.unmodifiableList(ioExceptionFilters),
               cookieStore,
+              expiredCookieEvictionDelay,
               tcpNoDelay,
               soReuseAddress,
               soKeepAlive,
--- a/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
+++ b/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
@@ -21,7 +21,6 @@ import org.asynchttpclient.util.MiscUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.Closeable;
 import java.io.IOException;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
--- a/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
+++ b/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
@@ -82,7 +82,6 @@ public final class ThreadSafeCookieStore implements CookieStore {
 
   ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-  // Visible for test
   public Map<String, Map<CookieKey, StoredCookie>> getUnderlying() {
     return new HashMap<>(cookieJar);
   }
--- a/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
+++ b/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
@@ -31,7 +31,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
-public final class ThreadSafeCookieStore implements CookieStore, Closeable {
+public final class ThreadSafeCookieStore implements CookieStore {
   private static final Logger log = LoggerFactory.getLogger(ThreadSafeCookieStore.class);
 
   private final Map<String, Map<CookieKey, StoredCookie>> cookieJar = new ConcurrentHashMap<>();
--- a/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
+++ b/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
@@ -18,15 +18,24 @@ import io.netty.handler.codec.http.cookie.Cookie;
 import org.asynchttpclient.uri.Uri;
 import org.asynchttpclient.util.Assertions;
 import org.asynchttpclient.util.MiscUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.io.Closeable;
+import java.io.IOException;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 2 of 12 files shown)


sivaalli:
@slandelle Please let me know if this looks good. Thanks

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -126,13 +149,6 @@ private boolean hasCookieExpired(Cookie cookie, long whenCreated) {
       return false;
   }
 
-  // rfc6265#section-5.1.3
-  // check "The string is a host name (i.e., not an IP address)" ignored
-  private boolean domainsMatch(String cookieDomain, String requestDomain, boolean hostOnly) {
-    return (hostOnly && Objects.equals(requestDomain, cookieDomain)) ||
-            (Objects.equals(requestDomain, cookieDomain) || requestDomain.endsWith("." + cookieDomain));
-  }
-
```
slandelle: It looks to me that your whole PR is built on the assumption matching domains on strict equality is enough.
This is wrong, please read RFC6265, in particular section about domain matching.

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -126,13 +149,6 @@ private boolean hasCookieExpired(Cookie cookie, long whenCreated) {
       return false;
   }
 
-  // rfc6265#section-5.1.3
-  // check "The string is a host name (i.e., not an IP address)" ignored
-  private boolean domainsMatch(String cookieDomain, String requestDomain, boolean hostOnly) {
-    return (hostOnly && Objects.equals(requestDomain, cookieDomain)) ||
-            (Objects.equals(requestDomain, cookieDomain) || requestDomain.endsWith("." + cookieDomain));
```
sivaalli: Taking advantage that:
1. Domain matches exactly, in which case it does not matter if its hostOnly domain or not
2. request domain is subdomain of cookie domain.

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -145,45 +161,76 @@ private void add(String requestDomain, String requestPath, Cookie cookie) {
...
-    final boolean[] removeExpired = {false};
+    final List<Cookie> results = new ArrayList<>();
+    boolean exactDomainMatch = true;
+    String topLevelSubdomain = domain;
 
-    List<Cookie> result = cookieJar.entrySet().stream().filter(pair -> {
+    while (MiscUtils.isNonEmpty(topLevelSubdomain)) {
+      results.addAll(getStoredCookies(topLevelSubdomain, path, secure, exactDomainMatch));
+      exactDomainMatch = false;
+      topLevelSubdomain = DomainUtils.getTopLevelSubdomain(topLevelSubdomain);
```
sivaalli: Taking advantage that:

1. Domain matches exactly, in which case it does not matter if its hostOnly domain or not
2. Request domain is subdomain of cookie domain.

slandelle:
Nice work.

However, I have some concerns regarding CookieStore handling its own scheduling.

For example, someone spawning multiple AHC instances with different configs would spawning multiple threads for nothing.
Or someone could want to share a single CookieStore amongst multiple instances with different configurations and would spawn multiple ThreadSafeCookieStores for nothing in DefaultAsyncHttpClientConfig#Builder (prior to overriding).

I think it would be better to use AHC's HashedWheelTimer instance for scheduling.

sivaalli:
Thanks for the feedback. Let me change code to use HashedWheelTimer.

client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClientConfig.java
```diff
@@ -1252,7 +1254,21 @@ private ProxyServerSelector resolveProxyServerSelector() {
...
+      timer.start();
+      return timer;
+    }
+
     public DefaultAsyncHttpClientConfig build() {
+      if (nettyTimer == null)
+        nettyTimer = newNettyTimer(this);
+
+      if (cookieStore == null)
+        cookieStore = new ThreadSafeCookieStore(nettyTimer);
```
slandelle: This causes a slight behavior change: there's no longer any way to disable CookieStore, except providing a noop implementation. And it causes all the existing `if (cookieStore != null)` guards useless.

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -15,18 +15,29 @@
...
 import java.util.stream.Collectors;
 
 public final class ThreadSafeCookieStore implements CookieStore {
+  private static final int COOKIE_REMOVAL_DELAY_IN_SECONDS = 30;
+
+  private final Map<String, Map<CookieKey, StoredCookie>> cookieJar = new ConcurrentHashMap<>();
+  private io.netty.util.Timer timer;
 
-  private Map<CookieKey, StoredCookie> cookieJar = new ConcurrentHashMap<>();
+  public ThreadSafeCookieStore(io.netty.util.Timer timer) {
```
slandelle: Again, why does the CookieStore need to be aware of the Timer that cleans it up? I think it should expose methods for an external timer to clean it up. As proof it would be a better design: it would make testing way easier.

sivaalli: Sure. Let me fix this. Thanks for the feedback.

sivaalli:
@slandelle Please let me know what you think. Thanks.

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -145,45 +149,64 @@ private void add(String requestDomain, String requestPath, Cookie cookie) {
...
+    else {
+      cookieJar.putIfAbsent(keyDomain, new ConcurrentHashMap<>());
+      cookieJar.get(keyDomain).put(key, new StoredCookie(cookie, hostOnly, cookie.maxAge() != Cookie.UNDEFINED_MAX_AGE));
+    }
   }
 
   private List<Cookie> get(String domain, String path, boolean secure) {
 
     final boolean[] removeExpired = {false};
+    final List<Cookie> results = new ArrayList<>();
```
slandelle: Please reduce allocations here.
Default ArrayList size is 10, it's pretty unlikely that we get more than 2 cookies for a given domain. Also, it can be allocated lazily on first match (lots of AHC usage won't involve cookies at all).

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -145,45 +149,64 @@ private void add(String requestDomain, String requestPath, Cookie cookie) {
...
-    CookieKey key = new CookieKey(cookie.name().toLowerCase(), keyDomain, keyPath);
+    CookieKey key = new CookieKey(cookie.name().toLowerCase(), keyPath);
 
     if (hasCookieExpired(cookie, 0))
-      cookieJar.remove(key);
-    else
-      cookieJar.put(key, new StoredCookie(cookie, hostOnly, cookie.maxAge() != Cookie.UNDEFINED_MAX_AGE));
+      cookieJar.getOrDefault(keyDomain, Collections.emptyMap()).remove(key);
+    else {
+      cookieJar.putIfAbsent(keyDomain, new ConcurrentHashMap<>());
```
slandelle: Please use computeIfAbsent instead to not allocate ConcurrentHashMap on existing key.

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -145,45 +149,64 @@ private void add(String requestDomain, String requestPath, Cookie cookie) {
...
+      cookieJar.get(keyDomain).put(key, new StoredCookie(cookie, hostOnly, cookie.maxAge() != Cookie.UNDEFINED_MAX_AGE));
```
slandelle: Don't perform an extra get, just use the value returned by computeIfAbsent above.

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -145,45 +149,64 @@ private void add(String requestDomain, String requestPath, Cookie cookie) {
...
+  }
 
-    List<Cookie> result = cookieJar.entrySet().stream().filter(pair -> {
+  private List<Cookie> getStoredCookies(String domain, String path, boolean secure,
+                                        boolean[] removeExpired, boolean isExactMatch) {
+    if (!cookieJar.containsKey(domain)) {
+      return Collections.emptyList();
+    }
+
+    return cookieJar.get(domain).entrySet().stream().filter(pair -> {
```
slandelle: Perform one single map look up. Perform a get and check for nullity above.

slandelle:
Looking good. Now, I think the last missing piece is to handle CookieStore sharing from multiple AHC instances. Currently, if we share a CookieStore, we'll end up scheduling cleaning multiple times, and never stop scheduling.

I suggest we use an AtomicInteger to keep track of AHC instance a given CookieStore is configured on, only start scheduling when starting first client and only stop scheduling when closing last one.

WDYT?

sivaalli:
Sounds good. Will start working on that.

slandelle:
Thanks a lot for your great work!

sivaalli:
@slandelle I've a few questions that I would like to get your opinion (Both cases assume sharing cookie store).

1. In case of shared netty timer supplied by the user. We just maintain a counter on cookiestore and start scheduling if it is the first client and user takes care of stopping netty timer (hopefully after stopping all AHC instances). No issue with this situation.
2. If user did not supply NettyTimer then we [create](https://github.com/AsyncHttpClient/async-http-client/blob/a0c036581ccb2c30114820a432254ae37d180c2f/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java#L89) new timer for each AHC instance. We can start scheduling **only** if it is first client on the first netty timer. But as soon as the user closes the first netty timer we would close the timer. But if user decides to use other AHC instances then we could never evict expired cookies.

I'm assuming that you made the comment above assuming that user shares a timer? Please correct me if my assumption is wrong.

If user does not share timer, then I guess we have to schedule eviction task on each timer.

sivaalli:
I've made some changes assuming you are talking about shared netty timer. Please let me know how this looks. Happy to implement any of your suggestions. 

I do not know if sharing netty timer across all AHC instances is useful or harms the overall performance (because all tasks share same timer). But if you'd like to see sharing a single netty timer (instead if creating a new one for each instance) in-case a user does not provide one, I can make that change as well.

slandelle:
> I'm assuming that you made the comment above assuming that user shares a timer? Please correct me if my assumption is wrong.

Yes

> If user does not share timer, then I guess we have to schedule eviction task on each timer.

Let's do that

> Also a follow up question regarding creating new Timer for each AHC instance. I feel like we could share the same Timer for multiple AHC's, but I think there might be a strong reason on why not too. But just curios to know. Timer starts to emit error logs when the count is greater that 64.

Could you please share the error logs?

sivaalli:
I've personally not created too many instances to see this message. But I was checking through source code of HasedWheelTimer and found that it spits out an error log when number of instances are greater than 64. Logs like this:

> You are creating too many HashedWheelTimer instances. HashedWheelTimer is a shared resource that must be reused across the JVM, so that only a few instances are created.

The changes are made according to your expectation. Please let me know otherwise.

slandelle:
@sivaalli Thanks!

Issue #1710, opened 2020-03-23
getClientStats throws NullPointerException


yunglin:
version: 2.5.4
```
java.lang.NullPointerException: null
    at org.asynchttpclient.netty.channel.ChannelManager.lambda$getClientStats$2(ChannelManager.java:483)
    at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:174)
    at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
    at java.util.Iterator.forEachRemaining(Iterator.java:116)
    at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
    at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
    at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
    at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:566)
    at org.asynchttpclient.netty.channel.ChannelManager.getClientStats(ChannelManager.java:484)
    at org.asynchttpclient.DefaultAsyncHttpClient.getClientStats(DefaultAsyncHttpClient.java:285)
```

PR #1711, opened 2020-03-23
fixed NPE caused by open but unconnected channels.

yunglin:
This PR fixed a NullPointerException caused by open but unconnected channels. 
The issue is reported: https://github.com/AsyncHttpClient/async-http-client/issues/1710

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
@@ -61,6 +61,7 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -490,7 +491,7 @@ public EventLoopGroup getEventLoopGroup() {
   }
 
   public ClientStats getClientStats() {
-    Map<String, Long> totalConnectionsPerHost = openChannels.stream().map(Channel::remoteAddress).filter(a -> a.getClass() == InetSocketAddress.class)
+    Map<String, Long> totalConnectionsPerHost = openChannels.stream().map(Channel::remoteAddress).filter(Objects::nonNull).filter(a -> a.getClass() == InetSocketAddress.class)
```
slandelle: Actually, there are 2 issues: you could be using one single filter, and (existing issue), equality comparison shouldn't be using ==.

`filter(a -> a instanceof InetSocketAddress)`

slandelle:
Thanks!

PR #1715, opened 2020-04-15
Upgrade to jakarta.activation 1.2.2

amarkevich:


slandelle:
This suggested change is broken. We use the activation framework because we need the `MimetypesFileTypeMap` class and its associated `mimetypes.default` file.

In `com.sun.activation:javax.activation` jar, `mimetypes.default` is shipped as expected in `/META-INF`.
This file is missing from `jakarta.activation:jakarta.activation-api`.

amarkevich:
@slandelle switched to 'com.sun.activation:jakarta.activation:jar:1.2.2'

slandelle:
Thanks

PR #1723, opened 2020-05-08
Issue #1721: Improve exceptional behavior in reactive streams

### LINKED ISSUE #1721: False timeouts when using StreamedAsyncHandler
When using the `StreamedAsyncHandler` to download large files, I found what I think is a bug in how timeouts are handled.

The timeout mechanism relies on when the `ListenableFuture` for the request was last `touch()`ed. `touch()` is only called when a `HttpResponseBodyPart` is delivered to the downstream. This works fine if the downstream is downloading the response as fast as possible.

However, if the downstream has no outstanding `request()` (like if it's serving as a proxy and its own downstream is slow or whatever), there may be long periods of time where there are no body parts published because none have been requested. In those periods, the request may "time out" because of inactivity.

I think we should change the timeout logic to only count this as a timeout if there has been no activity AND there is no outstanding request on the `Publisher`. This can be implemented by periodically `touch()`ing the `ListenableFuture` if there is no outstanding `request()` on the `Publisher`.

I am willing to implement it if we want this feature.

n-miles:
Fixes #1721

Changes:
* Errors on the request are now propagated to reactive subscribers instead of just to the request's ListenableFuture
* Read timeouts can no longer occur if a reactive streams subscriber has no outstanding request. Note that this does not affect request timeouts - only read timeouts.
* Cleaned up some random styling things

The original purpose of this change was to fix the read timeouts, but in the process of testing my change, I discovered that errors are not currently propagated to the subscriber, so I changed that as well.

I can split the changes into separate commits if you'd like.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java
@@ -39,8 +39,7 @@ public final class HttpHandler extends AsyncHttpClientHandler {
     super(config, channelManager, requestSender);
   }
 
-  private boolean abortAfterHandlingStatus(//
-                                           AsyncHandler<?> handler,
+  private boolean abortAfterHandlingStatus(AsyncHandler<?> handler,
                                            NettyResponseStatus status) throws Exception {
     return handler.onStatusReceived(status) == State.ABORT;
   }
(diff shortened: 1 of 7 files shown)


n-miles:
@slandelle have you had a chance to look at this change?

slandelle:
Honestly, I'm not familiar with reactive streams. Code style and test LGTM.

Issue #1720, opened 2020-05-05
InputStream lost on 308 redirect, ByteArray not


christiangroth:
Hi everybody,

I'm currently fiddling around with AHC and multipart requests. So far so good and everything is fine. However there is a special case were a PUT request gets a HTTP 308 response and in this case the second request does not send the body anymore. I've seen that this should work according to https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java

I'm using the following code to create the Request:

    BoundRequestBuilder requestBuilder = client.preparePut(url);
    requestBuilder
        .addBodyPart(new StringPart("properties", properties, "application/json"))
        .addBodyPart(new InputStreamPart("xxx", data, "xxx"));

If I do not provide the data as InputStream but read all bytes into memory everything works fine, also following the HTTP 308 redirect.

    BoundRequestBuilder requestBuilder = client.preparePut(url);
    try {
        byte[] dataBytes = new byte[data.available()];
        data.read(dataBytes);
        requestBuilder
            .addBodyPart(new StringPart("properties", properties, "application/json"))
            .addBodyPart(new ByteArrayPart("xxx", dataBytes, null, StandardCharsets.UTF_8, "xxx"));
        } catch (IOException e) {
            Logging.logError("Unable to read data for media to be published to " + url + "!", e, getClass());
    }

So my question is: Are redirects not supported when streaming the media content using an InputStream or am I doing something wrong or is this a bug?

Bets regards
  Chris

slandelle:
Yes, you are doing something wrong.

The network is not instantaneous.
If your server choses to respond early without waiting for having received the full request, including its body, there's a very good chance the user-agent has already started writing the request body on its side of the socket.
If your request body is/contains an InputStream, it has already been consumed and can't be regenerated.

There's no way to handle this gracefully. Period.

Protecting against the possibility of the server performing an early response is exactly what [Expect-Continue](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/100) is for.

christiangroth:
Hi, 

one last question / suggestion on this: Would it be possible to implement a InputStreamProviderPart instead of an InputStreamPart? The InputStreamProviderPart would use a Lamba to create a fresh InputStream every time it's needed. 

[...]

PR #1726, opened 2020-06-05
Implemented reusable InputStreamPart

christiangroth:
Referring to #1720 I've implemented a simple solution to provide an InputStreamSupplier as Multipart Body. Under the hood however the implementation just creates a new InputStream and then delegated everything else to the already existing InputStreamPart. Nevertheless I enhanced the MultipartUploadTest to be sure the logic is tested. Hope this is fine!


pejobo:
Any chance this gets merged?

Issue #1731, opened 2020-07-27
Leaking memory when handling redirects


will-lauer:
We've encountered a memory leak when handling redirects. It looks like when a redirect is encountered by the system, it ends up reusing the existing NettyResponseFuture to handle the request to the new destination. This ends in turn replaces the old request timeout with a new one, orphaning the old timeout. Since references to timeouts are kept in the HashedWheelTimer, and since timeouts have references to the associated RequestTimoutTimerTask, which has a reference to the NettyResponseFuture, none of these objects get cleaned up until the timeout expires and the HashedWheelTimer goes through its periodic cleanup.

In our case, we had accidentially set the request timeout up to 10 hours. We also are using an AsyncResponseHandler to process requests. The AsyncResponseHandler maintains a reference to the ResponseBuilder, and since the Future maintains a reference the AsyncResponseHandler, the result was that all responses from redirected requests were kept in memory until the timeout expired, even though the response handler had finished processing them. Due to oddities of the services we interact with, around 50% of our requests were redirects, causing the system to build up tens of gigabytes of requests in memory in only a couple of hours before the system finally fell over.

slandelle:
Closed by #1732

PR #1732, opened 2020-07-30
Cancel replaced timeouts to avoid leak

will-lauer:
Fix for issue #1731.

When setting the TimeoutHolder, cancel any prior timeout so that they don't leak. 

Previously, they would just be lost and remain in the timer until their timeout expired, which could be a long time. Previously, encountering a redirect would trigger this code, causing the old request timer to be replaced with a new one. The old timer would maintain a link back to this Future, but couldn't be canceled by this future (as its reference had been overwritten) causing the Future, is associated Response, and any AsyncResponseHandler to be retained in memory instead of being garbage collected once the request had been processed.


slandelle:
Thanks!

PR #1748, opened 2020-11-25
Testing GH PR Action

TomGranot:
This is a PR that will trigger a Github actions that should test the project. Let's see how it goes!


TomGranot:
One might say that testing CI against the actual repo and not a fork is idiotic. One might even be right.
Closing in favour of testing under the fork.

PR #1756, opened 2020-12-17
Package level constructor for AsyncHttpClientState to be changed to public

sms0070:
We have the following requirement : 

- We use client certificate authentication. However, we want to have the client certificate dynamic and can be changed at runtime.
- For that, in the 1.9.x versions, we extended AsyncHttpClient and created a new client config object with a new SSLContext and use that to create a NettyAsyncHttpProvider instance and execute the request.
- NettyAsyncHttpProvider is not available in 2.x versions.
- Hence, we are trying to use NettyRequestSender.
- NettyRequestSender constructor needs a AsyncHttpClientState object.
- Since, AsyncHttpClientState constructor is package, we are unable to proceed.
- If we keep clientState as null, execute method won't work in NettyRequestSender due to isClosed check.

Thanks,
Suyash

### DIFF
--- a/client/src/main/java/org/asynchttpclient/AsyncHttpClientState.java
+++ b/client/src/main/java/org/asynchttpclient/AsyncHttpClientState.java
@@ -19,7 +19,7 @@ public class AsyncHttpClientState {
 
   private final AtomicBoolean closed;
 
-  AsyncHttpClientState(AtomicBoolean closed) {
+  public AsyncHttpClientState(AtomicBoolean closed) {
     this.closed = closed;
   }
 


TomGranot:
@sms0070 The Travis test fails (as it often does) due to DNS errors (the `MultipartBasicAuthTest.unauthorizedNonPreemptiveRealmCausesServerToCloseSocket` test specifically). The GitHub Actions does not fail, which indicates that this is probably isolated to Travis.

I'm worried might be a bit more nuanced, though, and so I need to think about this a bit more. I don't have a concrete opposing point yet, though, so please ping me back in a few days if you don't hear back from me.

sms0070:
@TomGranot : Can you check now?

sms0070:
Also, will it be possible to put this change in earlier versions ?
right now, we are using 2.7.0

TomGranot:
@sms0070 I actually didn't have time, sorry.

@RahulKushwaha  - following the other issue, can you take a look here instead? The gist is pretty much detailed in the beginning of the ticket. 

@sms0070 Regarding putting this in 2.7.0 - I don't like changing old versions with new changes. Once it's set it's set - but feel free to upgrade to the next release once we roll it out. It will contain this change as well, should it be accepted.

sms0070:
@TomGranot , @turbolag : any updates ?

TomGranot:
@sms0070 Sorry about this - if Rahul will not have time during the weekend then I'll take a deeper look. This is however will not reach maven central until I finish the CI stuff properly, so be aware.

sms0070:
Hi @TomGranot,
Any updates on this? I see checks have passed

TomGranot:
@sms0070 Yes, but recall that I wanted Rahul to take a look at this too:)

These types of smaller (yet structural) PRs are a great place to allow a new contributor to jump into the waters.

In addition, keep in mind that this will not be released until a new AHC version is released, and I don't currently publish snapshot versions on a regular basis.

Let me check with him, and if he can't I'll take a look. Thanks for reminding me, and sorry for the wait.

sms0070:
@TomGranot  : Kindly check this. Also, any timeline when we are planning for next release ?

hyperxpro:
Please do a new PR for Beta on main branch.

Issue #1769, opened 2021-01-27
NullPointerException


aqingsao:
I've been using async-http-client for a long time, it works fine in many cases even at a high QPS of 100K.
But recently some errors happen when calls to a specific api(nothing special, just a POST http with Content-Type: multipart/form-data), lots of exceptions are thrown at a low QPS of 800, and below is the stack trace:

`java.lang.NullPointerException: null
        at org.asynchttpclient.netty.channel.Channels.getAttribute(Channels.java:31) ~[async-http-client-2.12.2.jar!/:na]
        at org.asynchttpclient.netty.request.NettyRequestSender.abort(NettyRequestSender.java:467) ~[async-http-client-2.12.2.jar!/:na]
        at org.asynchttpclient.netty.handler.AsyncHttpClientHandler.exceptionCaught(AsyncHttpClientHandler.java:201) ~[async-http-client-2.12.2.jar!/:na]
        at io.netty.channel.AbstractChannelHandlerContext.invokeExceptionCaught(AbstractChannelHandlerContext.java:302) [netty-all-4.1.55.Final.jar!/:4.1.55.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:264) [netty-all-4.1.55.Final.jar!/:4.1.55.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:248) [netty-all-4.1.55.Final.jar!/:4.1.55.Final]
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelInactive(AbstractChannelHandlerContext.java:241) [netty-all-4.1.55.Final.jar!/:4.1.55.Final]
        at io.netty.handler.stream.ChunkedWriteHandler.channelInactive(ChunkedWriteHandler.java:138) [netty-all-4.1.55.Final.jar!/:4.1.55.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:262) [netty-all-4.1.55.Final.jar!/:4.1.55.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:248) [netty-all-4.1.55.Final.jar!/:4.1.55.Final]
[...]

PR #1771, opened 2021-02-09
Fix for NPE when connection is reset by peer

rachidbm:
A fix for the NullPointerException problem reported in #1769 
It looks like the NPE is introduced as programming error. because there is a null check on `channel`, but in the body it uses `future.channel()` which potential returns null.

I ca't oversee the impact of this change, but all tests pass when running `mvn clean verify` on my machine.


yazgoo:
Hi, any idea of when this can be merged ? I've also had instances where this error came up

rachidbm:
@TomGranot is there anything I can do to help to get this fix merged?

TomGranot:
@yazgoo @rachidbm Merged. Sorry for the wait!

TomGranot:
Some issues this closes:

https://github.com/AsyncHttpClient/async-http-client/issues/1769
https://github.com/AsyncHttpClient/async-http-client/issues/1779

PR #1776, opened 2021-02-28
we should pass proxy instead of target url to the parameter of method…

pavelliu:
… userProxyconnect()


pavelliu:
the https proxy may be totally not working, this PR is incorrect

Issue #1775, opened 2021-02-25
Bump netty version to 4.1.59.Final


andres-de-biase-wcar:
Vulnerabilities have been found in netty version 4.1.53. (Currently used). Please see [https://nvd.nist.gov/vuln/detail/CVE-2021-21290](https://nvd.nist.gov/vuln/detail/CVE-2021-21290)

hysterus1:
Pr is open: https://github.com/AsyncHttpClient/async-http-client/pull/1781

TomGranot:
Updated to .60.

PR #1782, opened 2021-03-16
Bump netty to 4.1.60.Final #1775

hysterus1:
Hello, I'm kindly asking to treat this PR with high priority as it involves security of the library. #1775


hysterus1:
Netty has added a special normalization and validation for the Content-Length header. In case the headers are duplicated, the received should either reject the response or only pick and parse the correct one. Long story short, with newest Netty one can't have this header duplicated.

That is not a big problem for async-http-client, as the old behavior was only relied on in the tests. I've adjusted the tests to simply use a different header instead.

hysterus1:
I will wait for the build to pass and then squash the commits.

TomGranot:
@hysterus1 Note that that CircleCI build passed, but the GitHub Actions one failed. Ran this locally and this passed too - GH had a few timeouts issues in the past due to (what I think is) the amount of "power" each allocated test agent gets (CPU, RAM, etc...).

Merged.

Issue #1660, opened 2019-08-12
Duplicate handler name: request-body-streamer


rossabaker:
Discussed in #1627, but spinning off as it's a separate issue.

Under load, when using the `NettyReactiveStreamsBody`, I can reliably reproduce the following error:

```
[info] java.lang.IllegalArgumentException: Duplicate handler name: request-body-streamer
[info] 	at io.netty.channel.DefaultChannelPipeline.checkDuplicateName(DefaultChannelPipeline.java:1066)
[info] 	at io.netty.channel.DefaultChannelPipeline.filterName(DefaultChannelPipeline.java:284)
[info] 	at io.netty.channel.DefaultChannelPipeline.addLast(DefaultChannelPipeline.java:204)
[info] 	at io.netty.channel.DefaultChannelPipeline.addLast(DefaultChannelPipeline.java:195)
[info] 	at org.asynchttpclient.netty.request.body.NettyReactiveStreamsBody.write(NettyReactiveStreamsBody.java:59)
[info] 	at org.asynchttpclient.netty.request.NettyRequestSender.writeRequest(NettyRequestSender.java:426)
[info] 	at org.asynchttpclient.netty.request.NettyRequestSender.sendRequestWithOpenChannel(NettyRequestSender.java:259)
[info] 	at org.asynchttpclient.netty.request.NettyRequestSender.sendRequestWithCertainForceConnect(NettyRequestSender.java:141)
[info] 	at org.asynchttpclient.netty.request.NettyRequestSender.sendRequest(NettyRequestSender.java:113)
[info] 	at org.asynchttpclient.DefaultAsyncHttpClient.execute(DefaultAsyncHttpClient.java:241)
[info] 	at org.asynchttpclient.DefaultAsyncHttpClient.executeRequest(DefaultAsyncHttpClient.java:210)
```

I have a [Scala reproduction with http4s](https://github.com/rossabaker/http4s-bugs/blob/issue-2779).  It can be seen with `sbt run`.

The problem is that that AHC offers the channel back to the pool manager as soon as the last response chunk is read, which causes a race condition against the `NettySubscriber`'s completion.  If `removeFromPipeline` is not called before the channel is offered back to the pool and polled out to another reactive streams request, we hit the exception.

A workaround is to remove the pipeline in a hook in `AsyncHandler`:

```scala
      override def onConnectionOffer(channel: Channel) {
        try channel.pipeline.remove("request-body-streamer")
        catch {
[...]

PR #1787, opened 2021-04-09
Always remove subscriber from pipeline when completing a reactive stream

hamnis:
* Only send LastHttpContent if channel is active
* Avoid async scheduling if possible


hamnis:
might be related to #1660

hamnis:
this was not enough to fix the issue, seems like something is broken in either the reactive streams handling or the use in http4s.

PR #1824, opened 2022-09-14
Fix for Duplicate handler name: request-body-streamer

Squiry:
Solves [that](https://github.com/AsyncHttpClient/async-http-client/issues/1660) issue.


hyperxpro:
Reactive Stream is no more in v3.0.0.

PR #1830, opened 2022-11-19
Migrate tests to JUnit 5

hyperxpro:


hyperxpro:
Tests will continue to fail. It needs fixes in the main client module.

PR #1832, opened 2022-11-20
Remove Reactive Streams

hyperxpro:


Squiry:
But why?..

hyperxpro:
I want to simplify the moving elements. There are already good and well-maintained implementations of Reactive Streams for HTTP.

We can consider Reactive Streams in the future once beta of AHC V3 is out. But for now, let's keep things simple.

PR #1864, opened 2023-04-23
Changed requestTimeout's type to java.time.Duration

Asapin:


hyperxpro:
Btw, do you wanna investigate flaky tests? They fail randomly and I have to use repetitions to make them pass which is a bad idea.

hyperxpro:
Also, check logs of this: https://github.com/AsyncHttpClient/async-http-client/pull/1859

hyperxpro (approved):

PR #1876, opened 2023-05-05
Enabled NullAway for oauth, proxy and resolver packages

Asapin:


pom.xml
```diff
@@ -231,7 +231,7 @@
                             -Xep:NullablePrimitive:ERROR
                             -Xep:NullOptional:ERROR
                             -XepExcludedPaths:.*/src/test/java/.*
-                            -XepOpt:NullAway:AnnotatedPackages=org.asynchttpclient.channel,org.asynchttpclient.config,org.asynchttpclient.cookie,org.asynchttpclient.exception,org.asynchttpclient.filter
+                            -XepOpt:NullAway:AnnotatedPackages=org.asynchttpclient.channel,org.asynchttpclient.config,org.asynchttpclient.cookie,org.asynchttpclient.exception,org.asynchttpclient.filter,org.asynchttpclient.oauth,org.asynchttpclient.proxy,org.asynchttpclient.resolver
```
hyperxpro, line 234: Let's break lines

Asapin: Unfortunately breaking lines will break everything - I guess NullAway expects this `XepOpt:NullAway:AnnotatedPackages` parameter to be a single line without any breaks.
We can invert this parameter, i.e. instead of `include only these packages` we can have `include all packages except the following`. And I intend to eventually do that. But this requires fixing classes in the root `org.asynchttpclient` package first

hyperxpro: Alright, let's keep it then.

hyperxpro:
Let me know once ready to merge.

hyperxpro:
@Asapin Ping...

Asapin:
I'm currently working on enabling NullAway for the root `org.asynchttpclient` package, and there are quite a lot of changes (13 files at the moment), so maybe it's better to merge this PR as is?

PR #1878, opened 2023-05-06
Enabled NullAway for the root package

Asapin:
After this one every next PR will remove one package from the `-XepOpt:NullAway:UnannotatedSubPackages` argument until this argument is empty


client/src/main/java/org/asynchttpclient/util/EnsuresNonNull.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * NullAway supports @EnsuresNonNull(param) annotation, but org.jetbrains.annotations doesn't include this annotation.
+ * The purpose of this annotation is to tell NullAway that if the annotated method succeeded without any exceptions,
+ * all class's fields defined in "param" will be @NotNull.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target({ElementType.METHOD})
+public @interface EnsuresNonNull {
+    String[] value();
+}
```
hyperxpro, line 17:
```suggestion
}

```

PR #1887, opened 2023-06-05
Enabled NullAway for NTLM package

Asapin:


client/src/main/java/org/asynchttpclient/ntlm/NtlmEngine.java
```diff
@@ -381,19 +373,19 @@ public byte[] getTimestamp() {
         /**
          * Calculate the NTLMv2Blob
          */
-        public byte[] getNTLMv2Blob() {
+        public byte[] getNTLMv2Blob(byte[] nonNullTargetInformation) {
```
Asapin, line 376: This (and some other methods) use variables that were checked for null before the method call.
Essentially, they are used like this:
```java
public class MyClass {
    @Nullable
    private String field;

    void method1() {
        if (field != null) {
            method2();
        }
    }

    void method2() {
        System.out.println("Field length: " + field.length());
    }
}
```

Calling ` field.length()` inside `method2` is technically incorrect, but it's actually safe because this method is only called inside `method1` after we check that `field != null` . 
But this pattern doesn't work with NullAway, so instead, after we check that `field != null`, I introduced a new variable `String nonNullField = field;` and use this variable as a parameter when calling `method2`.

hyperxpro: We should remove `nonNull` prefix from these variables. Doesn't look too good.

Asapin: Removed `nonNull` prefix and added comments why these methods have `targetInformation` parameter and what is the difference between it and the field with the same name

hyperxpro: Perfect, thanks. :)

PR #1890, opened 2023-06-23
Add config for allowDuplicateContentLengths

### LINKED ISSUE #1889: Add configuration options to enable HttpClientCodec.allowDuplicateContentLengths flag
We're using AsyncHttpClient to proxy requests, some of which contain multiple Content-Length flags, and for a few different reasons we don't want to remove the duplicates. Currently AHC doesn't allow configuration of that flag, nor is there any easy way to replace the ChannelManager with our own that does what we need. This will require two configuration variables in AHC, since HttpClientCodec doesn't have a builder interface so we'll need to set both `parseHttpAfterConnectRequest` and `allowDuplicateContentLengths` (both with defaults of false). I'll submit a PR to implement this.

dgolombek:
Allow users to configure HttpClientCodec's
parseHttpAfterConnectRequest and allowDuplicateContentLengths flags.

Fixes #1889


hyperxpro:
PTAL at test cases

Error:  Failures: 
Error:    PerRequestRelative302Test.redirected302Test:86 expected: <403> but was: <200>
Error:    PerRequestRelative302Test.runAllSequentiallyBecauseNotThreadSafe:73->redirected302Test:86 expected: <403> but was: <200>

PR #1894, opened 2023-06-29
Enabled NullAway for ws package

Asapin:


Asapin:
Looks like Microsoft has banned programmatic access to their site or did something similar

PR #2026, opened 2024-11-10
use jspecify annotations

sullis:
https://jspecify.dev/docs/user-guide/


sullis:
WDYT? @hyperxpro

hyperxpro:
I think we should stick with JetBrains because it's widely used and integrates well with IntelliJ.

hyperxpro:
lol this thing is useless

Issue #1964, opened 2024-06-17
Cookie explicitly set on the request is getting discarded in favor of the one from the cookie jar


Spikhalskiy:
# Expected Behavior

A cookie explicitly specified on a RequestBuilder should take precedence over a cookie stored in cookie jar.

# Actual Behavior

```
cookieStore = httpClient.getCookieStore();
cookieStore.add("some_url", new DefaultCookie("name", "value1"));
BoundRequestBuilder requestBuilder = httpClient.prepareGet("some_url");
requestBuilder.addCookie(new DefaultCookie("name", "value2"));
httpClient.executeRequest(requestBuilder.build())
```

This code leads to a request being performed with cookie "name":"value1", which is not what is expected by a user when a user explicitly sets a cookie on a `RequestBuilder`.

# Root cause

Cookies from cookie jar [are overriding cookies coming from the RequestBuilder here](https://github.com/AsyncHttpClient/async-http-client/blob/2200b2477c84d2517fcbdf5d42f71d459c09066f/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java#L238)

# Affected version

2.12.3, 3.0 pre-releases

pickypg:
The supplied link puts the commit hash in the wrong place, but the root cause is that it blindly overrides Cookies by using `addOrReplace(cookie)` from the `CookieStore`.

https://github.com/AsyncHttpClient/async-http-client/blob/661ec72/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java#L238

pickypg:
I've added a PR that only uses `Cookie`s from the `CookieStore` if they're unset.

https://github.com/AsyncHttpClient/async-http-client/pull/2033

pickypg:
As a note, you can workaround this issue -- if you don't actually want the `CookieStore` -- by disabling it upon client creation:

```java
DefaultAsyncHttpClientConfig.Builder clientBuilder = Dsl.config()
    .setCookieStore(null)
```

PR #2033, opened 2024-11-14
[CookieStore] Only set `Cookie`s if they are not already set

pickypg:
This changes the behavior of the automatic usage of the `CookieStore` to avoid overwriting already-set `Cookie`s and, instead only sets them if they do not exist yet.

Closes https://github.com/AsyncHttpClient/async-http-client/issues/1964

### DIFF
--- a/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
+++ b/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClient.java
@@ -235,7 +235,7 @@ public class DefaultAsyncHttpClient implements AsyncHttpClient {
                 if (!cookies.isEmpty()) {
                     RequestBuilder requestBuilder = request.toBuilder();
                     for (Cookie cookie : cookies) {
-                        requestBuilder.addOrReplaceCookie(cookie);
+                        requestBuilder.addCookieIfUnset(cookie);
                     }
                     request = requestBuilder.build();
                 }
--- a/client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
+++ b/client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
@@ -323,6 +323,21 @@ public abstract class RequestBuilderBase<T extends RequestBuilderBase<T>> {
      * @return this
      */
     public T addOrReplaceCookie(Cookie cookie) {
+        return maybeAddOrReplaceCookie(cookie, true);
+    }
+
+    /**
+     * Add a cookie based on its name, if it does not exist yet. Cookies that
+     * are already set will be ignored.
+     *
+     * @param cookie the new cookie
+     * @return this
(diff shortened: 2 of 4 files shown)


hyperxpro:
I'll take a look this weekend.

hyperxpro:
Sadly this is a breaking change. `RequestBuilderBase` is an abstract class and adding anything will break compatibility. Perhaps, you might want to take a different approach.

pickypg:
I'm happy to if you have a suggestion? The current functionality is buggy given that it overrides explicitly provided cookies.

hyperxpro:
I'd say change this directly in `RequestBuilder`. People extending `RequestBuilderBase` will be safe from this breaking change.

pickypg:
That's doable, but this breaking change is clearly necessary.

hyperxpro:
Yup, I do feel this but as a library maintainer, I cannot. I am at fault, I should have made that as `interface`. But doing with directly changing should be best because people extending `RequestBuilderBase` should always take extra care of their custom implementations.

hyperxpro (approved):

lhotari:
Is there any way to backport this to 2.12.x to maintain Java 8 support?

We use AsyncHttpClient 2.12.x in the Apache Pulsar Admin client. Our challenge in Apache Pulsar is that the minimum Java requirement for clients is Java 8. Since AsyncHttpClient 3.0.x requires Java 11, we don't have a way to eliminate CVE-2024-53990 from our dependencies. While it's possible to set the `CookieStore` to `null` in AHC 2.12.x (as we're doing in https://github.com/apache/pulsar/pull/23725), the main issue for Apache Pulsar is that many users have security policies requiring the elimination of dependencies with high or critical vulnerabilities. Therefore, configuring the CookieStore isn't a sufficient mitigation for many of our users.

Any suggestions and recommendations for solving our use case in Apache Pulsar would be much appreciated.

hyperxpro:
@lhotari Sure, yes. I will push the emergency release for 2.12.x.

hyperxpro:
Released 2.12.4; it will sync to Maven Central in some time :)

@pickypg Can you do PR in the advisory database to exclude 2.12.4?

pickypg:
Can do!

xuwei-k:
missing 2.12.4 tag on GitHub? 👀

aidandempsey:
@hyperxpro Thanks for the fix, is there any chance this will be back ported to `2.12.3`?  Thanks!

hyperxpro:
> missing 2.12.4 tag on GitHub? 👀

Added `async-http-client-project-2.12.4`

xuwei-k:
thank!

pickypg:
Updated the GitHub Advisory: https://github.com/advisories/ghsa-mfj5-cf8g-g2fv

Issue #2071, opened 2025-03-09
Persistent HTTP 400 Errors After Failed CONNECT Request


jasonjoo2010:
Hi folks,

We rely heavily on AHC for its performance and async support in many internal projects. However, we've recently encountered multiple escalations regarding unusual HTTP 400 errors.

# Context

Our applications operate under strict network restrictions and must use an HTTP proxy to access external servers. This setup has worked reliably for years.

# Issue Details

We've identified that when initializing an HTTPS tunnel via a proxy:

1. If the initial CONNECT request fails (e.g., 503/403), the connection is pooled rather than abandoned.
2. All subsequent requests on this connection fail with HTTP 400 (INVALID_URL) because only the HTTP path is sent instead of a fully qualified URL (as required by the proxy).
3. Although the scheme is HTTPS, the payload is sent in cleartext after the failure.
4. Once this issue starts, it persists indefinitely.
5. Furthermore, as the proxy responds HTTP400 after receiving the client request immediately, the "bad" connections have higher chance of being selected for the requests.

# Expected Behavior

* The client should either abandon the connection or retry the CONNECT request to recover.
* Failed connections should not be reused in a way that leads to invalid HTTP requests.

# Actual Behavior

* The client neither abandons nor reattempts the CONNECT request.
* Instead, it continues sending requests in cleartext, leading to persistent failures.

This appears to be a bug in the error handling logic.

Thanks
Jason

PR #2072, opened 2025-03-09
fix: inappropriate connection reuse when using HTTP proxy if the initial CONNECT failed

jasonjoo2010:
# What This MR Resolves

A CONNECT request is needed to sent to the HTTP proxy first before the actual client request to establish the tunnel on the proxy. A `HTTP/1.1 200 Connection established` is expected for the initial CONNECT request. Only when the CONNECT is successful, the client continues sending the actual request through the "tunnel". And when CONNECT failed, the connection remains the initial state `unconnected`.  

There are following circumstances that a CONNECT fails under but not limited to following situations:  

- The destination is not whitelisted.
- The dest domain can't be resolved(timeout/SERVFAIL/NX/etc.).
- The dest IP can't be connected(timeout/unreachable/etc.).

There could be 2 following strategies to deal with CONNECT failures on the client side:  

1. Close the connection before return to the caller.
2. Mark this connection "unconnected" and put it into the pool. Then retry the CONNECT next time it's picked out of the pool.

The 2nd one needs to add extra state to Channel in the manager which brings bigger change to the code.
This MR employs the 1st strategy to resolve it. The issue is described in #2071 .

# Readings

The CONNECT is documented in `Section 5.3` in RFC2871: https://www.ietf.org/rfc/rfc2817.txt  

The proxy won't actively terminate the connection if the CONNECT failed if keep-alive is enabled. Unless the tunnel is established and there is any communication failures in the middle. Therefore the client needs to deal with this error by its own.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java
@@ -21,6 +21,7 @@ import io.netty.channel.ChannelHandler.Sharable;
 import io.netty.handler.codec.DecoderResultProvider;
 import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpHeaders;
+import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.HttpResponse;
 import io.netty.handler.codec.http.LastHttpContent;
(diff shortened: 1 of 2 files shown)


hyperxpro:
Thanks for the PR.

Can you please add an unit test around this?

jasonjoo2010:
Hi @hyperxpro 
Added a test for this scenario, pls take a look, thanks

hyperxpro (approved):

hyperxpro:
Thanks a lot!

jasonjoo2010:
Thanks for merging, this issue can be closed now https://github.com/AsyncHttpClient/async-http-client/issues/2071

May I back port this fix to 2.12.4? As 2.12.x is still commonly used

PR #2074, opened 2025-03-10
backport: fix inappropriate connection reuse when using HTTP proxy if the initial CONNECT failed

jasonjoo2010:
It's a backport fix(#2072) for #2071 on 2.12.x


hyperxpro:
Unfortunately, v2 is EoL and it'll not receive any updates.  But thanks a lot for fixing this in v3. 

:)

jasonjoo2010:
got it, I will notify internal users about this

PR #2088, opened 2025-05-07
Fix NPE race in NettyResponseFuture.cancel (#2042)

### LINKED ISSUE #2042: NPE in at org.asynchttpclient.netty.channel.Channels.setAttribute(Channels.java:36)
java.lang.NullPointerException
	at org.asynchttpclient.netty.channel.Channels.setAttribute(Channels.java:36)
	at org.asynchttpclient.netty.channel.Channels.setDiscard(Channels.java:40)
	at org.asynchttpclient.netty.NettyResponseFuture.cancel(NettyResponseFuture.java:183)

pratt4:
Fixes #2042 

This is a typical TOCTOU (time-of-check/time-of-use) race https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use. 
The NPE was occurring because the channel field could be set to null by another thread between the check and its use:

if (channel != null) {                 // time-of-check
    Channels.setDiscard(channel);      //  time-of-use
    Channels.silentlyCloseChannel(channel);
}

By copying channel into a local variable in one atomic read, we ensure that—even if another thread changes the field—the local reference remains valid.

P.S. It is hard to write a deterministic test that fails consistently, so this PR only includes the code fix.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
+++ b/client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
@@ -187,7 +187,7 @@ public final class NettyResponseFuture<V> implements ListenableFuture<V> {
             return false;
         }
 
-        Channel ch = channel; //atomic read, so that it won't end up in TOCTOU
+        final Channel ch = channel; //atomic read, so that it won't end up in TOCTOU
         if (ch != null) {
             Channels.setDiscard(ch);
             Channels.silentlyCloseChannel(ch);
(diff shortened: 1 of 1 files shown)


pratt4:
HI @hyperxpro ...
can you please review this pr whenever you get time?

client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
```diff
@@ -187,10 +187,10 @@ public boolean cancel(boolean force) {
             return false;
         }
 
-        // cancel could happen before channel was attached
-        if (channel != null) {
-            Channels.setDiscard(channel);
-            Channels.silentlyCloseChannel(channel);
+        Channel ch = channel; //atomic read, so that it won't end up in TOCTOU
```
hyperxpro, line 190: final

pratt4: Thank you for the review ...
updated the PR

hyperxpro (approved):

PR #2089, opened 2025-05-10
Implement HTTP Digest Access Authentication

### LINKED ISSUE #2068: any plan for RFC7616?
any plan for RFC7616?
https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/impl/auth/DigestScheme.java

hyperxpro:
RFC 7616 - HTTP Digest Access Authentication

Closes #2068

### DIFF
--- a/client/src/main/java/org/asynchttpclient/Realm.java
+++ b/client/src/main/java/org/asynchttpclient/Realm.java
@@ -35,6 +35,7 @@ import static org.asynchttpclient.util.MessageDigestUtils.pooledMd5MessageDigest
 import static org.asynchttpclient.util.MiscUtils.isNonEmpty;
 import static org.asynchttpclient.util.StringUtils.appendBase16;
 import static org.asynchttpclient.util.StringUtils.toHexString;
+import org.asynchttpclient.util.MessageDigestUtils;
 
 /**
  * This class is required when authentication is needed. The class support
@@ -452,6 +453,21 @@ public class Realm {
             return this;
         }
 
+        /**
+         * Extracts the value of a token from a WWW-Authenticate or Proxy-Authenticate header line.
+         * Example: match('Digest realm="test", nonce="abc"', "realm") returns "test"
+         */
+        private static @Nullable String match(String headerLine, String token) {
+            if (headerLine == null || token == null) return null;
+            String pattern = token + "=\"";
+            int start = headerLine.indexOf(pattern);
+            if (start == -1) return null;
+            start += pattern.length();
+            int end = headerLine.indexOf('"', start);
+            if (end == -1) return null;
+            return headerLine.substring(start, end);
+        }
+
         private void newCnonce(MessageDigest md) {
             byte[] b = new byte[8];
             ThreadLocalRandom.current().nextBytes(b);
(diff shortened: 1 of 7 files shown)


client/src/main/java/org/asynchttpclient/Realm.java
```diff
@@ -452,62 +453,65 @@ public Builder parseProxyAuthenticateHeader(String headerLine) {
...
+            start += pattern.length();
+            int end = headerLine.indexOf('"', start);
+            if (end == -1) return null;
+            return headerLine.substring(start, end);
+        }
+
         private void newCnonce(MessageDigest md) {
             byte[] b = new byte[8];
             ThreadLocalRandom.current().nextBytes(b);
             b = md.digest(b);
```
pratt4, line 474: MD5 =16 bytes; SHA-256 = 32 bytes;
rfc7616 doesn’t forbid long nonces... but wont the headers that big can be unwieldy, especially if you’re proxying or logging??

pratt4 (commented):
Please review the comments and let me know If the observations make sense...

If you'd like, I can pick up these fixes and push the changes (along with tests) on top of this PR to help wrap it up faster.
Let me know... happy to help!!!

client/src/main/java/org/asynchttpclient/Realm.java
```diff
@@ -452,62 +453,65 @@ public Builder parseProxyAuthenticateHeader(String headerLine) {
...
             sb.setLength(0);
             return md.digest();
         }
 
+        private static MessageDigest getDigestInstance(String algorithm) {
+            if (algorithm == null || "MD5".equalsIgnoreCase(algorithm) || "MD5-sess".equalsIgnoreCase(algorithm)) {
+                return MessageDigestUtils.pooledMd5MessageDigest();
+            } else if ("SHA-256".equalsIgnoreCase(algorithm) || "SHA-256-sess".equalsIgnoreCase(algorithm)) {
+                return MessageDigestUtils.pooledSha256MessageDigest();
+            } else if ("SHA-512-256".equalsIgnoreCase(algorithm) || "SHA-512-256-sess".equalsIgnoreCase(algorithm)) {
```
pratt4, line 489: will it handle "SHA-512/256" 
some server might send with / 's
and even "SHA-512/256" is mentioned in standard names docs... https://docs.oracle.com/en/java/javase/12/docs/specs/security/standard-names.html

client/src/main/java/org/asynchttpclient/Realm.java
```diff
@@ -452,62 +453,65 @@ public Builder parseProxyAuthenticateHeader(String headerLine) {
...
-            String value = headerLine.substring(match, trailingComa > 0 ? trailingComa : headerLine.length());
-            value = value.length() > 0 && value.charAt(value.length() - 1) == '"'
-                    ? value.substring(0, value.length() - 1)
-                    : value;
-            return value.charAt(0) == '"' ? value.substring(1) : value;
-        }
-
-        private static byte[] md5FromRecycledStringBuilder(StringBuilder sb, MessageDigest md) {
+        private static byte[] digestFromRecycledStringBuilder(StringBuilder sb, MessageDigest md) {
             md.update(StringUtils.charSequence2ByteBuffer(sb, ISO_8859_1));
```
pratt4, line 479: can utf-8 be implemented?

image

client/src/main/java/org/asynchttpclient/Realm.java
```diff
@@ -452,62 +453,65 @@ public Builder parseProxyAuthenticateHeader(String headerLine) {
...
             sb.setLength(0);
             return md.digest();
         }
 
+        private static MessageDigest getDigestInstance(String algorithm) {
```
pratt4, line 484: if the server sends multiple algos (eg: md5, sha256..eg)
then public Builder parseWWWAuthenticateHeader(String headerLine) --- can pick the first algo (here m5 can be chosen even if sha256 is present )

it should choose the algo based on strength right??
image

PR #2104, opened 2025-08-20
Future-proof HTTPS endpoint identification

chrisvest:
Netty 4.2 changes the default for hostname verification for TLS clients, so that it is now enabled by default.

As a result, clients that rely on the default being _off_ will find themselves unable to disable it.

Instead, clients should explicitly configure their desired endpoint identification algorithm in all cases.

Since Netty 4.1.112 we also have a convenient method on the `SslContextBuilder` for doing this, so we don't need multiple round-trips through `SSLParameters`.

This PR changes the `DefaultSslEngineFactory` to make use of this method, so it always configures the endpoint identification algorithm to match the desired setting of `AsyncHttpClientConfig..isDisableHttpsEndpointIdentificationAlgorithm()`.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/channel/ChannelPoolPartitioning.java
+++ b/client/src/main/java/org/asynchttpclient/channel/ChannelPoolPartitioning.java
@@ -15,12 +15,12 @@
  */
 package org.asynchttpclient.channel;
 
+import java.util.Objects;
+
 import org.asynchttpclient.proxy.ProxyServer;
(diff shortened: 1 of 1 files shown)


chrisvest:
FYI @slandelle @sullis @hyperxpro

hyperxpro (approved):

chrisvest:
@hyperxpro I don't have the merge button, so feel free to merge any time.

PR #2116, opened 2025-09-22
bump error_prone and nullaway

sullis:


hyperxpro:
error_prone has problems. I was wondering if it can be replaced with something better.

hyperxpro:
This is covered in this: https://github.com/AsyncHttpClient/async-http-client/pull/2149

PR #2140, opened 2026-02-08
Add BootstrapHandler inside `initChannel` for SOCKS

### LINKED ISSUE #2139: SOCKS proxy support broken
It appears that SOCKS proxy support has been silently broken somewhere in v3. As far as i can tell, none of the unit tests actually verify that traffic went through the proxy. I have an example where the request is expected to fail, and I'd expect to see a proxy connect error, which I am not seeing in v3.

I've created a repo to reproduce: https://github.com/jaredstehler/ahc-socks-proxy

Tested as fails in: `3.0.0...3.0.6`
Tested as succeeds in `2.12.3...2.12.4`

From debugging, it appears that there is an uncaught error in [`httpBootstrapHandler.handlerAdded(ctx)`](https://github.com/AsyncHttpClient/async-http-client/blob/ae59f51f70b2ad99601c0a0c23d8c6e9260a0400/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java#L513) which causes `super.handlerAdded(ctx)` (and subsequently, `initChannel()`) to not be invoked.

The actual error occurs at: 

```
java.util.NoSuchElementException: org.asynchttpclient.netty.channel.ChannelManager$1
	at io.netty.channel.DefaultChannelPipeline.getContextOrDie(DefaultChannelPipeline.java:1103)
	at io.netty.channel.DefaultChannelPipeline.remove(DefaultChannelPipeline.java:366)
	at io.netty.channel.ChannelInitializer.initChannel(ChannelInitializer.java:135)
	at io.netty.channel.ChannelInitializer.handlerAdded(ChannelInitializer.java:111)
	at org.asynchttpclient.netty.channel.ChannelManager$4.handlerAdded(ChannelManager.java:513)
```

hyperxpro:
Motivation:
SOCKS bootstrap pipeline wasn't wiring the HTTP handler correctly, causing proxy handshakes to fail or silently bypass proxies.

Modification:
- ChannelManager: Remove handlerAdded override and add httpBootstrapHandler directly into the SOCKS bootstrap pipeline.
- Tests:  Add findFreePort, replace flaky startup-based tests with deterministic assertions that requests fail when SOCKS (v4/v5) proxies are unreachable, and add timeouts.

Result:
Fixes #2139

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -30,7 +30,7 @@ jobs:
           distribution: 'corretto'
           java-version: '11'
       - name: Run Tests
-        run: ./mvnw -B -ntp test
+        run: ./mvnw -B -ntp test -Ddocker.tests=true
 
   RunOnMacOs:
     runs-on: macos-latest
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -27,8 +27,8 @@ jobs:
         with:
           distribution: 'corretto'
           java-version: '11'
-      - name: Run Tests
-        run: ./mvnw -B -ntp clean test
+      - name: Run Tests (force Docker tests)
+        run: ./mvnw -B -ntp clean test -Ddocker.tests=true
 
   RunOnMacOs:
     runs-on: macos-latest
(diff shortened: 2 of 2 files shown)


hyperxpro:
@jaredstehler PTAL

jaredstehler:
initial testing looks good! thanks for this quick fix!

PR #2163, opened 2026-04-10
Add pluggable DnsNameResolver

### LINKED ISSUE #1724: DNS lookups should not block the event loop
### The Problem
The default name resolver for AHC is:
```java
new DefaultNameResolver(ImmediateEventExecutor.INSTANCE)
```
This does DNS lookups by using the calling thread (a Netty event loop thread) to do a blocking DNS lookup via the OS mechanism. This is a bad default behavior.

### Proposed Solution
I propose that we move the blocking DNS lookups into a separate event loop. There does not appear to be a resolver in Netty that will do this, so we would likely need to create one.

Thoughts?

I am willing to do the development for this if we want this feature.

doom369:
Fixes https://github.com/AsyncHttpClient/async-http-client/issues/1724. Preserve the previous behaviour + ability to plug own DNS name resolver.

Found it yesterday on our prod server:

576462695-7db52b47-5d27-4cd1-a3e3-f399491a3648

576463467-537e2590-1bc8-424a-b5e2-d5370d3ac24a

### DIFF
--- a/client/src/test/java/org/asynchttpclient/AddressResolverGroupTest.java
+++ b/client/src/test/java/org/asynchttpclient/AddressResolverGroupTest.java
@@ -1,5 +1,5 @@
 /*
- *    Copyright (c) 2024 AsyncHttpClient Project. All rights reserved.
+ *    Copyright (c) 2026 AsyncHttpClient Project. All rights reserved.
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
  *    you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


client/src/test/java/org/asynchttpclient/AddressResolverGroupTest.java
```diff
@@ -0,0 +1,182 @@
+/*
+ *    Copyright (c) 2024 AsyncHttpClient Project. All rights reserved.
```
hyperxpro:
```suggestion
 *    Copyright (c) 2026 AsyncHttpClient Project. All rights reserved.
```

hyperxpro (changes requested):

hyperxpro (approved):

hyperxpro:
Thanks a lot for the great work! :)

PR #2170, opened 2026-05-01
Support MultiThreadIoEventLoopGroup for Netty 4.2+ compatibility

dduttaCodePractice:


dduttaCodePractice:
Added instanceof MultiThreadIoEventLoopGroup check in ChannelManager that selects transport based on availability (io_uring > epoll > kqueue > nio), since we cannot determine which IoHandler was used to create the group.

hyperxpro:
This is not correct implementation.

Handled it here: https://github.com/AsyncHttpClient/async-http-client/pull/2340

PR #2176, opened 2026-05-08
Guard against null TimeoutsHolder in NettyConnectListener.onSuccess

hyperxpro:
Motivation

NettyConnectListener.onSuccess NPEs on timeoutsHolder.setResolvedRemoteAddress(...) when a request timeout fires concurrently with a successful connect: abort() calls cancelTimeouts() (nulls the holder) before setting isDone, and per JMM the IO thread can observe holder==null while isDone==0. PR #2127 only guarded the remoteAddress parameter, not the holder.

Modification

- Add null guard on timeoutsHolder in onSuccess — close the channel and bail out if the holder was nulled out concurrently.
- Broaden futureIsAlreadyCancelled → futureIsAlreadyCompleted (isCancelled() → isDone()) so aborted futures don't slip past the early-out (abort()/done() set isDone, not isCancelled).
- Release the partition-key lock on both early-exit paths.
- Add NettyConnectListenerTest reproducing the exact NPE from the bug report.

Fixes
#2172

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/NettyConnectListener.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/NettyConnectListener.java
@@ -55,9 +55,11 @@ public final class NettyConnectListener<T> {
         this.connectionSemaphore = connectionSemaphore;
     }
 
-    private boolean futureIsAlreadyCancelled(Channel channel) {
-        // If Future is cancelled then we will close the channel silently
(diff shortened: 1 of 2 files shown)


hyperxpro:
@jaredstehler PTAL

jaredstehler:
This looks like it should do the trick! Thanks for the quick reply!

PR #2196, opened 2026-06-14
reject set-cookie domain that doesn't match the request host

jmestwa-coder:
ThreadSafeCookieStore stores a cookie under its Domain attribute without checking the response host is allowed to set it (rfc6265 5.3 step 6), so a host can plant cookies for unrelated domains:
- a response from www.evil.com with `Set-Cookie: SID=x; Domain=victim.com` is stored under victim.com
- the cookie is then sent on later requests to victim.com (cookie tossing)
- add() now drops the cookie when the request host does not domain-match the cookie domain

Added a CookieStoreTest case; the existing subdomain-match cases still pass.


hyperxpro (approved):

hyperxpro:
Thanks a lot!

PR #2202, opened 2026-06-16
Add LoadBalance.ROUND_ROBIN for per-request IP round-robin

pavel-ptashyts:
Motivation:

When a host resolves to multiple IP addresses, AsyncHttpClient currently pins almost all traffic to a single address. Although DNS resolution returns every A/AAAA record, the connector always attempts the first address first, using the remaining addresses only for failover. In addition, both the connection pool and the HTTP/2 connection registry are keyed only by `(scheme, host, port)`, never by the resolved IP address. As a result, with keep-alive enabled, the first reachable address receives almost all requests, making it impossible to distribute client traffic across multiple backend IPs.

Modification:

This change introduces an opt-in `LoadBalance` configuration with two modes: `DEFAULT` and `ROUND_ROBIN`. `DEFAULT` preserves the existing behavior unchanged.

`ROUND_ROBIN` rotates the resolved IP list on each eligible request so that a different address is tried first while preserving the remaining addresses as failover candidates. A per-host counter determines which IP is selected for each request.

Connection reuse is made IP-aware by including the resolved IP in the partition key, allowing both HTTP/1.1 pooled connections and HTTP/2 multiplexed connections to remain pinned to the selected address. If connection establishment falls back to another IP, the partition key is updated to reflect the address that was ultimately connected.

`maxConnectionsPerHost` continues to apply per hostname rather than per IP. The connection permit is acquired before the target IP is selected, so existing connection limit semantics remain unchanged.

To reduce repeated connection attempts to an unavailable address, failed connection attempts place that IP into a short 10-second cooldown. During this period the address is moved to the end of the rotation instead of being removed entirely, allowing it to remain as a last-resort failover target and to be retried automatically once the cooldown expires. This is intended only as a lightweight dampening mechanism rather than a health checker; determining endpoint health remains the responsibility of the DNS resolver, consistent with `DEFAULT` mode.

For HTTP/2, this intentionally creates one connection per backend IP so that streams are distributed across all resolved addresses. This is a deliberate deviation from RFC 9113 §9.1 and is documented in the `LoadBalance` enum.

One known limitation remains when `maxConnectionsPerHost` is configured lower than the number of resolved IPs. In that case, a request pinned to an IP that cannot obtain a connection permit will not reuse an existing HTTP/2 connection established for another IP. Since the default connection limit is unlimited, this does not affect most deployments. A complete solution, which would index the HTTP/2 registry by the base partition key, is tracked in issue https://github.com/AsyncHttpClient/async-http-client/issues/2214.

The entire feature is guarded by the new configuration flag and a per-request override that remains `null` in `DEFAULT` mode, leaving the existing send, connect, pool, and polling paths unchanged. `getLoadBalance()` is implemented as a default interface method returning `DEFAULT`, preserving both source and binary compatibility for existing configuration implementations.

Even when `ROUND_ROBIN` is enabled, the feature is a no-op for single-address hosts, literal IP addresses, requests with an explicitly configured remote address, and requests routed through HTTP or SOCKS proxies.

Result:

Requests are distributed evenly across all IP addresses associated with a multi-address host for both HTTP/1.1 and HTTP/2 while preserving TCP failover behavior and per-host connection limits. The default behavior remains unchanged.

The implementation is covered by unit tests validating round-robin rotation (including even distribution, deterministic ordering, single-IP no-op behavior, and independent per-host counters), cooldown behavior (deprioritization, automatic reprobing after expiry, fallback when all IPs are cooling, and no-op handling for unknown hosts), IP-aware partition key behavior (equality and re-pinning), and configuration parsing and builder round-tripping.

End-to-end tests using loopback addresses (`127.0.0.1`, `127.0.0.2`, and `127.0.0.3`) verify that `ROUND_ROBIN` distributes requests across all three addresses while `DEFAULT` continues to reuse a single connection, for both HTTP/1.1 and HTTP/2. The tests also validate failover handling and confirm that same-host redirects to a different port trigger DNS resolution again. All existing connection pool, connection limit, HTTP/2, retry, and redirect test suites continue to pass unchanged.


hyperxpro:
This is great - this is essentially client-side load balancing, Let me think through.

pavel-ptashyts:
Good afternoon. I would be very grateful if you could review my PR, and I will gladly try to improve this solution if you find any shortcomings or issues in it. Client-side load balancing is actually something I sometimes critically miss in the client, as solving the load balancing problem outside the client either works unreliably or requires configuring additional load balancers, which increases system complexity.

hyperxpro:
Agree - We will get this PR merged no worries. :)
Just give me a day or two to review as I am quite occupied at the moment.

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
```diff
@@ -0,0 +1,77 @@
+/*
+ *    Copyright (c) 2024 AsyncHttpClient Project. All rights reserved.
```
hyperxpro:
```suggestion
 *    Copyright (c) 2026 AsyncHttpClient Project. All rights reserved.
```

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
```diff
@@ -0,0 +1,77 @@
...
+public final class RoundRobinAddressSelector {
+
+    private static final Comparator<InetSocketAddress> STABLE_ORDER = Comparator.comparing(address -> {
+        if (address.getAddress() != null) {
+            return address.getAddress().getHostAddress();
+        }
+        return address.getHostString();
+    });
+
+    private final ConcurrentHashMap<String, AtomicInteger> counters = new ConcurrentHashMap<>();
```
hyperxpro: counters is keyed by host and never evicted - unbounded growth for a client that touches many distinct hosts (crawler/gateway). Entries are tiny, but it's a structure with no equivalent in DEFAULT mode. Please bound/LRU-cap it.

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * @param host     the request's target host
+     * @param resolved the resolved socket addresses (size {@code >= 1})
+     * @return the same list instance when there is nothing to rotate (size {@code <= 1}), otherwise
+     * a new list whose first element is the round-robin-selected address
+     */
+    public List<InetSocketAddress> rotate(String host, List<InetSocketAddress> resolved) {
```
hyperxpro, line 89: The selector just rotates through the full resolved list and never learns which IPs are actually healthy. So if one of a host's IPs is down for a while, a steady fraction of requests will keep picking it. Each of those misses the pool, since the working connection is parked under a different IP's key after the repin, opens a fresh connection to the dead IP, fails, falls over to a good IP, and repins there. That repeats for as long as the IP stays down, and the pool ends up lopsided around whichever live IP keeps absorbing the failovers.

Nothing breaks here: requests still succeed and the max-connections limit still holds, so there's no leak. But during a partial outage you're paying a wasted connection attempt, worst case a full connect timeout of latency, on a slice of traffic indefinitely. Probably we can do some kind of health check and it should be configurable; normal TCP healthcheck or HTTP healthcheck?

Also maybe we should be keeping a short list of recently-failed IPs and skipping them for a little while would do it - like a temporary blacklist.

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
```diff
@@ -0,0 +1,77 @@
...
+        int n = resolved.size();
+        if (n <= 1) {
+            return resolved;
+        }
+
+        List<InetSocketAddress> ordered = new ArrayList<>(resolved);
+        ordered.sort(STABLE_ORDER);
```
hyperxpro: This copies the list and re-sorts it on every call. The list is short so it doesn't matter much, but for a busy multi-IP host you're doing that allocation and sort on every request. If you wanted to avoid it, you could cache the sorted order for a given set of resolved addresses and reuse it.

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
```diff
@@ -0,0 +1,77 @@
...
+ * {@link org.asynchttpclient.RequestSendType#ROUND_ROBIN} is enabled.
+ *
+ * <p>{@link #rotate(String, List)} returns the resolved addresses re-ordered so that the
+ * round-robin-selected address comes first; the remaining addresses follow (in a stable order) so
+ * the connector can still fail over to them. The addresses are sorted into a stable order before
+ * rotation so that the per-host counter maps consistently to the same address across requests,
+ * regardless of the order the resolver returns them in.
+ *
+ * <p>Thread-safe.
+ */
```
hyperxpro, line 53: This needs to be aligned after the below comments are addressed.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -152,6 +172,69 @@ public <T> ListenableFuture<T> sendRequest(final Request request, final AsyncHan
...
 
+    // A request is eligible for round-robin only when it opens a direct connection to the target host
+    // (the connector targets the resolved IPs). Excluded: explicit address (bypasses resolution), and
+    // HTTP-proxied hosts (the proxy host is resolved, not the target).
+    private boolean isRoundRobinEligible(Request request, ProxyServer proxyServer) {
+        if (request.getAddress() != null || needConnect(request, proxyServer)) {
+            return false;
+        }
+        Uri uri = request.getUri();
+        return proxyServer == null || proxyServer.isIgnoredForHost(uri.getHost()) || !proxyServer.getProxyType().isHttp();
```
hyperxpro: This lets SOCKS proxies use round-robin too, which means we resolve the host ourselves, rotate the IPs, pin the pool per IP, and hand the chosen IP to the SOCKS server. With SOCKS5 people often expect the proxy to do the resolving and routing, so the combination is a bit odd. It isn't a regression, since we already resolve client-side for SOCKS today, but I'd either leave SOCKS out the way HTTP proxies are left out, or say plainly in the docs that round-robin resolves on the client even for SOCKS.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -152,6 +172,69 @@ public <T> ListenableFuture<T> sendRequest(final Request request, final AsyncHan
...
+     * connector targets it while keeping the others for failover), pin connection reuse to that IP via
+     * an IP-aware partition key, then run the normal reuse-or-connect logic.
+     */
+    private <T> ListenableFuture<T> sendRequestRoundRobin(Request request, AsyncHandler<T> asyncHandler, NettyResponseFuture<T> future,
+                                                          ProxyServer proxyServer) {
+        NettyResponseFuture<T> newFuture = newNettyRequestAndResponseFuture(request, asyncHandler, future, proxyServer, false);
+        Uri uri = request.getUri();
+        String host = uri.getHost();
+
+        resolveAddresses(request, proxyServer, newFuture, asyncHandler).addListener(new SimpleFutureListener<List<InetSocketAddress>>() {
```
hyperxpro, line 209: Worth a short comment here so the ordering is obvious to the next person: in this mode we resolve before checking the pool and before taking the semaphore, so every eligible request resolves first, even ones that immediately reuse a pooled connection and even single-IP hosts. 

On a pooled hit that also means the request timeout gets scheduled twice; the second call cancels the first so there's no leak, it's just a little redundant. It's all inherent to doing this per request and cheap with a caching resolver, but it's easy to miss reading the code.

client/src/main/java/org/asynchttpclient/config/AsyncHttpClientConfigDefaults.java
```diff
@@ -373,4 +376,16 @@ public static Duration defaultHttp2PingInterval() {
...
     }
+
+    public static RequestSendType defaultRequestSendType() {
+        String value = AsyncHttpClientConfigHelper.getAsyncHttpClientConfig().getString(ASYNC_CLIENT_CONFIG_ROOT + REQUEST_SEND_TYPE_CONFIG);
+        if (value == null || value.trim().isEmpty()) {
+            return RequestSendType.DEFAULT;
+        }
+        try {
+            return RequestSendType.valueOf(value.trim().toUpperCase(Locale.ROOT));
+        } catch (IllegalArgumentException e) {
```
hyperxpro, line 392: Quietly falling back to DEFAULT when the value doesn't parse will hide typos in someone's config. A warning log that includes the bad value would make that a lot easier to spot.

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinPartitionKey.java
```diff
@@ -0,0 +1,68 @@
+/*
+ *    Copyright (c) 2024 AsyncHttpClient Project. All rights reserved.
```
hyperxpro:
```suggestion
 *    Copyright (c) 2026 AsyncHttpClient Project. All rights reserved.
```

client/src/main/java/org/asynchttpclient/RequestSendType.java
```diff
@@ -0,0 +1,51 @@
+/*
+ *    Copyright (c) 2024 AsyncHttpClient Project. All rights reserved.
```
hyperxpro:
```suggestion
 *    Copyright (c) 2026 AsyncHttpClient Project. All rights reserved.
```

hyperxpro (changes requested):
Round 1 done - good work indeed!

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
pavel-ptashyts: Fixed

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinPartitionKey.java
pavel-ptashyts: Fixed

client/src/main/java/org/asynchttpclient/RequestSendType.java
pavel-ptashyts: Fixed

client/src/main/java/org/asynchttpclient/config/AsyncHttpClientConfigDefaults.java
pavel-ptashyts: Fixed

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Fixed. I left round robin logic only for direct connect without proxy. If need in the future this logic can be updated and extended

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
pavel-ptashyts: Thanks for the detailed write-up — agreed on the mechanics. During a partial outage, round-robin will keep handing a temporarily-dead IP its share of traffic, paying a connect failure (worst case a full connect-timeout) before failing over to a healthy IP. As you said, nothing breaks: requests still succeed and the connection limits hold — it's wasted latency on a slice of traffic while the IP is down.

My take is that liveness of resolved IPs is fundamentally a DNS/resolver concern, not something the round-robin selector should own. The current behavior is consistent with DEFAULT mode: if DNS returns a dead IP, AHC doesn't health-check it there either — it just tries it and fails over. Round-robin doesn't change that in kind, only in frequency (it touches every IP rather than pinning the first). The cleanest place to keep dead IPs out of rotation is the resolver/DNS layer (short TTLs, the LB pulling unhealthy backends, or a health-aware resolver) so liveness logic lives in one place instead of being reimplemented inside the client.

For this PR I've documented the limitation on RequestSendType.ROUND_ROBIN. If you'd like client-side handling, I see two increments we could do (here or as a follow-up):

1. Temporary failed-IP blacklist — on a failed connect attempt, mark that IP as "cooling down" for a short, configurable window and have the selector skip it until the window expires. Cheap, self-contained, no probing, recovers automatically.
2. Active health checks — configurable TCP or HTTP probes per resolved IP with intervals/thresholds, removing unhealthy IPs from rotation proactively. More powerful, but a much larger feature (new config surface, a prober lifecycle, failure semantics).

My preference is to defer both to a follow-up issue and let DNS own liveness for now, but I'm happy to implement (1) in this PR if you want it included. Do you want either of these as part of this PR, or is documenting the limitation enough for now?

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
pavel-ptashyts: For full LRU support, it would be a good idea to integrate the Caffeine library. However, I don’t want to do that without your approval.

So for now, I implemented an approach where the maximum possible size is 4096, and if an overflow occurs, 10% of the oldest data is cleared.

I think that in 99% of cases, the cleanup will never even be triggered, since 4096 is a large value. And even if someone does need more, the delay caused by cleanup in a single request should not significantly affect the overall performance of the library.

Should we keep my solution, or would it be better to use the Caffeine library?

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
pavel-ptashyts: Good point on the per-request allocation + sort. Rather than cache the sorted order, I've removed the client-side sort entirely and now rotate over the resolver's address list as-is. The ordering of a host's IPs is really the resolver's responsibility, so baking a sort into the round-robin selector was the wrong layer — it both duplicated work on every request and hid the fact that consistent rotation depends on a consistent resolution order.

I've updated the RequestSendType.ROUND_ROBIN javadoc to make the contract explicit:
- This mode takes the address order straight from the configured InetNameResolver and does not re-sort it, so for the rotation to map consistently across requests the resolver must return addresses in a stable order and not reorder them between resolutions — e.g. DnsNameResolver.
- It should not be paired with a resolver that intentionally rotates its results, such as RoundRobinInetAddressResolver — that one is meant for DEFAULT mode, where the client always targets the first address and the resolver provides the spreading instead.

Net effect: less per-request work (no copy+sort; we only allocate a rotated list when the selected index isn't already first), and the responsibility for IP ordering sits in the resolver where it belongs.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Good catch on both the ordering and the double timeout-schedule. I've added a comment at the resolveAddresses(...) call in sendRequestRoundRobin spelling out the non-obvious bits:
- Round-robin resolves up front — before the pool check and before the per-host semaphore — so every eligible request resolves first, including pooled-hit reuse and single-IP hosts. With a caching resolver that's cheap, but it's easy to miss when reading the code.
- On a pooled hit the request timeout is scheduled twice — once in resolveAddresses and again in sendRequestWithOpenChannel — and the second setTimeoutsHolder cancels the first, so it's redundant work, not a leak.

I kept it to a comment for now rather than suppressing the redundant schedule: scheduleRequestTimeout is on the shared resolve path used by every send mode, the extra schedule is harmless, and special-casing it risks a regression for negligible gain. That said — if you have ideas on how to do this better (e.g. a clean way to resolve-then-pool without the redundant schedule, or restructuring the ordering), I'm all ears.

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
pavel-ptashyts: Done

pavel-ptashyts (commented):
@hyperxpro could you review PR again. Thanks in advance

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -136,6 +141,21 @@ public <T> ListenableFuture<T> sendRequest(final Request request, final AsyncHan
         validateWebSocketRequest(request, asyncHandler);
         ProxyServer proxyServer = getProxyServer(config, request);
 
+        // Round-robin across the host's resolved IPs: resolve first, pick the next IP, then proceed.
+        // Re-evaluated when the target host changes (e.g. a cross-host redirect); same-host retries keep
+        // their pick. See RequestSendType.ROUND_ROBIN.
+        if (config.getRequestSendType() == RequestSendType.ROUND_ROBIN) {
+            boolean overrideMatchesHost = future != null && request.getUri().getHost().equals(future.getRoundRobinHost());
```
hyperxpro: This match is host only, but everything it gates is scheme and port specific. `roundRobinAddresses` has the port baked in from `getExplicitPort()`, and the override key is built from the scheme, host, and port base. So a same host redirect that changes scheme or port, such as the ordinary HTTP to HTTPS upgrade, still sees `overrideMatchesHost == true`, which skips both the re resolve and the `clearRoundRobinOverrides()` below.

The reused future then carries the cached port 80 addresses and the `RoundRobinPartitionKey` built from the HTTP key into the HTTPS request. `sendRequestWithNewChannel` reuses those addresses as is at line 446, so we dial port 80 and start a TLS handshake against the plaintext port. The request hangs or fails. If the just drained keep alive channel is still pooled under the stale HTTP key, the poll hits it and the HTTPS request goes out in cleartext instead. Single IP hosts are affected too, since every eligible request caches `roundRobinAddresses` and `roundRobinHost`.

I'd gate on the full base identity rather than just the host. Store the base `Uri` and compare it with `isSameBase`, or simply re run `sendRequestRoundRobin` whenever the redirected request is not same base. The same condition should also drive the clear so stale state cannot leak across a scheme or port change.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -986,15 +1090,17 @@ private static void validateWebSocketRequest(Request request, AsyncHandler<?> as
...
-    private Channel waitForHttp2Connection(Request request, ProxyServer proxy) {
+    private Channel waitForHttp2Connection(Request request, ProxyServer proxy, NettyResponseFuture<?> future) {
         Uri uri = request.getUri();
         // WebSocket requests must never multiplex onto an HTTP/2 connection (no RFC 8441 support). See #2160.
         if (uri.isWebSocket()) {
             return null;
         }
         String virtualHost = request.getVirtualHost();
+        // In round-robin mode, only multiplex onto the H2 connection for the IP this request is pinned to.
+        Object override = future != null ? future.getPartitionKeyOverride() : null;
```
hyperxpro, line 1125: The H2 fallback poll is keyed only by the per IP override, but the semaphore is per host. With HTTP/2 and a finite `maxConnectionsPerHost` smaller than the resolved IP count, this can stall. For example, if the cap is 1 and the host resolves to two IPs, request A takes the permit and registers its connection under the round robin key for IP A. Request B, pinned to IP B, cannot acquire a permit and polls the registry only for the round robin key for IP B, which never appears. B spins until `connectTimeout` and throws `TooManyConnectionsException`, whereas the default mode would have found A's connection on the per host key and multiplexed immediately.

The default `maxConnectionsPerHost` is unlimited, so most users will not encounter this. However, for anyone who sets a finite per host limit, this changes what would normally be graceful HTTP/2 multiplexing into a timeout followed by failure. At minimum this is worth documenting as a limitation. Alternatively, when the per IP poll misses under permit exhaustion, the implementation could fall back to the per host HTTP/2 poll.

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
```diff
@@ -0,0 +1,121 @@
...
+            evictOldest();
+        }
+        return value;
+    }
+
+    // Evict the least-recently-used hosts (by last-access time) down to LOW_WATER_MARK. Only one
+    // thread sweeps at a time; the rest skip and proceed, so the map may briefly exceed
+    // MAX_TRACKED_HOSTS. Dropping a host's counter is harmless: its rotation simply restarts at the
+    // first address the next time the host is seen.
+    private void evictOldest() {
```
hyperxpro: This is the cap I asked for, so I'd treat this as a suggestion rather than a must fix. The repository already has the same bounded counter pattern in `util/NonceCounter`: a `ConcurrentHashMap<String, Counter>` with arbitrary eviction on overflow and no timestamps. Here we instead have an `AtomicBoolean` sweep guard, a full `entrySet()` snapshot, an `O(n log n)` sort by `lastAccessNanos`, and a `System.nanoTime()` volatile write on every request. That sort runs on the resolver or event loop thread in exactly the high host churn case the cap is intended to handle.

The comment already notes that dropping a counter is harmless, so exact LRU seems like more machinery than this really needs. Matching `NonceCounter`'s arbitrary eviction would simplify the implementation and keep the hot path free of the per request `nanoTime()` update. That said, I'm also fine keeping the current LRU approach if you prefer it.

client/src/main/java/org/asynchttpclient/RequestSendType.java
```diff
@@ -0,0 +1,69 @@
...
+ *    limitations under the License.
+ */
+package org.asynchttpclient;
+
+/**
+ * Controls how requests are dispatched to a host that resolves to several IP addresses.
+ *
+ * <p>Configured globally through {@link AsyncHttpClientConfig#getRequestSendType()}.
+ */
+public enum RequestSendType {
```
hyperxpro: nit: Let's call this `LoadBalance` - so it's more clear and concise.

client/src/main/java/org/asynchttpclient/LoadBalance.java
```diff
@@ -0,0 +1,69 @@
...
+     * host. The net effect is that consecutive requests to a multi-IP host are spread evenly across
+     * all of its addresses, even when connections are kept alive.
+     *
+     * <p>Notes:
+     * <ul>
+     *   <li>Has no effect for hosts that resolve to a single address, literal-IP hosts, requests
+     *       with an explicit {@link Request#getAddress() address}, or requests routed through a
+     *       proxy (HTTP or SOCKS) — the socket is established to the proxy, not directly to the
+     *       rotated target IPs. (Round-robin still applies when the proxy is bypassed for the host.)</li>
+     *   <li>Connection limits ({@code maxConnectionsPerHost}) remain per host, not per IP.</li>
```
hyperxpro, line 52: One thing worth stating outright in these notes: in HTTP/2, `ROUND_ROBIN` deliberately opens more than one connection to the same host and port, one per resolved IP, so streams can be spread across the host's backends. That goes against the recommendation in RFC 9113 (and RFC 7540) Section 9.1 that clients **SHOULD NOT** open more than one HTTP/2 connection to a given host and port pair. Since this is a **SHOULD**, not a **MUST**, and distributing streams across multiple backend endpoints is a legitimate reason to deviate, this is not a protocol conformance issue. However, it is a conscious tradeoff that differs from the default behavior, so it is worth documenting to avoid surprising users who enable this mode expecting HTTP/2's usual single connection per origin.


```suggestion
     *   <li>Connection limits ({@code maxConnectionsPerHost}) remain per host, not per IP.</li>
     * <li>For HTTP/2, this strategy deliberately opens more than one connection to the same host and port, one per resolved IP, so streams can be spread across the host's backends. This is a conscious deviation from the recommendation in RFC 9113 (and RFC 7540) Section 9.1 that clients SHOULD NOT open more than one HTTP/2 connection to a given host and port pair. {@link #DEFAULT} mode preserves the standard single connection per origin behavior.</li>
```

hyperxpro (changes requested):
Looks good overall

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Fixed

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Add information to javadocs

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
pavel-ptashyts: Updated code for using aproache as in NonceCounter

client/src/main/java/org/asynchttpclient/RequestSendType.java
pavel-ptashyts: Renamed everywhere

client/src/main/java/org/asynchttpclient/LoadBalance.java
pavel-ptashyts: Done

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
hyperxpro: The doc captures the failure mode well. One correction on the fallback I floated though, it doesn't work as written, and I'd rather flag that than have someone implement a no-op.

"Fall back to the per-host H2 poll" returns null here. In this mode the connection is registered under the future's partition key, which is the per-IP round-robin key, not the base key (NettyConnectListener calls registerHttp2Connection(future.getPartitionKey(), ...)). The registry is an exact-key map, so request A's connection only exists under RoundRobinPartitionKey(base, IP_A). Polling the base per-host key finds nothing. So that swap would compile and change nothing.

For B to actually reuse A's connection it has to find any open H2 connection for the host across its per-IP keys, which means indexing the registry by base key or scanning for a sibling round-robin key. That's a real change, not a one-line poll swap.

I'm OK leaving the doc as the stopgap for this PR since the default cap is unlimited. Can we open a follow-up for "reuse a sibling-IP H2 connection when the per-IP one is permit-starved" and call out in it that it needs base-key indexing, so it doesn't get picked up as a trivial fix? Worth noting there too that off the event loop B also spins the full connectTimeout in waitForHttp2Connection before throwing, so today it's a stall and then a failure.

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
hyperxpro: Agree; Liveness really does belong in the resolver/DNS layer, and round-robin matching DEFAULT's try-and-fail-over behavior is consistent, just at higher frequency. I'm good deferring active health checks to a follow-up.

Two small things first. The Javadoc note doesn't actually state the cost, it says requests are "retried via TCP failover to a healthy IP" but not that a black-holed IP (packets dropped, no RST) burns a full connectTimeout per pinned request before that failover. You say it in your reply here, the doc should say it too so someone reading only the Javadoc knows what they're enabling.

And on increment 1, the temporary failed-IP cooldown, I'd take that if you're up for it. It's cheap and self-contained and it's the difference between wasting a connect timeout on a slice of traffic indefinitely versus once and then routing around. If you'd rather keep this PR tight, let's at least file the follow-up and link it here. Either way's fine by me.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Updated javadoc and also created issue https://github.com/AsyncHttpClient/async-http-client/issues/2214. If need I can take this issue after merging this PR

client/src/main/java/org/asynchttpclient/netty/channel/RoundRobinAddressSelector.java
pavel-ptashyts: Agree. So at current moment I add cooldown  for host which was failed connection and updated javadocs

hyperxpro (approved):

pavel-ptashyts:
Thanks for approving the PR.

It looks like the GitHub Actions workflow hasn't started yet. Could you please also approve/run the workflow for this PR?

pavel-ptashyts:
Hi @hyperxpro  Thanks. All tests are passed. But I don't have permission to merge PR

hyperxpro:
Thanks a lot for the great work!

pavel-ptashyts:
Hi @hyperxpro thanks you too. Please note that I already did fix for isssue 2214. https://github.com/AsyncHttpClient/async-http-client/pull/2217. And I also have some ideas for improvements

PR #2211, opened 2026-06-23
fail websocket handshake on invalid sec-websocket-accept

jmestwa-coder:
**WebSocket upgrade continues after a failed challenge check**
When the server's Sec-WebSocket-Accept doesn't match the key (or the header is absent), `upgrade()` aborts the future but does not return, so it goes on to swap in the WebSocket pipeline and call `onOpen` on the listener. RFC 6455 section 4.1 requires the client to fail the connection here, which is what the sibling handshake check right above already does. Added the missing return so a bad or missing accept value ends the connection instead of surfacing an unverified peer as an open WebSocket.

Covered by a new test that drives the client against a server returning a 101 with a wrong Sec-WebSocket-Accept and asserts onOpen never fires.


hyperxpro (approved):

hyperxpro:
Thanks a lot!

PR #2217, opened 2026-06-29
Round robin issue (#2214) 

### LINKED ISSUE #2214: Round-robin: reuse a sibling-IP HTTP/2 connection when the per-IP connection is permit-starved
### Context

Follow-up from #2202 (`LoadBalance.ROUND_ROBIN`), raised in [this review comment](https://github.com/AsyncHttpClient/async-http-client/pull/2202#discussion_r3486891276).

In round-robin mode each request is pinned to a resolved IP via `RoundRobinPartitionKey(base, IP)`, but the connection permit is still per host (`maxConnectionsPerHost`, keyed by the base host key). These two facts interact badly under a configured connection cap.

### Problem

- Request **A** opens an HTTP/2 connection to `IP_A`. It is registered in the H2 registry under its per-IP key `(base, IP_A)` (`NettyConnectListener` calls `registerHttp2Connection(future.getPartitionKey(), …)`, and `getPartitionKey()` returns the per-IP override).
- Request **B** is pinned to `IP_B`. The host is already at `maxConnectionsPerHost`, so `acquirePartitionLockLazily()` fails and B enters `waitForHttp2Connection`.
- B polls the registry with its own key `(base, IP_B)` and finds nothing — A's connection lives only under `(base, IP_A)`.
- B can neither open a new connection (host permit exhausted) nor reuse A's sibling connection. Off the event loop it spins for the **full `connectTimeout`** in `waitForHttp2Connection` and then fails with the original permit exception — a stall followed by a failure.

This only bites when `maxConnectionsPerHost` is configured; the default is unlimited, so most users never hit it. That is why #2202 ships an accurate doc as the stopgap rather than a behavioral change.

### Why the obvious fix does not work

Falling back to a poll on the per-host **base** key is a **no-op**: the H2 registry (`ChannelManager.http2Connections`) is an exact-key `ConcurrentHashMap`, and nothing is ever registered under the base key in round-robin mode. A one-line "poll the base key instead" swap would compile and change nothing.

### Real fix

For B to reuse A's connection, the registry must let a request find **any** open H2 connection for the host across its per-IP keys — i.e. index the H2 registry by base key (or scan sibling round-robin keys for the same base). This is a genuine change to registry indexing, **not** a trivial poll swap, and should be scoped/reviewed as such.

### Notes

- The stall-then-fail on `connectTimeout` (off the event loop) is worth covering in any fix or test.
- Active health checks / failed-IP handling are tracked separately.

pavel-ptashyts:
Fixes https://github.com/AsyncHttpClient/async-http-client/issues/2214.
Two HTTP/2 robustness fixes for `LoadBalance.ROUND_ROBIN`, on top of the feature
branch (`requests-round-robin-per-host`):

1. A permit-starved request now reuses a sibling-IP HTTP/2 connection instead of
   stalling for `connectTimeout` and then failing (#2214).
2. `maxConnectionsPerHost` is now actually *enforced* as a per-host connection cap
   in round-robin mode (it previously only throttled connection establishment).

Two commits:
- `Fix #2214: reuse a sibling-IP HTTP/2 connection when permit-starved`
- `Enforce maxConnectionsPerHost for round-robin HTTP/2`

## 1. Reuse a sibling-IP HTTP/2 connection when permit-starved (#2214)

In round-robin mode each request is pinned to a resolved IP via
`RoundRobinPartitionKey(base, IP)`, but the connection permit is per host. With a
finite `maxConnectionsPerHost`, a request pinned to `IP_B` that couldn't acquire a
permit polled only its own per-IP key in the HTTP/2 registry, never found a
connection already open on `IP_A`, and — off the event loop — spun the full
`connectTimeout` before failing with the permit exception.

- The HTTP/2 registry (`ChannelManager.http2Connections`) changes from a flat
  exact-key map into a **base-keyed nested map** (`base host → per-IP key →
  channel`): single source of truth, O(1) exact lookup, sibling lookup scoped to
  the host. Register runs inside `compute(baseKey, …)` and removal inside
  `computeIfPresent(baseKey, …)` on the same outer bucket, so a concurrent
  insert/prune can't orphan an entry.
- On the permit-starved path, `waitForHttp2Connection` falls back to
  `pollHttp2SiblingConnection(baseKey)` — any active, non-draining connection to
  the same host on a sibling IP — and multiplexes onto it. Multiplexing onto an
  existing HTTP/2 connection takes no permit, so this is safe. The fallback is
  confined to the permit-starved path; the happy path stays strictly per IP, so
  round-robin spreading is unaffected.

## 2. Enforce `maxConnectionsPerHost` for round-robin HTTP/2

The per-host permit was released immediately after ALPN (as in `DEFAULT` mode), so
it only throttled connection *establishment*. Because each request pins to a
different IP and opens its own connection, a host could accumulate up to one
connection per resolved IP regardless of the cap — it wasn't enforced.

- In round-robin mode the permit is now held for the HTTP/2 connection's lifetime
  (as HTTP/1.1 already does). The per-host semaphore then bounds live connections:
  at most `maxConnectionsPerHost` (K) connections, spread across K of the IPs; a
  request pinned to a further IP fails to acquire a permit and multiplexes onto a
  sibling (via the fix above) rather than opening another. `DEFAULT` mode is
  unchanged (one connection per host, released post-ALPN so concurrent requests
  multiplex onto it).

## Behavior & compatibility

- `DEFAULT` mode is unchanged.
- `maxConnectionsPerHost` now caps live HTTP/2 connections per host in round-robin
  mode. When it's `< #resolved IPs`, the cap wins over full spreading: K
  connections are held open (kept alive by PING) and stay pinned to K IPs. Set it
  `≥ #IPs` for full per-IP spreading. The default is unlimited, so this only
  affects clients that configure a finite per-host cap.
- A live round-robin HTTP/2 connection now also occupies a global `maxConnections`
  slot for its lifetime.

## Testing

- `ChannelManagerHttp2SiblingLookupTest` (new) — nested registry + sibling lookup:
  active hit, draining/dead/closed skipped, base mismatch ignored, empty-inner-map
  pruning.
- `RoundRobinPartitionKeyTest` — `getBaseKey` coverage.
- `BasicHttp2Test`:
  - permit-starved sibling reuse (concurrent burst, `maxConnectionsPerHost=1`,
    multi-IP) — completes by multiplexing instead of stalling;
  - spreading regression guard (cap ≥ #IPs → every IP still attempted);
  - cap enforcement (3 IPs, cap=2 → ≤ 2 live connections, all requests succeed).
- Existing HTTP/2 multiplex / orphan / conformance regression suites and
  `RoundRobinSendTypeTest` pass; the two reflective `http2Connections()` test
  helpers were updated to flatten the nested map.

Both fixes were adversarially verified: reverting the change makes the sibling
reuse test stall-then-fail, and the cap test report `expected <= 2, got 3`.

## Out of scope

- Active health checks / authoritative failed-IP handling remain the resolver/DNS
  layer's responsibility (round-robin's IP cooldown is only a short-lived dampener).

### DIFF
--- a/client/src/main/java/org/asynchttpclient/LoadBalance.java
+++ b/client/src/main/java/org/asynchttpclient/LoadBalance.java
@@ -49,20 +49,25 @@ public enum LoadBalance {
      *       with an explicit {@link Request#getAddress() address}, or requests routed through a
      *       proxy (HTTP or SOCKS) — the socket is established to the proxy, not directly to the
      *       rotated target IPs. (Round-robin still applies when the proxy is bypassed for the host.)</li>
-     *   <li>Connection limits ({@code maxConnectionsPerHost}) remain per host, not per IP.</li>
-     *   <li>For HTTP/2 this mode deliberately opens more than one connection to the same host and port
-     *       — one per resolved IP — so streams can be spread across the host's backends. That is a
-     *       conscious deviation from RFC 9113 (and RFC 7540) Section 9.1, which recommends that clients
-     *       SHOULD NOT open more than one HTTP/2 connection to a given host and port pair.
-     *       {@link #DEFAULT} mode preserves the standard single-connection-per-origin behavior.</li>
-     *   <li>With HTTP/2 and a finite {@code maxConnectionsPerHost} smaller than the number of resolved
-     *       IPs, a request pinned to an IP that cannot acquire a per-host connection permit does not open
-     *       a new connection; instead it multiplexes onto an active HTTP/2 connection already open to a
-     *       sibling IP of the same host, if one exists (multiplexing onto an existing connection takes no
-     *       permit). Only if no such connection exists does it fail with
-     *       {@code TooManyConnectionsPerHostException}. The default {@code maxConnectionsPerHost} is
-     *       unlimited, so this only affects clients that set a finite per-host limit below the
-     *       resolved-IP count.</li>
+     *   <li>Connection limits ({@code maxConnectionsPerHost}) are enforced per host, not per IP: at most
+     *       that many HTTP/2 connections are kept open to a host at once, spread across that many of its
+     *       resolved IPs (see the cap note below).</li>
+     *   <li>For HTTP/2 this mode deliberately opens more than one connection to the same host and port —
+     *       up to one per resolved IP, bounded by {@code maxConnectionsPerHost} — so streams can be spread
+     *       across the host's backends. That is a conscious deviation from RFC 9113 (and RFC 7540)
+     *       Section 9.1, which recommends that clients SHOULD NOT open more than one HTTP/2 connection to a
+     *       given host and port pair. {@link #DEFAULT} mode preserves the standard
+     *       single-connection-per-origin behavior.</li>
+     *   <li>When {@code maxConnectionsPerHost} (call it K) is smaller than the number of resolved IPs, the
+     *       cap wins over full spreading: at most K HTTP/2 connections are opened, to the first K rotated
+     *       IPs, and a request pinned to any further IP cannot acquire a per-host permit — it multiplexes
+     *       onto one of those K sibling connections (multiplexing onto an existing connection takes no
+     *       permit) instead of opening another, failing with {@code TooManyConnectionsPerHostException}
+     *       only if none is available. Those K connections are held open (kept alive by HTTP/2 PING), so
+     *       they stay pinned to their K IPs rather than rotating across all of them; set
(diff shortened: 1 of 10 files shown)


pavel-ptashyts:
Hi @hyperxpro could you check this PR. Thanks in advance

hyperxpro:
Please rebase this PR with main - the commit history is messed up.

pavel-ptashyts:
@hyperxpro done

hyperxpro (approved):

PR #2219, opened 2026-06-30
Avoid per-connect lock on shared Bootstrap options map (#2218)

### LINKED ISSUE #2218: AHC: per-connect lock contention on the shared Bootstrap options map
# AHC: per-connect lock contention on the shared Bootstrap options map

## Summary

`async-http-client` configures channel options on its shared Netty `Bootstrap`
via `Bootstrap.option(...)`. On **every** connection attempt, Netty's
`AbstractBootstrap.newOptionsArray()` takes a lock on that options map and copies
it into a fresh array. Because AHC reuses a single shared `Bootstrap` for all
outbound connections, all connects serialize on this one monitor. Under a high
rate of connection establishment this becomes a contention hotspot, even though
the options never change after startup.

Affected versions: AHC 3.x (and the equivalent code path in 2.x) on Netty 4.1.x / 4.2.x.

## Root cause

AHC builds the bootstrap once, in `ChannelManager.newBootstrap(...)`, and sets
the channel options on it:

```java
new Bootstrap()
    .channelFactory(factory)
    .group(eventLoopGroup)
    .option(ChannelOption.ALLOCATOR, ...)
    .option(ChannelOption.TCP_NODELAY, ...)
    .option(ChannelOption.SO_REUSEADDR, ...)
    .option(ChannelOption.SO_KEEPALIVE, ...)
    .option(ChannelOption.AUTO_CLOSE, ...)
    .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, ...)
    .option(ChannelOption.SO_LINGER, ...)
    .option(ChannelOption.SO_SNDBUF, ...)
    .option(ChannelOption.SO_RCVBUF, ...)
    // + AUTO_READ and any user-supplied config.getChannelOptions()
```

These options are stored in the bootstrap's internal `options` map and are never
mutated afterwards.

On each connect, Netty's `Bootstrap.init(Channel)` calls
`AbstractBootstrap.newOptionsArray()`, which is implemented (Netty 4.1/4.2) as:

```java
static Map.Entry<ChannelOption<?>, Object>[] newOptionsArray(
        Map<ChannelOption<?>, Object> options) {
    synchronized (options) {
        return new ArrayList<Map.Entry<ChannelOption<?>, Object>>(options.entrySet())
                .toArray(EMPTY_OPTION_ARRAY);
    }
}
```

So every connection establishment:

1. Acquires the monitor on the shared `options` map (`synchronized (options)`).
2. Allocates a fresh copy of the map and converts it to an array.

Since the same `Bootstrap` instance (and therefore the same `options` map) is
used for all outbound connections, step 1 makes the connect path a single global
serialization point, and step 2 adds per-connect allocations. The work done
under the lock is recomputed identically every time, despite the options being
immutable after setup.

## When it manifests

The contention only becomes visible under a **high rate of new connection
establishment**, for example:

- High outbound request concurrency where many threads initiate connects at the
  same time.
- High connection churn / low connection reuse: short-lived connections, request
  timeouts that close in-flight connections, servers responding with
  `Connection: close`, or a large number of distinct destination hosts so the
  pool rarely has an idle connection to reuse.
- Configurations without an effective per-host connection cap, where bursts open
  many new connections concurrently instead of reusing pooled ones.

When connections are reused effectively (steady-state pools, few connects), the
lock is cold and the issue does not appear. The problem is proportional to the
connect rate and to the number of threads concurrently entering the connect path.

## Effect

- **Lock contention on the connect path.** All connection attempts serialize on
  one monitor. As the number of threads concurrently connecting grows, this
  develops into a lock convoy: the achievable handoff throughput of the single
  monitor falls below the connect demand, a backlog forms, and the time spent
  blocked waiting to enter the monitor grows. Under heavy contention the blocking
  time per connect can become large (well beyond the cost of the trivial
  critical section itself), because the bottleneck is monitor handoff /
  park-unpark overhead rather than the work under the lock.
- **CPU underutilization at saturation.** Because contending threads are parked
  waiting for the monitor rather than running, the system can be throughput-bound
  on this lock while CPU is far from saturated.
- **Extra allocation pressure.** A new `LinkedHashMap`/`ArrayList` copy plus an
  array is allocated on every connect, adding avoidable GC churn proportional to
  the connect rate.

The critical section itself is tiny (copying a handful of immutable options), so
none of this reflects real work — it is pure overhead from recomputing an
immutable value under a shared lock on every connect.

## Proposed fix (Solution A): apply channel options in the ChannelInitializer instead of on the Bootstrap

Stop putting the channel options into the bootstrap's `options` map. Compute them
once at startup, and apply them to each channel in AHC's channel initializer
using the public `Channel.config().setOption(...)` API.

With the bootstrap's `options` map empty, `newOptionsArray()` still runs but
locks on an empty map and copies nothing — the lock hold time drops to
near-zero, so the convoy cannot build, and the per-connect map/array allocation
disappears. This applies the exact same options at the same point in the channel
lifecycle that Netty's `init()` would (during channel registration, before
connect), so behavior is unchanged.

Sketch (in `ChannelManager`):

```java
// Computed once at startup — same options that used to be set via bootstrap.option(...)
private final Map<ChannelOption<?>, Object> channelOptions = buildChannelOptions(config);

private Bootstrap newBootstrap(ChannelFactory<? extends Channel> factory,
        EventLoopGroup group, AsyncHttpClientConfig config) {
    // No .option(...) calls here — the options map stays empty.
    return new Bootstrap()
            .channelFactory(factory)
            .group(group);
}
```

```java
// In the channel initializer that AHC already installs on the bootstrap:
@Override
protected void initChannel(Channel ch) {
    for (Map.Entry<ChannelOption<?>, Object> e : channelOptions.entrySet()) {
        @SuppressWarnings("unchecked")
        ChannelOption<Object> option = (ChannelOption<Object>) e.getKey();
        ch.config().setOption(option, e.getValue());
    }
    // ... existing pipeline setup ...
}
```

Notes:

- Uses only public Netty API (`Channel.config().setOption`); no package-private
  access and no dependency on Netty-internal `Bootstrap` details, so it is robust
  across Netty versions.
- `CONNECT_TIMEOUT_MILLIS` and the socket options (`SO_*`) are read during/after
  channel init and connect; setting them in `initChannel` (invoked on channel
  registration, before connect) applies them in time, matching the current
  behavior.
- Scope is limited to `ChannelManager` (bootstrap construction + the channel
  initializer). Apply it consistently to all bootstraps AHC creates
  (e.g. the HTTP and WebSocket bootstraps).

### Orthogonal mitigation

Reducing connection churn (effective keep-alive/pooling, a sensible per-host
connection cap, request timeouts aligned with server latency) lowers the connect
rate and therefore how often the lock is hit. That is complementary: Solution A
removes the cost of each connect, while churn reduction lowers the number of
connects. Both together give the largest improvement.

## Alternative approaches (not preferred)

- **Custom `Bootstrap` subclass that caches the options array.** Subclass
  `Bootstrap` (in package `io.netty.bootstrap`) and override `init(Channel)` to
  use a pre-computed, frozen `Map.Entry[]` instead of calling
  `newOptionsArray()`. Works, but depends on package-private Netty internals
  (`init`, `setChannelOptions`, `logger`) and must be kept in sync with the Netty
  version — fragile.
- **Fix in Netty (`AbstractBootstrap`).** Cache the options array in a volatile
  field, rebuild it only inside `option(...)` (setup time), and read it
  lock-free in `newOptionsArray()`. This is the cleanest root fix and benefits
  every `Bootstrap` user, but requires a Netty change (upstream PR or a
  maintained fork). Worth pursuing in addition to the AHC-side fix.

pavel-ptashyts:
## Summary

Fixes #2218.

AsyncHttpClient reuses a single `Bootstrap` instance for every outbound
connection. On each connect, Netty's `AbstractBootstrap.newOptionsArray()`
copies the shared options map under `synchronized (options)`:

```java
synchronized (options) {
    return new ArrayList<Map.Entry<ChannelOption<?>, Object>>(options.entrySet())
            .toArray(EMPTY_OPTION_ARRAY);
}
```
Because the same Bootstrap (and its options map) backs every connection,
this is a single global monitor that all connection attempts contend on.
Under high connection-establishment rates — many concurrent outbound
requests, short-lived connections with low reuse, or no effective per-host
connection cap — threads queue on this lock and a convoy forms.

Fix

Stop configuring options on the Bootstrap. Instead:

- Resolve the configured ChannelOptions once at construction into a
fixed array (buildChannelOptions). Conditional options (connect timeout,
SO_LINGER, send/receive buffers) and all config getters are evaluated a
single time rather than per connection.
- Apply that pre-resolved array to each channel from AHC's existing channel
initializer via the public Channel.config().setOption(...) API
(applyChannelOptions), leaving the Bootstrap options map empty.

This removes the global lock from the connect path entirely while preserving
identical behavior:

- Same option values and ordering — user-supplied options are still
applied last so they override defaults, matching the previous Bootstrap
order.
- Same timing — ChannelInitializer.initChannel runs during channel
registration (handlerAdded, channel already registered), which is before
the connect listener fires, so options are set before the channel connects.
The connect-timeout option is honored because Netty reads it at connect time.
- The SOCKS proxy path is covered automatically, since it installs the HTTP
channel initializer into its pipeline.

Scope is limited to ChannelManager.

Benefits

- Eliminates the lock convoy on the connect path.
- Avoids re-reading config and rebuilding the options array per connection
(fewer per-connect allocations).
- No behavioral or configuration changes for users.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
@@ -18,6 +18,7 @@ package org.asynchttpclient.netty.channel;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelConfig;
 import io.netty.channel.ChannelFactory;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
@@ -93,8 +94,8 @@ import javax.net.ssl.SSLException;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
@@ -129,6 +130,9 @@ public class ChannelManager {
     private final boolean allowReleaseEventLoopGroup;
     private final Bootstrap httpBootstrap;
     private final Bootstrap wsBootstrap;
+    // Channel options, resolved from config once at construction, applied to each channel from the channel
+    // initializer instead of via Bootstrap#option to avoid Netty's synchronized per-connect options map (issue #2218).
+    private final Map.Entry<ChannelOption<?>, Object>[] channelOptions;
     private final long handshakeTimeout;
     private final @Nullable AddressResolverGroup<InetSocketAddress> addressResolverGroup;
 
@@ -200,8 +204,9 @@ public class ChannelManager {
             }
         }
 
-        httpBootstrap = newBootstrap(transportFactory, eventLoopGroup, config);
-        wsBootstrap = newBootstrap(transportFactory, eventLoopGroup, config);
+        channelOptions = buildChannelOptions(config);
+        httpBootstrap = newBootstrap(transportFactory, eventLoopGroup);
+        wsBootstrap = newBootstrap(transportFactory, eventLoopGroup);
 
         // Use the address resolver group from config if provided; otherwise null (legacy per-request resolution)
         addressResolverGroup = config.getAddressResolverGroup();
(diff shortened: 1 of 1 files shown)


pavel-ptashyts:
Hi @hyperxpro could you check this PR. Thanks in advance

hyperxpro:
I will have a look soon along with your other PRs.

PR #2220, opened 2026-06-30
use SecureRandom for digest auth cnonce

jmestwa-coder:
The digest cnonce was drawn from a non-cryptographic PRNG:
- Realm.Builder.newCnonce seeds the cnonce from ThreadLocalRandom, whose output is predictable
- RFC 7616 section 3.3 needs the cnonce unpredictable so it guards against chosen-plaintext and precomputation attacks on the credentials
- NtlmEngine and ScramEngine in this repo already use SecureRandom for their nonces

Switched the cnonce source to SecureRandom to match.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/Realm.java
+++ b/client/src/main/java/org/asynchttpclient/Realm.java
@@ -24,9 +24,8 @@ import org.jetbrains.annotations.Nullable;
 
 import java.nio.charset.Charset;
 import java.security.MessageDigest;
-import java.util.Arrays;
+import java.security.SecureRandom;
 import java.util.Map;
-import java.util.concurrent.ThreadLocalRandom;
 
 import static java.nio.charset.StandardCharsets.ISO_8859_1;
 import static java.nio.charset.StandardCharsets.UTF_8;
@@ -282,6 +281,9 @@ public class Realm {
      * A builder for {@link Realm}
      */
     public static class Builder {
+
+        // cnonce must be unpredictable (RFC 7616 section 3.3), like the NTLM and SCRAM nonces
+        private static final ThreadLocal<SecureRandom> CNONCE_RANDOM = ThreadLocal.withInitial(SecureRandom::new);
 
         private final @Nullable String principal;
         private final @Nullable String password;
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/asynchttpclient/Realm.java
```diff
@@ -283,6 +283,9 @@ public enum AuthScheme {
      */
     public static class Builder {
 
+        // cnonce must be unpredictable (RFC 7616 section 3.3), like the NTLM and SCRAM nonces
+        private static final SecureRandom CNONCE_RANDOM = new SecureRandom();
```
hyperxpro: Don't share one static SecureRandom; use a ThreadLocal. newCnonce runs on Netty event-loop threads (digest header build + 401/407 interceptors), and `SecureRandom#nextBytes` locks internally, so a single shared instance makes all event loops queue on one lock. ThreadLocalRandom had no such lock. ScramEngine already handles this the right way:

```java
private static final ThreadLocal<SecureRandom> CNONCE_RANDOM = ThreadLocal.withInitial(SecureRandom::new);
```

and in newCnonce: 
```java
CNONCE_RANDOM.get().nextBytes(b);
```

client/src/main/java/org/asynchttpclient/Realm.java
```diff
@@ -610,7 +613,7 @@ public Builder parseProxyAuthenticateHeader(String headerLine) {
 
         private void newCnonce(MessageDigest md) {
             byte[] b = new byte[8];
-            ThreadLocalRandom.current().nextBytes(b);
+            CNONCE_RANDOM.nextBytes(b);
             byte[] full = md.digest(b);
```
hyperxpro: the `md.digest(b)` step is now pointless. The 8 bytes are already cryptographically random, so hashing them and trimming back to 8 bytes adds nothing. The method can just be:

```java
private void newCnonce() {
    byte[] b = new byte[8];
    CNONCE_RANDOM.get().nextBytes(b);
    cnonce = toHexString(b);
}
```

Same 16-hex-char cnonce. This is safe because `pooledMessageDigest` already resets the digest before `build()` gets it, so nothing downstream depends on the `digest(b)` call.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/Realm.java
jmestwa-coder: good catch. switched to `ThreadLocal.withInitial(SecureRandom::new)` to match ScramEngine so the event loops don't serialize on one lock.

client/src/main/java/org/asynchttpclient/Realm.java
jmestwa-coder: makes sense. dropped the digest step, newCnonce now hex-encodes the 8 random bytes directly, and removed the now-unused Arrays import. same 16-hex cnonce and the digest handed to newResponse is still the reset pooled instance.

hyperxpro (approved):

PR #2224, opened 2026-07-01
stop config realm leaking to a cross-origin redirect target

jmestwa-coder:
Cross-origin and https-to-http redirects strip the request realm, but exitAfterIntercept re-derived it from config.getRealm(), reattaching a client-wide realm to the redirect target:
- the target answers 401 and then receives the configured credentials, leaking them to a different origin
- Redirect30xInterceptor already clears future.getRealm() on the strip, so read the realm from the future instead of falling back to config
Added a cross-origin 401 case to RedirectCredentialSecurityTest that fails without the change.


jmestwa-coder:
gentle ping

hyperxpro (approved):

PR #2228, opened 2026-07-02
Cap cookies retained per domain (RFC 6265 5.3)

pavel-ptashyts:
ThreadSafeCookieStore kept unbounded cookies per domain, so a server could grow the jar — and the per-request get(Uri) retrieval scan, which visits every cookie in the matching domain buckets — without bound. Bound it at MAX_COOKIES_PER_DOMAIN (200, well above browser per-domain limits of ~50-180 so it only trips under abuse, never for realistic usage). On add, once a domain's bucket exceeds the cap, evictExcessCookies drops expired entries first, then the oldest by creation time until back at the cap.

Eviction uses the two-arg remove(key, value) so a cookie another thread just replaced under the same key is never collaterally removed; a benign race between concurrent adders may evict one extra, which is self-correcting. This bounds both memory and the retrieval scan regardless of how cookies are distributed across paths.

No public API change: the cap is package-private, eviction is private, and getUnderlying() is unchanged. Adds tests that a flooded domain is capped at the limit, that cookies under the cap are all retained, and that the cap is per-domain (flooding one domain does not evict another's).

The audit rated the underlying scan a non-issue for realistic cookie counts; this cap targets its stated pathological case (100+ cookies on one domain) and doubles as mild protection against cookie flooding.


client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -37,6 +37,12 @@
 
 public final class ThreadSafeCookieStore implements CookieStore {
 
+    // RFC 6265 §5.3 (step 12) lets a user agent cap the cookies it retains. Bounding cookies per domain
```
hyperxpro, line 40: The RFC pointer is off. §5.3's storage-model algorithm has no numbered "step 12". The per-domain number limit lives in §5.5 (Implementation Limits, "at least 50 cookies per domain"), and the eviction behavior is the unnumbered "remove excess cookies" prose in §5.3. Also worth noting the RFC breaks ties by least-recently-accessed, not creation time, so we're deviating here on purpose. Suggest citing §5.5 plus §5.3 and saying so in one line. Same fix applies to the javadoc on evictExcessCookies.

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -196,6 +202,33 @@ private void add(String requestDomain, String requestPath, Cookie cookie) {
...
+     * entry. Concurrency-safe: entries are dropped with the two-arg {@code remove(key, value)} so a cookie
+     * another thread just replaced under the same key is never collaterally removed; a benign race between
+     * concurrent adders may evict one extra, which is self-correcting.
+     */
+    private static void evictExcessCookies(Map<CookieKey, StoredCookie> innerMap) {
+        innerMap.entrySet().removeIf(entry -> hasCookieExpired(entry.getValue().cookie, entry.getValue().createdAt));
+        while (innerMap.size() > MAX_COOKIES_PER_DOMAIN) {
+            Map.Entry<CookieKey, StoredCookie> oldest = null;
+            for (Map.Entry<CookieKey, StoredCookie> entry : innerMap.entrySet()) {
+                if (oldest == null || entry.getValue().createdAt < oldest.getValue().createdAt) {
```
hyperxpro, line 224: Ordering by createdAt is weak. It's millisecond granularity, so under a flood most cookies land in the same millisecond and "oldest" becomes whatever ConcurrentHashMap iteration hits first, effectively random. It's also wall-clock based, so an NTP step backward breaks the order. A per-store AtomicLong stamped onto each StoredCookie at construction gives a strict, tie-free order at basically zero cost, and lets you grab the k smallest in a single pass instead of this O(n) rescan on every removal. That would also tighten the over-eviction race above

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -196,6 +202,33 @@ private void add(String requestDomain, String requestPath, Cookie cookie) {
...
+    /**
+     * Bounds a single domain's cookie bucket at {@link #MAX_COOKIES_PER_DOMAIN} (RFC 6265 §5.3 step 12).
+     * Evicts expired entries first, then the oldest by creation time until back at the cap. Called from
+     * {@link #add} right after an insert pushes the bucket over the cap, so it normally removes a single
+     * entry. Concurrency-safe: entries are dropped with the two-arg {@code remove(key, value)} so a cookie
+     * another thread just replaced under the same key is never collaterally removed; a benign race between
+     * concurrent adders may evict one extra, which is self-correcting.
+     */
+    private static void evictExcessCookies(Map<CookieKey, StoredCookie> innerMap) {
+        innerMap.entrySet().removeIf(entry -> hasCookieExpired(entry.getValue().cookie, entry.getValue().createdAt));
```
hyperxpro, line 220: Minor: this re-scans all ~200 entries and calls System.currentTimeMillis() on each, on every over-cap add, even when nothing is expired. Fine at n=200, just flagging that the 200/201 oscillation means every steady-state add pays a full eviction pass with no low-water hysteresis. Not worth optimizing on its own, but the single-pass sequence-number approach would fold this in.

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -196,6 +202,33 @@ private void add(String requestDomain, String requestPath, Cookie cookie) {
...
+            Map.Entry<CookieKey, StoredCookie> oldest = null;
+            for (Map.Entry<CookieKey, StoredCookie> entry : innerMap.entrySet()) {
+                if (oldest == null || entry.getValue().createdAt < oldest.getValue().createdAt) {
+                    oldest = entry;
+                }
+            }
+            if (oldest == null) {
+                break;
+            }
+            innerMap.remove(oldest.getKey(), oldest.getValue());
```
hyperxpro, line 231: Good call on the two-arg remove. Add a one-liner noting it only works because StoredCookie has no equals(), so this is identity based and won't clobber a cookie another thread just re-put under the same key. That invariant is load-bearing and easy to break by accident.

client/src/test/java/org/asynchttpclient/cookie/ThreadSafeCookieStoreGetTest.java
```diff
@@ -112,6 +112,44 @@ public void returnsMultipleDistinctCookiesAtSameDomainPath() {
         assertEquals(setOf("ALPHA=AV", "BETA=BV"), namesValues(store.get(uri)));
     }
 
+    @Test
```
hyperxpro, line 115: One gap: consider a test that mixes expired and live cookies over the cap and asserts eviction drops the expired ones first, to lock in the RFC priority order the code implements.

client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
```diff
@@ -196,6 +202,33 @@ private void add(String requestDomain, String requestPath, Cookie cookie) {
...
+     * Bounds a single domain's cookie bucket at {@link #MAX_COOKIES_PER_DOMAIN} (RFC 6265 §5.3 step 12).
+     * Evicts expired entries first, then the oldest by creation time until back at the cap. Called from
+     * {@link #add} right after an insert pushes the bucket over the cap, so it normally removes a single
+     * entry. Concurrency-safe: entries are dropped with the two-arg {@code remove(key, value)} so a cookie
+     * another thread just replaced under the same key is never collaterally removed; a benign race between
+     * concurrent adders may evict one extra, which is self-correcting.
+     */
+    private static void evictExcessCookies(Map<CookieKey, StoredCookie> innerMap) {
+        innerMap.entrySet().removeIf(entry -> hasCookieExpired(entry.getValue().cookie, entry.getValue().createdAt));
+        while (innerMap.size() > MAX_COOKIES_PER_DOMAIN) {
```
hyperxpro, line 221: Two threads can both observe size == 201, both enter here, and each remove a different oldest entry, so the bucket drops to 199 and we silently evict a valid in-cap cookie. Cookie saves run on Netty IO threads, so two concurrent same-origin responses hitting add() is realistic. It's bounded and the victims are the oldest anyway, so not a blocker, but the PR note ("may evict one extra, self-correcting") undersells it: the store can sit below the cap, not just at it. The sequence-number idea below shrinks this window.

hyperxpro (changes requested):

hyperxpro:
@pavel-ptashyts - I merged it wrong, can you re-open fresh with everything addressed on latest main?

pavel-ptashyts:
@hyperxpro  done. All comment will be fix in new PR

PR #2233, opened 2026-07-02
Read InputStream bodies straight into the buffer, no staging array

pavel-ptashyts:
Motivation:

`InputStreamBodyGenerator.Body.transferTo` allocated a new `byte[]` on every invocation, sized to the target `ByteBuf`'s writable region, then read the `InputStream` into that array before copying the bytes into the `ByteBuf`. This incurred both a per-chunk allocation and an unnecessary copy. Additionally, the implementation reserved a 10-byte margin (`writableBytes() - 10`), which could cause `NegativeArraySizeException` or prematurely stop transfers when the target had 10 or fewer writable bytes.

Modification:

`transferTo` now reads directly from the `InputStream` into the target `ByteBuf` using `ByteBuf.writeBytes(InputStream, int)`, eliminating the intermediate staging array, the extra copy, and the per-instance chunk field.

The state machine is unchanged: each call performs a single stream read, returns `CONTINUE` while data remains, and returns `STOP` at EOF or on an I/O error (with logging unchanged). The legacy 10-byte writable margin has been removed, allowing each transfer to use the full writable capacity and correctly handle small writable regions.

Although AsyncHttpClient's internal request path uses `NettyInputStreamBody`, this implementation remains part of the public `InputStreamBodyGenerator.createBody()` API and is used by external callers.

Result:

Reduces allocation and copy overhead during transfers, correctly supports small writable buffers, preserves existing transfer semantics, and adds tests covering multi-read transfers, single-read completion, empty streams, and transfers through a 10-byte buffer. No public API changes.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/request/body/generator/InputStreamBodyGenerator.java
+++ b/client/src/main/java/org/asynchttpclient/request/body/generator/InputStreamBodyGenerator.java
@@ -58,7 +58,6 @@ public final class InputStreamBodyGenerator implements BodyGenerator {
 
         private final InputStream inputStream;
         private final long contentLength;
-        private byte[] chunk;
 
         private InputStreamBody(InputStream inputStream, long contentLength) {
             this.inputStream = inputStream;
@@ -72,23 +71,17 @@ public final class InputStreamBodyGenerator implements BodyGenerator {
 
         @Override
         public BodyState transferTo(ByteBuf target) {
-
-            // To be safe.
-            chunk = new byte[target.writableBytes() - 10];
-
-            int read = -1;
-            boolean write = false;
+            // Read straight from the stream into the target buffer: no per-call staging byte[] and no extra
+            // copy (ByteBuf.writeBytes(InputStream, int) fills the buffer directly, like FileLikeMultipartPart).
+            // The "- 10" margin preserves the previous behaviour of never fully filling the writable region.
+            int read;
             try {
-                read = inputStream.read(chunk);
+                read = target.writeBytes(inputStream, target.writableBytes() - 10);
             } catch (IOException ex) {
                 LOGGER.warn("Unable to read", ex);
+                return BodyState.STOP;
             }
-
-            if (read > 0) {
-                target.writeBytes(chunk, 0, read);
-                write = true;
-            }
-            return write ? BodyState.CONTINUE : BodyState.STOP;
+            return read > 0 ? BodyState.CONTINUE : BodyState.STOP;
         }
 
         @Override
--- a/client/src/main/java/org/asynchttpclient/request/body/generator/InputStreamBodyGenerator.java
+++ b/client/src/main/java/org/asynchttpclient/request/body/generator/InputStreamBodyGenerator.java
@@ -71,12 +71,13 @@ public final class InputStreamBodyGenerator implements BodyGenerator {
 
         @Override
         public BodyState transferTo(ByteBuf target) {
-            // Read straight from the stream into the target buffer: no per-call staging byte[] and no extra
-            // copy (ByteBuf.writeBytes(InputStream, int) fills the buffer directly, like FileLikeMultipartPart).
-            // The "- 10" margin preserves the previous behaviour of never fully filling the writable region.
+            // Read straight from the stream into the target buffer instead of staging through a per-call byte[].
+            // For heap target buffers this drops both the staging array and the copy; for direct buffers Netty
+            // still stages through a temporary heap array internally (InputStream can only read into a byte[]),
+            // so there the win is smaller. Mirrors InputStreamMultipartPart, which writes the full writable region.
             int read;
             try {
-                read = target.writeBytes(inputStream, target.writableBytes() - 10);
+                read = target.writeBytes(inputStream, target.writableBytes());
             } catch (IOException ex) {
                 LOGGER.warn("Unable to read", ex);
                 return BodyState.STOP;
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/asynchttpclient/request/body/generator/InputStreamBodyGenerator.java
```diff
@@ -72,23 +71,17 @@ public long getContentLength() {
...
-
-            int read = -1;
-            boolean write = false;
+            // Read straight from the stream into the target buffer: no per-call staging byte[] and no extra
+            // copy (ByteBuf.writeBytes(InputStream, int) fills the buffer directly, like FileLikeMultipartPart).
+            // The "- 10" margin preserves the previous behaviour of never fully filling the writable region.
+            int read;
             try {
-                read = inputStream.read(chunk);
+                read = target.writeBytes(inputStream, target.writableBytes() - 10);
```
hyperxpro: Can we drop the `- 10` here instead of keeping it? I dug into the history, and it comes from an old commit literally titled "More safe guess, still need a better solution", so it doesn't appear to have been an intentional requirement. None of the current consumers (`BodyChunkedInput`, `NettyBodyBody`, `AuthenticatorUtils`) write anything into the buffer after `transferTo`, and `InputStreamMultipartPart` already uses `target.writeBytes(inputStream, target.writableBytes())` with no reserved margin.

Keeping it also preserves two edge-case bugs. If `writableBytes() < 10`, the computed length becomes negative and Netty throws `IllegalArgumentException`, which isn't caught by the existing `IOException` handler. (The old code failed here too, but with `NegativeArraySizeException`.) If `writableBytes() == 10`, the computed length is `0`, `read()` returns `0`, and we return `STOP` without writing anything, silently truncating a stream that still has data.

This is reachable via something like `new InputStreamBodyGenerator(stream, 10L)` through `BodyChunkedInput`, since its chunk size is `min(contentLength, 8192)`. Only external callers can hit this today, but since this PR is cleaning up exactly this legacy code, it seems like the right opportunity to remove the workaround rather than preserve its behavior. A small test covering `contentLength <= 10` would lock the fix in.

client/src/main/java/org/asynchttpclient/request/body/generator/InputStreamBodyGenerator.java
```diff
@@ -72,23 +71,17 @@ public long getContentLength() {
 
         @Override
         public BodyState transferTo(ByteBuf target) {
-
-            // To be safe.
-            chunk = new byte[target.writableBytes() - 10];
-
-            int read = -1;
-            boolean write = false;
+            // Read straight from the stream into the target buffer: no per-call staging byte[] and no extra
```
hyperxpro: Two corrections are needed in this comment.

First, "no per-call staging `byte[]` and no extra copy" is only true for heap buffers. `InputStream.read()` can only read into a `byte[]`, so for direct buffers Netty still has to stage through heap memory internally. For an 8 KB chunk, `threadLocalTempArray` allocates a new array on every call because the size exceeds the 1024-byte thread-local cache, and the unsafe path acquires a pooled heap buffer and performs a `copyMemory()` on every call. `BodyChunkedInput` allocates from the channel allocator, which prefers direct buffers by default, so on that path this change is roughly allocation-for-allocation equivalent to the previous implementation. The improvement is real for heap buffers, which is what the new test exercises, so I'd reword the comment to make that distinction.

Second, `FileLikeMultipartPart` is the wrong reference. It's abstract and never calls `writeBytes(InputStream, int)`. The class that actually uses this pattern is `InputStreamMultipartPart`, and it passes the full `writableBytes()` without reserving any margin.

client/src/test/java/org/asynchttpclient/request/body/generator/InputStreamBodyGeneratorTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Covers {@link InputStreamBodyGenerator}'s {@link Body#transferTo(ByteBuf)}, which now reads straight from
+ * the stream into the target buffer (no per-call staging {@code byte[]} and no extra copy). The whole stream
+ * must still be transferred byte-for-byte, CONTINUE while data remains and STOP at EOF.
+ */
+public class InputStreamBodyGeneratorTest {
+
+    private static final int CHUNK_SIZE = 1024 * 8;
+
```
hyperxpro, line 40: Minor: the existing tests in this package (`ByteArrayBodyGeneratorTest` and `FeedableBodyGeneratorTest`) use `@RepeatedIfExceptionsTest(repeats = 5)` on every test method, so using plain `@Test` breaks the local convention.

`ByteArrayBodyGeneratorTest` also already has the same single-read and multi-read scaffold (same chunk size, same `3 * CHUNK_SIZE + 42` sizing, and the same drain loop), while `FeedableBodyGeneratorTest` provides a `readFromBody` helper that covers the short-stream cases. It would be worth aligning with the existing annotation and reusing the existing test patterns rather than introducing a second drain idiom to the package.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/request/body/generator/InputStreamBodyGenerator.java
pavel-ptashyts: Done

client/src/main/java/org/asynchttpclient/request/body/generator/InputStreamBodyGenerator.java
pavel-ptashyts: Done

client/src/test/java/org/asynchttpclient/request/body/generator/InputStreamBodyGeneratorTest.java
pavel-ptashyts: Done

hyperxpro (approved):

PR #2234, opened 2026-07-04
skip origin Authorization on CONNECT proxy request

madib06ops:
newNettyRequest adds the origin Authorization header to every request it builds, including the CONNECT that opens a proxy tunnel. That CONNECT goes to the proxy in the clear before the TLS tunnel exists, so preemptive Basic or Digest credentials meant for the origin are handed to the proxy. Gate the origin Authorization on the non-CONNECT case, matching the proxy-auth line right below it; the credentials still travel on the request sent once the tunnel is up.

NTLM/Kerberos/SPNEGO realms take a different path: their header is attached in `NettyRequestSender.sendRequestWithNewChannel` via `perConnectionAuthorizationHeader`, applied to `future.getNettyRequest()`, which is the CONNECT on the tunnel path. Same leak, so that call is now guarded on the request method too. Added `ConnectRequestNtlmAuthorizationTest`, which drives a preemptive NTLM request through an HTTP proxy and asserts the CONNECT the proxy receives carries no Authorization header (it fails without the sender-side guard).

Behavior note: `connect` is derived from the method, so a caller using `prepareConnect()` directly with a preemptive realm previously got an Authorization header on that CONNECT and no longer does. That's the intended fix (the header should never ride the plaintext CONNECT), but it's an observable change worth calling out for release notes.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
@@ -306,8 +306,12 @@ public final class NettyRequestFactory {
             headers.set(HOST, virtualHost != null ? virtualHost : hostHeader(uri));
         }
 
-        // don't override authorization but append
-        addAuthorizationHeader(headers, perRequestAuthorizationHeader(request, realm));
+        // don't override authorization but append. Skip it on a CONNECT: that request is sent to the
+        // proxy in the clear to open the tunnel, so the origin Authorization would be exposed to the
+        // proxy. It is added to the tunneled request, which is built separately once the tunnel is up.
+        if (!connect) {
+            addAuthorizationHeader(headers, perRequestAuthorizationHeader(request, realm));
+        }
         // only set proxy auth on request over plain HTTP, or when performing CONNECT
         if (!uri.isSecured() || connect) {
             setProxyAuthorizationHeader(headers, perRequestProxyAuthorizationHeader(request, proxyRealm));
--- a/client/src/test/java/org/asynchttpclient/netty/request/ConnectRequestAuthorizationTest.java
+++ b/client/src/test/java/org/asynchttpclient/netty/request/ConnectRequestAuthorizationTest.java
@@ -1,0 +1,65 @@
+/*
+ *    Copyright (c) 2026 AsyncHttpClient Project. All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package org.asynchttpclient.netty.request;
+
+import io.netty.handler.codec.http.HttpHeaderNames;
+import org.asynchttpclient.Realm;
+import org.asynchttpclient.Request;
+import org.asynchttpclient.proxy.ProxyServer;
+import org.junit.jupiter.api.Test;
+
+import static org.asynchttpclient.Dsl.basicAuthRealm;
+import static org.asynchttpclient.Dsl.config;
+import static org.asynchttpclient.Dsl.get;
+import static org.asynchttpclient.Dsl.proxyServer;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * A CONNECT request opens the proxy tunnel and is sent to the proxy in the clear, so it must not carry the
+ * origin {@code Authorization} header (which would expose the origin credentials to the proxy). The header
+ * belongs only on the request sent through the established tunnel.
+ */
+public class ConnectRequestAuthorizationTest {
+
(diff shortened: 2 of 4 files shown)


client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
```diff
@@ -306,8 +306,12 @@ public NettyRequest newNettyRequest(Request request, boolean performConnectReque
             headers.set(HOST, virtualHost != null ? virtualHost : hostHeader(uri));
         }
 
-        // don't override authorization but append
-        addAuthorizationHeader(headers, perRequestAuthorizationHeader(request, realm));
+        // don't override authorization but append. Skip it on a CONNECT: that request is sent to the
+        // proxy in the clear to open the tunnel, so the origin Authorization would be exposed to the
+        // proxy. It is added to the tunneled request, which is built separately once the tunnel is up.
+        if (!connect) {
```
hyperxpro, line 312: Heads up that `connect` here is derived from the method, not just `performConnectRequest` (line 207). This also changes behavior for a user calling `prepareConnect()` directly. With a preemptive realm, they previously received an `Authorization` header on that CONNECT request, and now it is silently omitted with no tunneled follow-up to restore it. I think dropping it is actually the right call, but it is an observable change to a public API, so it is worth mentioning in the PR description or release notes.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
```diff
@@ -306,8 +306,12 @@ public NettyRequest newNettyRequest(Request request, boolean performConnectReque
...
+            addAuthorizationHeader(headers, perRequestAuthorizationHeader(request, realm));
```
hyperxpro, line 313: Not blocking for this PR, but the CONNECT header policy now lives in two places: the positive whitelist at lines 231–234 and this negative gate down here. Anything added unconditionally later still gets sent on CONNECT today (for example, the default `Accept: */*` at line 318. It is harmless, but it illustrates the pattern). 

A self-contained CONNECT header assembly, or a final whitelist applied when `connect` is true, would make this whole class of leaks impossible instead of relying on everyone remembering the `!connect` check. Fine as a follow-up.

client/src/test/java/org/asynchttpclient/netty/request/ConnectRequestAuthorizationTest.java
```diff
@@ -0,0 +1,65 @@
...
+ * belongs only on the request sent through the established tunnel.
+ */
+public class ConnectRequestAuthorizationTest {
+
+    private static NettyRequestFactory factory() {
+        return new NettyRequestFactory(config().build());
+    }
+
+    @Test
+    public void connectRequestDoesNotCarryOriginAuthorization() {
```
hyperxpro, line 43: Could you add a test case with an NTLM (or SPNEGO) realm? Those use `perConnectionAuthorizationHeader` in `NettyRequestSender` rather than this code path, so they would still leak onto the CONNECT request. A failing test would pin down the remaining gap.

hyperxpro (changes requested):
Good catch on the preemptive Basic/Digest case. The tunneled request still picks the header back up after the CONNECT succeeds, so nothing breaks there.

This doesn't fully close the leak, though. In `NettyRequestSender.sendRequestWithNewChannel` around line 413:

```java
requestFactory.addAuthorizationHeader(headers, perConnectionAuthorizationHeader(request, proxy, realm));
```

On the tunnel path, `future.getNettyRequest()` is the CONNECT request, so preemptive NTLM/Kerberos/SPNEGO realms still attach the origin `Authorization` header to the plaintext CONNECT. It's the same bug through a different code path. Could you guard this as well? An NTLM test would also be a nice addition to keep this path covered.

client/src/test/java/org/asynchttpclient/netty/request/ConnectRequestAuthorizationTest.java
madib06ops: Added ConnectRequestNtlmAuthorizationTest in the proxy package. It runs a preemptive NTLM request through an HTTP proxy to an https origin and asserts the CONNECT the proxy receives has no Authorization header. Confirmed it fails without the sender-side guard (CONNECT carries `NTLM TlRM...`) and passes with it.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
madib06ops: Good point, that's an observable change for anyone calling prepareConnect() directly. Dropping the header off the plaintext CONNECT is the behavior I want, so I've called it out in the PR description for release notes.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
madib06ops: Agreed, the split policy is fragile. A whitelist applied when connect is true would make the whole class of leaks structural instead of relying on the !connect check being remembered everywhere. I'd rather keep this PR to the security fix and do that consolidation as a focused follow-up so it doesn't churn the header assembly here.

madib06ops:
You're right, the per-connection path was the same bug through a different door. I've guarded the addAuthorizationHeader call in sendRequestWithNewChannel on the request method, so preemptive NTLM/Kerberos/SPNEGO no longer attach the origin header to the CONNECT. New ConnectRequestNtlmAuthorizationTest covers that path and fails without the guard. Existing HttpsProxyTest/HttpsProxyBasicTest/NTLMProxyTest still pass, and I noted the prepareConnect() behavior change in the description.

hyperxpro (approved):

hyperxpro:
Thanks a lot!

PR #2237, opened 2026-07-15
reject backslash in the parsed URI authority

madib06ops:
UriParser.parseAuthority never rejects a backslash in the authority, so it slips past the '@' userinfo split and Uri.create("https://trusted.com\@evil.com/") resolves the connect host to evil.com, while java.net.URI rejects that string and a WHATWG parser reads trusted.com. An app that allowlists the host before handing the raw URL, or a server-supplied redirect Location, to the client then connects to a host it never checked. This rejects a backslash in the parsed authority per RFC 3986, which does not permit it there, and leaves a backslash in the path untouched.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/uri/UriParser.java
+++ b/client/src/main/java/org/asynchttpclient/uri/UriParser.java
@@ -334,6 +334,15 @@ final class UriParser {
             currentIndex += 2;
 
             String nonNullAuthority = computeAuthority();
+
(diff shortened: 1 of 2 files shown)


hyperxpro:
There is a separate patch for this I cannot pull this in via PR.

madib06ops:
Sounds good, glad the authority backslash case is getting fixed. Happy to have it land however is cleanest on your end.

PR #2240, opened 2026-07-17
ignore a Set-Cookie scoped to a bare TLD

madib06ops:
ThreadSafeCookieStore.add only applies rfc6265 section 5.3 step 6, so a Domain attribute that is a bare public suffix survives: domainsMatch("com", "evil.com") is true because evil.com ends with ".com". A response from evil.com carrying `Set-Cookie: SESSIONID=attacker; Domain=com; Path=/` (cookies arrive here from Interceptors.java:107) is stored under key domain "com" with hostOnly false. get() then walks the request host's parent labels through DomainUtils.getSubDomain, so a later request to bank.com resolves "bank.com", then "com", and picks the cookie up. Any host can plant a session cookie on every sibling under its TLD.

Step 5 is the missing rule, so the guard goes next to the existing step 6 check in add() rather than in get(): the walk is only reachable because the entry was accepted in the first place, and keying the jar correctly keeps the read path untouched. Without a public suffix list the approximation is a single-label domain, matching the embedded-dot requirement java.net.HttpCookie already enforces. Domain=localhost on a localhost request keeps working and becomes host-only, which is what it already resolved to in practice. Multi-label suffixes like co.uk still need a real list and are out of reach here. Two regression tests in ThreadSafeCookieStoreGetTest fail without the guard.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
+++ b/client/src/main/java/org/asynchttpclient/cookie/ThreadSafeCookieStore.java
@@ -181,6 +181,17 @@ public final class ThreadSafeCookieStore implements CookieStore {
         String keyDomain = pair.getKey();
         boolean hostOnly = pair.getValue();
 
+        // rfc6265#section-5.3 step 5: a Domain attribute that is a public suffix must not scope a
+        // cookie to that suffix, otherwise any host under it can plant cookies for every sibling.
+        // Approximated without a public suffix list by treating a single-label domain (a TLD such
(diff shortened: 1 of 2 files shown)


hyperxpro:
There is a separate patch for this I cannot pull this in via PR.

madib06ops:
Understood, thanks for the heads up. Glad it's already being handled. If the separate patch covers the bare-TLD case in ThreadSafeCookieStore.add (a single-label Domain like `com` passing the step 6 domainsMatch check), then we're good. Happy to help review or add the regression tests once it lands if that's useful.

PR #2250, opened 2026-07-18
Support HTTP/2 after proxy CONNECT

### LINKED ISSUE #2241: HTTP/2 can be negotiated after proxy CONNECT, but AHC still sends HTTP/1.1
```markdown
## Summary

When HTTP/2 is enabled and an HTTPS request is sent through an HTTP proxy, AHC can negotiate `h2` with the target server after the proxy `CONNECT`, but still send the target request as HTTP/1.1.

In that case the server expects HTTP/2 frames, receives an HTTP/1.1 request instead, and closes the connection. Client-side this shows up as:

```
org.asynchttpclient.exception.RemotelyClosedException: Remotely closed
```

## Version

Reproduced with:

- `org.asynchttpclient:async-http-client:3.0.10`

This appears related to the HTTP/2 multiplexing support added in #2144.

## Why I think this is happening

For direct HTTPS, `NettyConnectListener` checks the negotiated ALPN protocol after TLS:

```java
String alpnProtocol = sslHandler.applicationProtocol();
if (ApplicationProtocolNames.HTTP_2.equals(alpnProtocol)) {
    channelManager.upgradePipelineToHttp2(channel.pipeline());
    registerHttp2AndReleaseSemaphore(channel);
    releaseSemaphoreImmediately(partitionKeyLock);
}
```

But after proxy `CONNECT`, the tunnel path appears to install the target TLS handler and then add `HttpClientCodec`:

```java
SslHandler sslHandler = createSslHandler(requestUri.getHost(), requestUri.getExplicitPort());
whenHandshaked = sslHandler.handshakeFuture();
pipeline.addBefore(INFLATER_HANDLER, SSL_HANDLER, sslHandler);
pipeline.addAfter(SSL_HANDLER, HTTP_CLIENT_CODEC, newHttpClientCodec());
```

I do not see a corresponding post-CONNECT ALPN check or HTTP/2 pipeline upgrade in `updatePipelineForHttpTunneling(...)` / `updatePipelineForHttpsTunneling(...)`.

There is also an existing test comment that seems to acknowledge this path is unsupported:

```java
// HTTP/2 ALPN upgrade after proxy CONNECT tunnel is not yet supported
.setHttp2Enabled(false)
```

The concern is that HTTP/2 is enabled by default, so this unsupported path can still be reached unless users know to disable HTTP/2 manually.

## Minimal repro shape

A local test can reproduce this with:

- a Netty TLS HTTP/2 server that advertises `h2` through ALPN;
- an HTTP proxy that supports `CONNECT`;
- an AHC client with `setUseInsecureTrustManager(true)` and default HTTP/2 enabled;
- a request to `https://localhost:<h2-server-port>/hello` through the HTTP proxy.

Expected:

- AHC should either upgrade the post-CONNECT tunnel to HTTP/2 when ALPN selects `h2`, or avoid advertising `h2` for proxied HTTPS requests until HTTP/2 after CONNECT is supported.

Actual:

- AHC sends the target request as HTTP/1.1.
- Client fails with:

```text
org.asynchttpclient.exception.RemotelyClosedException: Remotely closed
```

- Server-side Netty reports:

```text
io.netty.handler.codec.http2.Http2Exception: Unexpected HTTP/1.x request: GET /hello
```

## Possible fixes

I think either of these would be reasonable.

### Option 1: Support HTTP/2 after CONNECT

After the target TLS handshake completes in the proxy tunnel path:

1. inspect `sslHandler.applicationProtocol()`;
2. if it is `ApplicationProtocolNames.HTTP_2`, call `upgradePipelineToHttp2(...)`;
3. register the physical tunnel channel in the HTTP/2 connection registry using the same partitioning key used by the request;
4. send the target request through the existing HTTP/2 request path.

This would mirror the direct HTTPS behavior.

### Option 2: Safe fallback until supported

If HTTP/2 after proxy `CONNECT` is not intended to be supported yet, AHC could avoid advertising `h2` in the target TLS handshake for proxied HTTPS requests. In that case, proxied HTTPS traffic would safely negotiate/use HTTP/1.1 even when global `http2Enabled=true`.

This would avoid requiring every user of proxy `CONNECT` to know that they must call `setHttp2Enabled(false)` manually.

I’m happy to contribute a fix for option 1 if that direction sounds acceptable. I have a local patch that checks the ALPN result after the target TLS handshake in the CONNECT tunnel path, switches to the HTTP/2 request path when `h2` is selected, and adds a regression test for HTTPS-over-CONNECT with HTTP/2.

## Question

Which behavior would maintainers prefer here?

rampreeth:
## Motivation

When an HTTPS request uses an HTTP proxy, AHC establishes a `CONNECT` tunnel and then performs TLS with the target. If ALPN selects `h2`, the tunnel path currently leaves the HTTP/1.1 codec in place. The target then receives HTTP/1.1 bytes on a connection that negotiated HTTP/2 and closes it.

## Modification

- Check the negotiated ALPN protocol after the target TLS handshake in both proxy tunnel paths.
- When ALPN selects `h2`, upgrade the existing channel pipeline to HTTP/2 and register the physical tunnel connection with its existing pool partition key.
- Preserve the existing HTTP/1.1 fallback and WebSocket behavior.
- Add regression tests using local HTTP and HTTPS `CONNECT` proxies with a TLS HTTP/2 target.
- Verify that the upgraded tunnel is registered and reused for a second request.
- Preserve the existing public tunneling method signatures for binary compatibility.

## Result

HTTPS requests through HTTP and HTTPS `CONNECT` proxies now send HTTP/2 frames when the target negotiates `h2`. The regression tests verify successful HTTP/2 responses and reuse of a single physical target connection.

## Testing

- `./mvnw -pl client -Dtest=BasicHttp2Test,HttpsProxyTest test` — 60 tests passed.
- Full client test suite passed as part of `verify`; the local run then stopped at artifact signing because `gpg` is not installed.
- `./mvnw -pl client -DskipTests -Dgpg.skip=true verify` — packaging and API compatibility checks passed.

Fixes #2241

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
@@ -122,6 +122,7 @@ public class ChannelManager {
     public static final String HTTP2_FRAME_CODEC = "http2-frame-codec";
     public static final String HTTP2_MULTIPLEX = "http2-multiplex";
     public static final String AHC_HTTP2_HANDLER = "ahc-http2";
+    private static final String TARGET_SSL_HANDLER = "target-ssl";
     private static final Logger LOGGER = LoggerFactory.getLogger(ChannelManager.class);
     // Guards the one-time WARN emitted when a native transport was requested but is unavailable and we
     // fall back to NIO. Logged once per JVM to avoid spamming logs when many clients are created.
@@ -630,10 +631,6 @@ public class ChannelManager {
     }
 
     public Future<Channel> updatePipelineForHttpTunneling(ChannelPipeline pipeline, Uri requestUri) {
-        return updatePipelineForHttpTunneling(pipeline, requestUri, null);
-    }
-
-    public Future<Channel> updatePipelineForHttpTunneling(ChannelPipeline pipeline, Uri requestUri, Object partitionKey) {
         Future<Channel> whenHandshaked = null;
 
         if (pipeline.get(HTTP_CLIENT_CODEC) != null) {
@@ -651,7 +648,6 @@ public class ChannelManager {
             whenHandshaked = sslHandler.handshakeFuture();
             pipeline.addBefore(INFLATER_HANDLER, SSL_HANDLER, sslHandler);
             pipeline.addAfter(SSL_HANDLER, HTTP_CLIENT_CODEC, newHttpClientCodec());
-            upgradePipelineToHttp2AfterHandshake(whenHandshaked, sslHandler, pipeline, requestUri, partitionKey);
 
         } else {
             // For HTTP targets, remove any existing SSL handler (from HTTPS proxy) since target is not secured
@@ -674,11 +670,6 @@ public class ChannelManager {
     }
 
     public Future<Channel> updatePipelineForHttpsTunneling(ChannelPipeline pipeline, Uri requestUri, ProxyServer proxyServer) {
-        return updatePipelineForHttpsTunneling(pipeline, requestUri, proxyServer, null);
-    }
-
-    public Future<Channel> updatePipelineForHttpsTunneling(ChannelPipeline pipeline, Uri requestUri, ProxyServer proxyServer,
-                                                           Object partitionKey) {
         Future<Channel> whenHandshaked = null;
 
         // Remove HTTP codec as tunnel is established
@@ -698,14 +689,13 @@ public class ChannelManager {
             // This creates a nested SSL setup: Target SSL -> Proxy SSL -> Network
             if (isSslHandlerConfigured(pipeline)) {
                 // Insert target SSL handler after the proxy SSL handler
-                pipeline.addAfter(SSL_HANDLER, "target-ssl", sslHandler);
+                pipeline.addAfter(SSL_HANDLER, TARGET_SSL_HANDLER, sslHandler);
             } else {
                 // This shouldn't happen for HTTPS proxy, but fallback
                 pipeline.addBefore(INFLATER_HANDLER, SSL_HANDLER, sslHandler);
             }
             
-            pipeline.addAfter("target-ssl", HTTP_CLIENT_CODEC, newHttpClientCodec());
-            upgradePipelineToHttp2AfterHandshake(whenHandshaked, sslHandler, pipeline, requestUri, partitionKey);
+            pipeline.addAfter(TARGET_SSL_HANDLER, HTTP_CLIENT_CODEC, newHttpClientCodec());
 
         } else {
             // For HTTPS proxy to HTTP target, just add HTTP codec
@@ -724,25 +714,6 @@ public class ChannelManager {
         }
         
         return whenHandshaked;
-    }
-
-    private void upgradePipelineToHttp2AfterHandshake(Future<Channel> whenHandshaked,
-                                                       SslHandler sslHandler,
-                                                       ChannelPipeline pipeline,
-                                                       Uri requestUri,
-                                                       Object partitionKey) {
-        if (!config.isHttp2Enabled() || requestUri.isWebSocket()) {
-            return;
-        }
-
-        whenHandshaked.addListener(f -> {
-            if (f.isSuccess() && ApplicationProtocolNames.HTTP_2.equals(sslHandler.applicationProtocol())) {
-                upgradePipelineToHttp2(pipeline);
-                if (partitionKey != null) {
-                    registerHttp2Connection(partitionKey, pipeline.channel());
-                }
-            }
-        });
     }
 
     public SslHandler addSslHandler(ChannelPipeline pipeline, Uri uri, String virtualHost, boolean hasSocksProxyHandler) {
@@ -993,6 +964,23 @@ public class ChannelManager {
         if (pingIntervalMs > 0) {
             pipeline.addLast("http2-idle-state", new IdleStateHandler(0, 0, pingIntervalMs, TimeUnit.MILLISECONDS));
             pipeline.addLast("http2-ping", new Http2PingHandler());
+        }
+    }
+
+    /**
+     * Upgrades and registers a proxy tunnel when the target TLS handshake negotiated HTTP/2.
+     * The caller controls when this runs so the upgrade can happen immediately before the
+     * tunneled request is sent.
+     */
+    public void upgradePipelineToHttp2AfterProxyConnect(ChannelPipeline pipeline, Object partitionKey) {
+        SslHandler targetSslHandler = (SslHandler) pipeline.get(TARGET_SSL_HANDLER);
+        if (targetSslHandler == null) {
+            targetSslHandler = (SslHandler) pipeline.get(SSL_HANDLER);
+        }
+        if (targetSslHandler != null
+                && ApplicationProtocolNames.HTTP_2.equals(targetSslHandler.applicationProtocol())) {
+            upgradePipelineToHttp2(pipeline);
+            registerHttp2Connection(partitionKey, pipeline.channel());
         }
     }
 
--- a/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
@@ -51,6 +51,7 @@ import io.netty.handler.logging.LoggingHandler;
 import io.netty.handler.proxy.ProxyHandler;
 import io.netty.handler.proxy.Socks4ProxyHandler;
 import io.netty.handler.proxy.Socks5ProxyHandler;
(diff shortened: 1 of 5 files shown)


client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -714,6 +726,25 @@ public Future<Channel> updatePipelineForHttpsTunneling(ChannelPipeline pipeline,
...
+    private void upgradePipelineToHttp2AfterHandshake(Future<Channel> whenHandshaked,
+                                                       SslHandler sslHandler,
+                                                       ChannelPipeline pipeline,
+                                                       Uri requestUri,
+                                                       Object partitionKey) {
+        if (!config.isHttp2Enabled() || requestUri.isWebSocket()) {
+            return;
+        }
+
+        whenHandshaked.addListener(f -> {
```
hyperxpro: This upgrade has to run before the `sendNextRequest` listener that `NettyRequestSender.drainChannelAndExecuteNextRequest` adds to the same `whenHandshaked` future, otherwise `writeRequest` would see the pipeline still on HTTP/1.1 and send HTTP/1.1. It works today because this listener is registered first (during CONNECT handling) and the drain listener is registered later, when the CONNECT last-content is read, so Netty fires them in that order. That is a real correctness invariant, but it lives across two files and is easy to break by accident. Could we either add a short comment here spelling it out, or move the ALPN check and upgrade into the same listener that calls `sendNextRequest` so the ordering is guaranteed locally?

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -629,6 +630,10 @@ private SslHandler createSslHandler(String peerHost, int peerPort, boolean http2
     }
 
     public Future<Channel> updatePipelineForHttpTunneling(ChannelPipeline pipeline, Uri requestUri) {
+        return updatePipelineForHttpTunneling(pipeline, requestUri, null);
+    }
+
+    public Future<Channel> updatePipelineForHttpTunneling(ChannelPipeline pipeline, Uri requestUri, Object partitionKey) {
```
hyperxpro: Small design note: these methods were pure pipeline setup, and now they also schedule a post-handshake side effect. Passing partitionKey in just so it can be handed to registerHttp2Connection inside the listener works, but it does blur the responsibility a bit. Keeping the overloads for binary compatibility is a good call though.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -714,6 +726,25 @@ public Future<Channel> updatePipelineForHttpsTunneling(ChannelPipeline pipeline,
         return whenHandshaked;
     }
 
+    private void upgradePipelineToHttp2AfterHandshake(Future<Channel> whenHandshaked,
+                                                       SslHandler sslHandler,
+                                                       ChannelPipeline pipeline,
+                                                       Uri requestUri,
+                                                       Object partitionKey) {
+        if (!config.isHttp2Enabled() || requestUri.isWebSocket()) {
```
hyperxpro: Minor: the config.isHttp2Enabled() half is belt and suspenders, since with HTTP/2 disabled the SSL engine never advertises h2 and applicationProtocol() will not equal h2. Harmless to keep, just noting it is not strictly required. The WebSocket guard is worth keeping.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -714,6 +726,25 @@ public Future<Channel> updatePipelineForHttpsTunneling(ChannelPipeline pipeline,
...
+            return;
+        }
+
+        whenHandshaked.addListener(f -> {
+            if (f.isSuccess() && ApplicationProtocolNames.HTTP_2.equals(sslHandler.applicationProtocol())) {
+                upgradePipelineToHttp2(pipeline);
+                if (partitionKey != null) {
+                    registerHttp2Connection(partitionKey, pipeline.channel());
```
hyperxpro: `future.getPartitionKey()` is effectively never null here, so the partitionKey != null guard is defensive only. Fine to keep, just flagging that if it ever were null the tunnel would upgrade but never get registered for reuse, and that would be silent.

client/src/test/java/org/asynchttpclient/BasicHttp2Test.java
```diff
@@ -430,6 +438,50 @@ private AsyncHttpClient http2ClientWithConfig(Consumer<DefaultAsyncHttpClientCon
...
+        assertHttpsRequestsThroughProxyNegotiateAndReuseHttp2AfterConnect(false);
+    }
+
+    @Test
+    public void httpsRequestsThroughHttpsProxyNegotiateAndReuseHttp2AfterConnect() throws Exception {
+        assertHttpsRequestsThroughProxyNegotiateAndReuseHttp2AfterConnect(true);
+    }
+
+    private void assertHttpsRequestsThroughProxyNegotiateAndReuseHttp2AfterConnect(boolean secureProxy) throws Exception {
+        Server proxy = new Server();
```
hyperxpro, line 452: `proxy.start()` is outside the try/finally. Nothing between here and the try throws today, so it is fine, but moving the start inside the try (or the Server proxy creation into a try-with-resources style teardown) would make the cleanup airtight if setup ever grows.

hyperxpro (changes requested):
Not in this PR's diff, so leaving it here: this change makes HTTP/2 after CONNECT work, which makes the workaround in `client/src/test/java/org/asynchttpclient/proxy/HttpsProxyTestcontainersIntegrationTest.java` stale and its comment misleading. Could you remove `.setHttp2Enabled(false)` and the `// HTTP/2 ALPN upgrade after proxy CONNECT tunnel is not yet supported` comment in both `testHttpProxyToHttpsTarget` (lines 149-150) and `testHttpsProxyToHttpsTarget` (lines 172-173)? That lets those external tests exercise the new HTTP/2-through-proxy path instead of forcing HTTP/1.1, and the comment stops claiming something that is no longer true.

rampreeth:
Done - removed `.setHttp2Enabled(false)` and the stale comment from both proxy-to-HTTPS Testcontainers tests. They now use the default HTTP/2-enabled configuration.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
rampreeth: Good catch. I moved the ALPN check and pipeline upgrade into the same handshake listener that sends the next request, so the ordering is local now.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
rampreeth: Agreed. I removed the new overloads and kept the pipeline setup methods focused on setup. The post-handshake upgrade now happens explicitly from `NettyRequestSender`, which already has the partition key.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
rampreeth: Yep, removed the redundant config check. We now rely on the negotiated ALPN result, with the WebSocket guard still in `NettyRequestSender`.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
rampreeth: Removed the null guard. Registration now always uses `future.getPartitionKey()`, so a bad null cannot silently leave an upgraded connection unregistered.

client/src/test/java/org/asynchttpclient/BasicHttp2Test.java
rampreeth: Done -`proxy.start()` is now inside the `try`, so `proxy.stop()` still runs from `finally` if setup fails.

rampreeth:
Hi @hyperxpro, just a gentle reminder when you get a chance. I've addressed the review comments and would appreciate a re-review. Thanks!

hyperxpro (approved):

hyperxpro:
Thanks a lot!

rampreeth:
Hey @hyperxpro - thanks again for reviewing and merging this! Could you share how the release process usually works for this project? Will this change be included in the next release, and is there a rough timeline for it?

hyperxpro:
There is no fixed release cycle - but I hope to cut a release next week,

rampreeth:
Hey @hyperxpro - just wanted to follow up on the release. Would there be one this week?

hyperxpro:
> Hey @hyperxpro - just wanted to follow up on the release. Would there be one this week?

It will be Wednesday

PR #2260, opened 2026-07-19
Avoid boxed redirect status lookup

pavel-ptashyts:
### Summary
- Replace the redirect status `Set<Integer>` lookup with a primitive status-code helper.
- Remove `Redirect30xInterceptor.REDIRECT_STATUSES`.
- Use the helper from the response interceptor hot path before redirect handling.
- Add unit coverage for redirect and non-redirect status codes.

### Motivation
The response interceptor checks redirect status codes for every response. The previous `Set<Integer>` lookup boxed each response status code before lookup. A primitive comparison helper keeps the redirect behavior unchanged while avoiding that allocation.

### Validation
- `/opt/maven/bin/mvn -pl client -DskipTests compile`
- `/opt/maven/bin/mvn -pl client -Dtest=Redirect30xInterceptorTest,Relative302Test test`

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
@@ -125,7 +125,7 @@ public class Interceptors {
             return continue100Interceptor.exitAfterHandling100(channel, future);
         }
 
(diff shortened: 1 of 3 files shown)


pavel-ptashyts:
Closing because this performance change will not be pursued.

PR #2261, opened 2026-07-19
Short-circuit empty cookie store lookups

pavel-ptashyts:
### Summary
- Return an empty cookie list immediately when `ThreadSafeCookieStore` has no retained cookies.
- Keep null handling explicit with `requireNonNull(uri, "uri")` before the shortcut.
- Add coverage for `get(Uri)` on an empty store.

### Motivation
The default cookie store is consulted for each request. When no cookies have been stored, the lookup does not need to normalize the request host, derive the request path, or scan subdomains.

### Validation
- `/opt/maven/bin/mvn -pl client -DskipTests compile`
- `/opt/maven/bin/mvn -pl client -Dtest=ThreadSafeCookieStoreGetTest test`


pavel-ptashyts:
Closing because this performance change will not be pursued.

PR #2263, opened 2026-07-19
Use field updaters for timeout guards

pavel-ptashyts:
## Summary
- replace per-instance timeout AtomicBoolean guards with volatile int fields managed by static AtomicIntegerFieldUpdaters
- preserve existing run-once, reschedule, and cleanup semantics
- add unit coverage for idempotent timeout task cleanup

## Verification
- ./mvnw -pl client -DskipTests compile
- ./mvnw -pl client -Dtest=TimeoutTimerTaskTest,PerRequestTimeoutTest,NettyConnectListenerTest test

## Attribution


pavel-ptashyts:
Closing this for the 3.x line because Revapi correctly reports the removal of the protected TimeoutTimerTask.done field as an incompatible API change. Retaining the AtomicBoolean field would retain the allocation this change is intended to remove, while adding a Revapi exception would silently break external subclasses. This optimization can be reconsidered for a major release; the branch is retained as a reference.

PR #2264, opened 2026-07-19
Batch HTTP/2 upload flushes and prevent duplicate endStream

pavel-ptashyts:
## Summary
- batch HTTP/2 upload DATA frame flushes while the stream remains writable
- flush only when the body pump yields for source suspension/backpressure or writes the terminal DATA frame
- prevent synchronous writability callbacks from re-entering the pump
- keep the pump parked while the terminal DATA write is pending, preventing duplicate `endStream` frames
- cover multi-chunk batching, suspension, backpressure resume, re-entry, and cleanup timing in `Http2BodyWriterTest`

## Rationale
The HTTP/2 upload body pump previously flushed after every intermediate DATA chunk. That kept writes moving, but it also forced one flush per chunk even when the stream child channel stayed writable. Batching until a yield point preserves bounded in-flight data while reducing flush overhead for multi-chunk uploads.

Netty may synchronously fire `channelWritabilityChanged` from `flush()`. Explicit pumping and terminal-write states prevent that callback from mutating pump state recursively or emitting a second terminal frame before the first write completes.

## Validation
- `./mvnw -pl client -Dtest='org.asynchttpclient.netty.request.body.Http2BodyWriterTest' test` (3 tests)
- `./mvnw -pl client -DskipTests compile`
- `./mvnw -pl client -Dtest=Http2StreamingBodyFlowControlTest test`
- `./mvnw -pl client -Dtest=BasicHttp2Test test`

## Attribution

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
@@ -124,6 +124,11 @@ final class Http2BodyWriter {
     private ByteBuf pending;
     private boolean done;
 
+    // flush() can synchronously fire channelWritabilityChanged. Prevent that callback from re-entering the
+    // pump, and prevent later callbacks from writing a second endStream frame before the first one completes.
+    private boolean pumping;
+    private boolean terminalWritePending;
+
     // Transient handler that resumes the pump when the channel becomes writable again. Added lazily the
     // first time the pump parks, removed by finish().
     private WritabilityResumeHandler resumeHandler;
@@ -181,9 +186,10 @@ final class Http2BodyWriter {
      * event loop. Any written frames are flushed before the pump parks or completes.
      */
     private void pump() {
-        if (done) {
+        if (done || pumping || terminalWritePending) {
             return;
         }
+        pumping = true;
         try {
             while (true) {
                 if (done) {
@@ -211,6 +217,7 @@ final class Http2BodyWriter {
                     ByteBuf terminal = last != null ? last
                             // Empty body — preserve existing behaviour: a single empty DATA frame ends the stream.
                             : channel.alloc().buffer(0);
+                    terminalWritePending = true;
                     writeLastFrame(terminal);
                     channel.flush();
                     return;
@@ -238,6 +245,8 @@ final class Http2BodyWriter {
             }
         } catch (Throwable t) {
             finish(t);
+        } finally {
+            pumping = false;
         }
     }
 
@@ -314,7 +323,7 @@ final class Http2BodyWriter {
     private final class WritabilityResumeHandler extends ChannelInboundHandlerAdapter {
         @Override
         public void channelWritabilityChanged(ChannelHandlerContext ctx) {
-            if (!done && ctx.channel().isWritable()) {
+            if (!done && !terminalWritePending && ctx.channel().isWritable()) {
                 pump();
             }
             ctx.fireChannelWritabilityChanged();
--- a/client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
@@ -38,8 +38,9 @@ import java.io.IOException;
  * <p>
  * This writer is a one-chunk-at-a-time pump driven entirely on the stream channel's event loop:
  * <ul>
- *   <li>It produces and writes exactly one chunk, then flushes, so frames actually leave the process
- *       instead of accumulating.</li>
+ *   <li>It writes chunks while the stream remains writable, flushing when the pump yields for flow control,
+ *       a source suspension, or the terminal DATA frame, so frames leave the process without one flush per
+ *       chunk.</li>
  *   <li>It only produces the next chunk while {@link Http2StreamChannel#isWritable()} is {@code true}.
  *       A stream child channel becomes unwritable when the HTTP/2 flow-control window is exhausted or
  *       the local high-water mark is reached (child writes go through {@code incrementPendingOutboundBytes}).
@@ -176,7 +177,8 @@ final class Http2BodyWriter {
 
     /**
      * Produces and writes chunks until the channel goes unwritable (then parks for
-     * {@code channelWritabilityChanged}) or the body is exhausted. Always runs on the event loop.
+     * {@code channelWritabilityChanged}), the source suspends, or the body is exhausted. Always runs on the
+     * event loop. Any written frames are flushed before the pump parks or completes.
      */
     private void pump() {
         if (done) {
@@ -196,6 +198,7 @@ final class Http2BodyWriter {
                     // No data available yet but the body is not finished (feedable body). Park until the
                     // source signals more via the onResume callback. Any already-buffered `pending` chunk is
                     // retained (O(1)); we deliberately do not flush an early endStream.
+                    channel.flush();
                     suspended = true;
                     return;
                 }
@@ -219,9 +222,12 @@ final class Http2BodyWriter {
                 pending = next;
                 if (toWrite != null) {
                     writeFrame(toWrite, false);
-                    channel.flush();
 
                     if (!channel.isWritable()) {
+                        channel.flush();
+                        if (channel.isWritable()) {
+                            continue;
+                        }
                         // Flow-control window exhausted / high-water mark reached: stop producing and resume
                         // from channelWritabilityChanged. `pending` (one chunk) is retained until then.
                         ensureResumeHandler();
--- a/client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
@@ -50,8 +50,8 @@ import java.io.IOException;
  *       an empty body still sends a single empty DATA frame with {@code endStream=true}.</li>
  *   <li>This writer holds at most one buffered "pending" chunk across a writability wait. Production stops
  *       once the channel goes unwritable, so total in-flight heap is bounded by the channel's write
- *       high-water mark (the already-written chunks the channel's outbound buffer still owns) plus that one
- *       pending chunk — not by the size of the whole body.</li>
+ *       high-water mark (64 KiB by default; the already-written chunks the channel's outbound buffer still
+ *       owns) plus that one pending chunk, rather than by the size of the whole body.</li>
  * </ul>
  * <p>
  * <strong>Lifecycle / cleanup.</strong> Because the pump completes asynchronously (after {@code writeHttp2}
@@ -125,9 +125,10 @@ final class Http2BodyWriter {
     private boolean done;
 
     // flush() can synchronously fire channelWritabilityChanged. Prevent that callback from re-entering the
-    // pump, and prevent later callbacks from writing a second endStream frame before the first one completes.
+    // pump, and remember permanently once the terminal frame has been emitted so later callbacks cannot write
+    // a second endStream frame before its write completes.
     private boolean pumping;
-    private boolean terminalWritePending;
+    private boolean terminalWritten;
 
     // Transient handler that resumes the pump when the channel becomes writable again. Added lazily the
     // first time the pump parks, removed by finish().
@@ -186,7 +187,7 @@ final class Http2BodyWriter {
      * event loop. Any written frames are flushed before the pump parks or completes.
      */
     private void pump() {
-        if (done || pumping || terminalWritePending) {
+        if (done || pumping || terminalWritten) {
             return;
         }
         pumping = true;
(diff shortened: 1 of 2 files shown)


client/src/test/java/org/asynchttpclient/netty/request/body/Http2BodyWriterTest.java
```diff
@@ -0,0 +1,225 @@
...
+            ReferenceCountUtil.release(invocation.getArgument(0));
+            return succeededFuture(channel);
+        });
+
+        SuspendingChunkSource source = new SuspendingChunkSource();
+        Http2BodyWriter.start(channel, source);
+
+        assertEquals(0, source.closed);
+        verify(channel, times(1)).write(any(DefaultHttp2DataFrame.class));
+        verify(channel, times(1)).flush();
```
hyperxpro: I think this one is a no-op unfortunately. I ran it against main, where the SUSPEND branch doesn't flush at all, and it still passes. Both versions end up with exactly one flush by the time you get here, main's just comes from the flush after writeFrame, so times(1) can't separate them.

Could you assert the ordering instead? Either an InOrder over nextChunk and flush, or grab the flush count right when nextChunk hands back SUSPEND. Either one would pin the behaviour you're actually after.

client/src/test/java/org/asynchttpclient/netty/request/body/Http2BodyWriterTest.java
```diff
@@ -0,0 +1,225 @@
...
+public class Http2BodyWriterTest {
+
+    @Test
+    public void multiChunkBodyBatchesFlushUntilTerminalFrame() {
+        Http2StreamChannel channel = mock(Http2StreamChannel.class);
+        EventLoop eventLoop = mock(EventLoop.class);
+        when(eventLoop.inEventLoop()).thenReturn(true);
+        when(channel.eventLoop()).thenReturn(eventLoop);
+        when(channel.alloc()).thenReturn(UnpooledByteBufAllocator.DEFAULT);
+        when(channel.isWritable()).thenReturn(true);
```
hyperxpro, line 66: Since isWritable() is always true here the mock never goes unwritable no matter how much you write, so the flush counts below are really just checking the stub rather than anything Netty does. You'd get the same numbers for a 4 GiB body.

Would you mind adding one EmbeddedChannel test on top of Http2FrameCodecBuilder.forClient() plus Http2MultiplexHandler, checking that peak unflushed bytes stays under writeBufferHighWaterMark plus a chunk? That's the thing that would actually catch a regression here, and it lines up with how we test the rest of this area.

client/src/test/java/org/asynchttpclient/netty/request/body/Http2BodyWriterTest.java
```diff
@@ -0,0 +1,225 @@
...
+        ChannelHandlerContext pipelineContext = mock(ChannelHandlerContext.class);
+        ChannelHandlerContext eventContext = mock(ChannelHandlerContext.class);
+        AtomicReference<ChannelHandler> resumeHandler = new AtomicReference<>();
+        AtomicBoolean nestedWritabilityCallbackFired = new AtomicBoolean();
+        AtomicInteger terminalFrames = new AtomicInteger();
+        DefaultChannelPromise terminalWrite = new DefaultChannelPromise(channel, ImmediateEventExecutor.INSTANCE);
+        when(eventLoop.inEventLoop()).thenReturn(true);
+        when(channel.eventLoop()).thenReturn(eventLoop);
+        when(channel.alloc()).thenReturn(UnpooledByteBufAllocator.DEFAULT);
+        when(channel.isWritable()).thenReturn(false, false, true);
```
hyperxpro: Minor one, but this bakes in 'isWritable() gets called exactly three times, in this order', so it's tied to the call count rather than the behaviour. If someone later hoists isWritable() into a local the test quietly changes meaning instead of going red. An AtomicBoolean with thenAnswer would hold up better.

client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
```diff
@@ -196,6 +204,7 @@ private void pump() {
                     // No data available yet but the body is not finished (feedable body). Park until the
                     // source signals more via the onResume callback. Any already-buffered `pending` chunk is
                     // retained (O(1)); we deliberately do not flush an early endStream.
+                    channel.flush();
```
hyperxpro: One thing I noticed here. The flush now happens before suspended = true on the next line, and flush can run callbacks synchronously. If a resume ever lands in that gap, resumeFromSuspend sees suspended still false, quietly drops it, and the pump sits there until the request times out.

I couldn't actually trigger it, the only onResume wiring feeds from an app thread and bounces back onto the loop, so it's theoretical today. But it's a gap we didn't have before and it's cheap to close. Worth saying that just moving suspended = true above the flush doesn't do it, pumping would eat the re-entrant pump() instead. Simplest is probably a re-check after the flush (if (!suspended) continue;), or have the guard remember the request with a pumpRequested flag.

client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
```diff
@@ -123,6 +124,11 @@ default void onResume(Runnable resume) {
     private ByteBuf pending;
     private boolean done;
 
+    // flush() can synchronously fire channelWritabilityChanged. Prevent that callback from re-entering the
+    // pump, and prevent later callbacks from writing a second endStream frame before the first one completes.
+    private boolean pumping;
+    private boolean terminalWritePending;
```
hyperxpro: This never goes back to false, and finish() doesn't clear it either. Harmless in practice since done covers every path anyway, but the name threw me for a minute. It reads like 'a write is currently in flight' when it really means 'we've emitted the terminal frame, permanently'. Maybe terminalWritten? The comment above could follow suit.

client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
```diff
@@ -208,6 +217,7 @@ private void pump() {
                     ByteBuf terminal = last != null ? last
                             // Empty body — preserve existing behaviour: a single empty DATA frame ends the stream.
                             : channel.alloc().buffer(0);
+                    terminalWritePending = true;
```
hyperxpro: Good catch, though this is a genuine bug fix hiding inside a perf PR. On main, if channelWritabilityChanged lands after writeLastFrame but before that write finishes, pump() re-enters, nextChunk() returns null again, pending is null, and we send a second endStream DATA frame on a half closed stream. This flag stops it.

Could you mention it in the title or the description? Nobody reviewing 'batch flushes' is going to be looking for a conformance fix.

client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
```diff
@@ -38,8 +38,9 @@
  * <p>
  * This writer is a one-chunk-at-a-time pump driven entirely on the stream channel's event loop:
  * <ul>
- *   <li>It produces and writes exactly one chunk, then flushes, so frames actually leave the process
- *       instead of accumulating.</li>
+ *   <li>It writes chunks while the stream remains writable, flushing when the pump yields for flow control,
```
hyperxpro, line 41: Might be worth naming the new batching quantum while you're in here. Peak in flight heap per upload stream goes from roughly a chunk to the 64 KiB writeBufferHighWaterMark, I measured it landing at exactly 65536. The bound further down is still accurate, it's just moved, and anyone sizing concurrent uploads would want to know.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
pavel-ptashyts: Agreed. I documented that the write-buffer high-water mark is 64 KiB by default and that per-stream in-flight heap is bounded by that mark plus the single pending chunk. This is in 6c2178a20.

client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
pavel-ptashyts: Agreed. I reproduced the lost-resume path with a synchronous flush callback; with the old ordering the source remained parked and unclosed. The pump now marks itself suspended before flush and continues if that callback consumed the resume while re-entry was guarded. I added the regression test in 6c2178a20.

client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
pavel-ptashyts: Agreed. I renamed terminalWritePending to terminalWritten and updated the comment to make its permanent meaning explicit once the terminal frame has been emitted. This is in 6c2178a20.

client/src/test/java/org/asynchttpclient/netty/request/body/Http2BodyWriterTest.java
pavel-ptashyts: Agreed. I replaced the sequential isWritable() return values with an AtomicBoolean-backed answer and flip the state only before firing the explicit writability callback. This is in 6c2178a20.

client/src/test/java/org/asynchttpclient/netty/request/body/Http2BodyWriterTest.java
pavel-ptashyts: Agreed. I added a real client Http2FrameCodec plus Http2MultiplexHandler stream test. It drives the stream past a configured 8 KiB high-water mark and asserts that peak unflushed DATA stays between the mark and the mark plus one chunk. This is in 6c2178a20.

client/src/main/java/org/asynchttpclient/netty/request/body/Http2BodyWriter.java
pavel-ptashyts: Agreed. I updated the PR title to "Batch HTTP/2 upload flushes and prevent duplicate endStream" so the conformance fix is visible alongside the batching change.

client/src/test/java/org/asynchttpclient/netty/request/body/Http2BodyWriterTest.java
pavel-ptashyts: Agreed. I replaced the flush-count-only assertion with an ordering-sensitive check: the source records the flush count when it returns SUSPEND, so the test now proves there was no earlier per-chunk flush and that the suspension path flushes before parking. This is in 6c2178a20.

client/src/test/java/org/asynchttpclient/netty/request/body/Http2BodyWriterTest.java
```diff
@@ -0,0 +1,348 @@
...
+            }
+            ReferenceCountUtil.release(frame);
+            return succeededFuture(channel);
+        });
+        when(channel.flush()).thenAnswer(invocation -> {
+            ChannelHandler handler = resumeHandler.get();
+            if (handler != null
+                    && terminalFrames.get() != 0
+                    && nestedWritabilityCallbackFired.compareAndSet(false, true)) {
+                ((io.netty.channel.ChannelInboundHandlerAdapter) handler).channelWritabilityChanged(eventContext);
```
hyperxpro: You imported ChannelInboundHandlerAdapter for the multiplex handler, so this doesn't need qualifying any more.

```suggestion
                ((ChannelInboundHandlerAdapter) handler).channelWritabilityChanged(eventContext);
```

client/src/test/java/org/asynchttpclient/netty/request/body/Http2BodyWriterTest.java
```diff
@@ -0,0 +1,348 @@
...
+        Http2BodyWriter.start(channel, source);
+
+        assertEquals(0, source.closed);
+        assertNotNull(resumeHandler.get());
+        verify(channel, times(1)).write(any(DefaultHttp2DataFrame.class));
+        verify(channel, times(1)).flush();
+
+        writable.set(true);
+        ((io.netty.channel.ChannelInboundHandlerAdapter) resumeHandler.get())
+                .channelWritabilityChanged(eventContext);
```
hyperxpro: Same here, and it fits on one line once the package prefix goes.

```suggestion
        ((ChannelInboundHandlerAdapter) resumeHandler.get()).channelWritabilityChanged(eventContext);
```

hyperxpro (changes requested):

client/src/test/java/org/asynchttpclient/netty/request/body/Http2BodyWriterTest.java
pavel-ptashyts: Agreed, thanks. I verified the applied suggestion and the cast now uses the existing ChannelInboundHandlerAdapter import. The focused test and the full JDK 11 build both pass.

client/src/test/java/org/asynchttpclient/netty/request/body/Http2BodyWriterTest.java
pavel-ptashyts: Agreed, thanks. I verified the applied cleanup here as well; the callback invocation now fits on one line. The focused test and the full JDK 11 build both pass.

hyperxpro (approved):

PR #2266, opened 2026-07-19
Intern common Train-Case header names

pavel-ptashyts:
## Summary
- add prebuilt Train-Case AsciiString entries for common request header names
- keep lowercase Netty header constants in the intern table
- update the focused interning tests to cover Train-Case interning and odd-casing passthrough

## Rationale
NettyRequestFactory already interns known lowercase header names so HttpHeadersEncoder can bulk-copy AsciiString bytes on the event loop. Real request code often supplies names such as Content-Type and Authorization in Train-Case, which previously missed the table and fell back to per-character encoding. Adding same-spelling AsciiString entries preserves wire casing while putting those common names on the same fast path.

## Validation
- ./mvnw -pl client -Dtest=NettyRequestFactoryHeaderInterningTest test
- ./mvnw -pl client -DskipTests compile
- ./mvnw -pl client -Dtest=BasicHttpTest,BasicHttp2Test,AutomaticDecompressionTest test

## Attribution

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
@@ -85,18 +85,16 @@ public final class NettyRequestFactory {
     private static final Integer ZERO_CONTENT_LENGTH = 0;
 
     /**
-     * Canonical (lowercase) header-name spelling -> the pre-built Netty {@link AsciiString} constant.
-     * Built once. When a user-supplied {@code String} name is byte-identical to a canonical spelling, the
-     * outbound header uses the shared {@code AsciiString} so {@code HttpHeadersEncoder} hits its bulk
-     * array-copy branch instead of the per-char US-ASCII encode loop on the event loop. The map is keyed
-     * case-sensitively, so a name that is not byte-identical to its canonical spelling (e.g. Train-Case
-     * {@code Content-Type}) misses and is emitted verbatim — on-wire bytes are never altered, and a custom
-     * name costs a single {@link HashMap#get} miss with zero allocation.
+     * Known header-name spelling -> a pre-built {@link AsciiString} with the same bytes. Built once. When a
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
```diff
@@ -112,21 +110,43 @@ private static Map<String, AsciiString> buildKnownHeaderNames() {
...
+        }
+        for (AsciiString name : trainCaseNames) {
+            putKnownHeaderName(map, name);
         }
         return map;
     }
 
+    private static void putKnownHeaderName(Map<String, AsciiString> map, AsciiString name) {
+        map.put(name.toString(), name);
+    }
```
hyperxpro:
```suggestion
```

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestFactory.java
```diff
@@ -112,21 +110,43 @@ private static Map<String, AsciiString> buildKnownHeaderNames() {
...
+                new AsciiString("Transfer-Encoding"),
+                new AsciiString("User-Agent")
+        };
+        Map<String, AsciiString> map = new HashMap<>(
+                (int) ((lowerCaseNames.length + trainCaseNames.length) / 0.75f) + 1);
+        for (AsciiString name : lowerCaseNames) {
+            putKnownHeaderName(map, name);
+        }
+        for (AsciiString name : trainCaseNames) {
+            putKnownHeaderName(map, name);
```
hyperxpro:
```suggestion
        Map<String, AsciiString> map = new HashMap<>((int) (names.length / 0.75f) + 1);
        for (AsciiString name : names) {
            map.put(name.toString(), name);
```

hyperxpro (changes requested):

hyperxpro (approved):

PR #2271, opened 2026-07-20
Use HPACK spelling for HTTP/2 compression

pavel-ptashyts:
## Summary

- rewrite only AHC's generated default `Accept-Encoding` value from `gzip,deflate` to `gzip, deflate` while copying headers into HTTP/2 frames
- allow HPACK to use static-table entry 16 for the generated value
- preserve user-supplied `Accept-Encoding` spelling and leave HTTP/1 output unchanged
- add HTTP/2 behavioral coverage and a deterministic encoded-size test
- correct the existing JMH benchmark documentation about its returned value

## Performance evidence

With Netty 4.2.15.Final and a fresh HPACK encoder:

- `gzip,deflate`: 14 encoded bytes
- `gzip, deflate`: 1 encoded byte through the indexed static-table representation

A short JMH validation on JDK 21 with `-prof gc` measured 1032 B/op for the literal spelling and 680 B/op for the static-table spelling. Timing was noisy, so this PR relies on the deterministic wire-size test rather than claiming a stable CPU improvement.

## Validation

- `./mvnw -pl client -Dtest='org.asynchttpclient.netty.request.AcceptEncodingHpackTest,org.asynchttpclient.BasicHttp2Test#generatedAcceptEncodingUsesHpackStaticValueOverHttp2+userAcceptEncodingSpellingIsPreservedOverHttp2' test` (3 tests)
- `AcceptEncodingHpackBenchmark -f 1 -wi 3 -i 5 -w 500ms -r 500ms -prof gc`
- existing full `BasicHttp2Test` validation from the initial change

## Attribution

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
@@ -106,6 +106,7 @@ import static org.asynchttpclient.util.AuthenticatorUtils.perConnectionAuthoriza
 import static org.asynchttpclient.util.AuthenticatorUtils.perConnectionProxyAuthorizationHeader;
 import static org.asynchttpclient.util.HttpConstants.Methods.CONNECT;
 import static org.asynchttpclient.util.HttpConstants.Methods.GET;
+import static org.asynchttpclient.util.HttpUtils.GZIP_DEFLATE;
 import static org.asynchttpclient.util.HttpUtils.hostHeader;
 import static org.asynchttpclient.util.MiscUtils.getCause;
 import static org.asynchttpclient.util.ProxyUtils.getProxyServer;
@@ -113,6 +114,8 @@ import static org.asynchttpclient.util.ProxyUtils.getProxyServer;
 public final class NettyRequestSender {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NettyRequestSender.class);
+    private static final AsciiString HTTP2_HPACK_GZIP_DEFLATE =
+            new AsciiString(HttpHeaderValues.GZIP + ", " + HttpHeaderValues.DEFLATE);
 
     private final AsyncHttpClientConfig config;
     private final ChannelManager channelManager;
@@ -918,6 +921,7 @@ public final class NettyRequestSender {
             // Copy the HTTP/1.1 headers, dropping connection-specific names forbidden in HTTP/2 (RFC 7540
             // §8.1.2.2). iteratorCharSequence() avoids the per-name String the String-typed iterator forces;
             // see isHttp2ExcludedHeader and toLowerCaseHeaderName for the skip-check and lowercasing rules.
+            boolean preferHpackAcceptEncoding = preferHpackAcceptEncoding(future.getCurrentRequest());
             Iterator<Map.Entry<CharSequence, CharSequence>> it = httpRequest.headers().iteratorCharSequence();
             while (it.hasNext()) {
                 Map.Entry<CharSequence, CharSequence> entry = it.next();
@@ -933,7 +937,7 @@ public final class NettyRequestSender {
                         && !HttpHeaderValues.TRAILERS.contentEqualsIgnoreCase(value)) {
                     continue;
                 }
-                h2Headers.add(toLowerCaseHeaderName(name), value);
+                h2Headers.add(toLowerCaseHeaderName(name), http2HeaderValue(name, value, preferHpackAcceptEncoding));
             }
 
             // Determine the body to send: an in-memory buffer (DefaultFullHttpRequest content or a
@@ -969,6 +973,20 @@ public final class NettyRequestSender {
                 nettyRequest.release();
             }
         }
+    }
+
+    private boolean preferHpackAcceptEncoding(Request request) {
+        return config.isCompressionEnforced()
+                && !request.getHeaders().contains(HttpHeaderNames.ACCEPT_ENCODING);
+    }
+
+    private static CharSequence http2HeaderValue(CharSequence name, CharSequence value, boolean preferHpackAcceptEncoding) {
+        if (preferHpackAcceptEncoding
+                && HttpHeaderNames.ACCEPT_ENCODING.contentEqualsIgnoreCase(name)
+                && GZIP_DEFLATE.contentEquals(value)) {
+            return HTTP2_HPACK_GZIP_DEFLATE;
+        }
+        return value;
     }
 
     /**
--- a/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
+++ b/client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
@@ -107,6 +107,7 @@ import static org.asynchttpclient.util.AuthenticatorUtils.perConnectionProxyAuth
 import static org.asynchttpclient.util.HttpConstants.Methods.CONNECT;
 import static org.asynchttpclient.util.HttpConstants.Methods.GET;
 import static org.asynchttpclient.util.HttpUtils.GZIP_DEFLATE;
+import static org.asynchttpclient.util.HttpUtils.GZIP_DEFLATE_HPACK;
 import static org.asynchttpclient.util.HttpUtils.hostHeader;
 import static org.asynchttpclient.util.MiscUtils.getCause;
 import static org.asynchttpclient.util.ProxyUtils.getProxyServer;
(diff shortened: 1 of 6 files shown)


client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -971,6 +975,20 @@ private <T> void sendHttp2Frames(NettyResponseFuture<T> future, Http2StreamChann
...
 
+    private boolean preferHpackAcceptEncoding(Request request) {
+        return config.isCompressionEnforced()
+                && !request.getHeaders().contains(HttpHeaderNames.ACCEPT_ENCODING);
+    }
+
+    private static CharSequence http2HeaderValue(CharSequence name, CharSequence value, boolean preferHpackAcceptEncoding) {
+        if (preferHpackAcceptEncoding
+                && HttpHeaderNames.ACCEPT_ENCODING.contentEqualsIgnoreCase(name)
+                && GZIP_DEFLATE.contentEquals(value)) {
```
hyperxpro: Since GZIP_DEFLATE is a singleton AsciiString and NettyRequestFactory stores that exact reference when it calls headers.set(ACCEPT_ENCODING, GZIP_DEFLATE), you can check value == GZIP_DEFLATE here instead of contentEquals. I checked and the reference does survive into the iterator, and a user typed string with the same text is never the same reference, so identity alone tells you whether this is the auto generated header without needing the separate preferHpackAcceptEncoding check at all. That would let you delete preferHpackAcceptEncoding and the getCurrentRequest call above, which removes a second read of state that has to stay in sync with what actually built the headers.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -918,6 +921,7 @@ private <T> void sendHttp2Frames(NettyResponseFuture<T> future, Http2StreamChann
             // Copy the HTTP/1.1 headers, dropping connection-specific names forbidden in HTTP/2 (RFC 7540
             // §8.1.2.2). iteratorCharSequence() avoids the per-name String the String-typed iterator forces;
             // see isHttp2ExcludedHeader and toLowerCaseHeaderName for the skip-check and lowercasing rules.
+            boolean preferHpackAcceptEncoding = preferHpackAcceptEncoding(future.getCurrentRequest());
```
hyperxpro: This reads the policy from future.getCurrentRequest() while the header values a few lines below come from future.getNettyRequest(). They happen to always be set together today in newNettyRequestAndResponseFuture, so this works, but it is two different reads off the same future that have to stay aligned. If you switch to the identity check on GZIP_DEFLATE mentioned below this whole method and this line can go away.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -106,13 +106,16 @@
...
 import static org.asynchttpclient.util.HttpConstants.Methods.GET;
+import static org.asynchttpclient.util.HttpUtils.GZIP_DEFLATE;
 import static org.asynchttpclient.util.HttpUtils.hostHeader;
 import static org.asynchttpclient.util.MiscUtils.getCause;
 import static org.asynchttpclient.util.ProxyUtils.getProxyServer;
 
 public final class NettyRequestSender {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NettyRequestSender.class);
+    private static final AsciiString HTTP2_HPACK_GZIP_DEFLATE =
```
hyperxpro: Consider moving this next to HttpUtils.GZIP_DEFLATE in HttpUtils rather than keeping it here. They are really the same concept, the HPACK static table spelling of the same value, and keeping them in separate files makes it easier for someone to change one and forget the other later. A short comment noting this is RFC 7541 appendix A static table entry 16 would also help since that context is currently only in the benchmark javadoc, not on the field itself.

client/src/test/java/org/asynchttpclient/BasicHttp2Test.java
```diff
@@ -1102,6 +1049,31 @@ public void mixedCaseHeaderIsLowercasedAndConnectionHeadersExcludedOverHttp2() t
...
 
+    @Test
+    public void generatedAcceptEncodingUsesHpackStaticValueOverHttp2() throws Exception {
+        try (AsyncHttpClient client = http2ClientWithConfig(builder -> builder.setCompressionEnforced(true))) {
+            Response response = client.prepareGet(httpsUrl("/echo"))
+                    .execute()
+                    .get(30, SECONDS);
+
+            assertEquals(200, response.getStatusCode());
+            assertEquals("gzip, deflate", response.getHeader("X-accept-encoding"));
```
hyperxpro: `response.getHeader` only returns the first value for a repeated header. When compressionEnforced is true and Brotli or Zstd are available on the classpath, AHC actually sends three separate Accept-Encoding entries, not one comma joined value. Worth adding a test that runs with those codecs available and asserts on response.getHeaders("X-accept-encoding") for all values, so a rewrite that accidentally touched or duplicated the br or zstd entries would actually get caught.

client/src/test/java/org/asynchttpclient/BasicHttp2Test.java
```diff
@@ -1102,6 +1049,31 @@ public void mixedCaseHeaderIsLowercasedAndConnectionHeadersExcludedOverHttp2() t
...
+        try (AsyncHttpClient client = http2ClientWithConfig(builder -> builder.setCompressionEnforced(true))) {
+            Response response = client.prepareGet(httpsUrl("/echo"))
+                    .setHeader(ACCEPT_ENCODING, "gzip,deflate")
+                    .execute()
+                    .get(30, SECONDS);
+
+            assertEquals(200, response.getStatusCode());
+            assertEquals("gzip,deflate", response.getHeader("X-accept-encoding"));
+        }
+    }
```
hyperxpro, line 1085: These two tests only cover HTTP/2. Since the whole point of the change is that HTTP/1 output stays as gzip,deflate with no space, it would be worth adding a small HTTP/1 test with compressionEnforced true asserting the header is still sent without the space, so that contract is pinned instead of just asserted in the PR description.

client/src/test/java/org/asynchttpclient/netty/request/AcceptEncodingHpackTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private static final AsciiString ACCEPT_ENCODING = AsciiString.cached("accept-encoding");
+
+    @Test
+    public void staticTableSpellingUsesOneByteIndexedRepresentation() throws Exception {
+        assertTrue(encodedLength("gzip,deflate") > 1);
+        assertEquals(1, encodedLength("gzip, deflate"));
+    }
+
+    private static int encodedLength(String value) throws Exception {
```
hyperxpro, line 41: This test builds headers by hand and drives Netty's encoder directly, it does not go through AHC's own request building or NettyRequestSender at all. It is a good proof that the two spellings differ in encoded size, but it is not proof that AHC actually produces the gzip, deflate spelling in production. That part is only covered by the two BasicHttp2Test cases. Might be worth a one line comment on the class saying that explicitly so nobody later assumes this test exercises the real code path.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Moved the HPACK spelling to HttpUtils next to GZIP_DEFLATE and documented RFC 7541 Appendix A entry 16.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Updated. The rewrite now uses value == GZIP_DEFLATE, relying on the interned generated value that survives the header iterator.

client/src/test/java/org/asynchttpclient/BasicHttp2Test.java
pavel-ptashyts: Added an HTTP/1 integration test that verifies the generated first value remains gzip,deflate.

client/src/test/java/org/asynchttpclient/BasicHttp2Test.java
pavel-ptashyts: Expanded the HTTP/2 integration assertion to verify all generated Accept-Encoding values, including br and zstd when their codecs are available.

client/src/test/java/org/asynchttpclient/netty/request/AcceptEncodingHpackTest.java
pavel-ptashyts: Clarified the test scope: this class verifies Netty HPACK wire size, while BasicHttp2Test covers AHC's production request path.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Removed the request/config lookup. The generated-value identity is now the only marker, so HTTP/2 conversion no longer reads currentRequest.

hyperxpro (approved):

PR #2272, opened 2026-07-20
Offload fallback DNS from event loops

pavel-ptashyts:
## Summary
- offload AHC's default blocking fallback `NameResolver` when resolution starts on a Netty event loop
- cover both redirect replay and SOCKS proxy bootstrap fallback resolution
- add client settings to enable or disable offload and tune its worker count and bounded queue size
- track pending resolutions so saturation and client shutdown complete them exceptionally instead of hanging
- preserve the existing caller-thread contract for custom `NameResolver` implementations
- document that the public resolver support types exist only for cross-package client wiring

## Configuration
- `fallbackNameResolverOffloadEnabled` defaults to `false` and must be enabled explicitly
- `fallbackNameResolverOffloadThreadsCount` defaults to `-1`, which uses the configured I/O thread count
- `fallbackNameResolverOffloadQueueSize` defaults to `0`, which selects an automatic bounded capacity

## Motivation
The legacy fallback path uses Netty's JDK-backed `DefaultNameResolver`. Redirect replays and SOCKS proxy bootstrap can enter that path from Netty event loops, so a cold or slow OS DNS lookup can block the loop. `AddressResolverGroup` users remain on the existing asynchronous path, off-loop calls retain their current behavior, and custom resolvers are not moved to an unexpected thread.

A bounded shared pool prevents an unbounded request backlog. Pending promises are tracked and failed when the pool rejects work or the client closes, including tasks removed from the queue by shutdown.

The resolver support classes are public only because client wiring crosses package boundaries. Applications should enable and tune the feature through `AsyncHttpClientConfig`.

## Validation
- `./mvnw -pl client -am -Dtest='AddressResolverGroupTest#defaultSocksProxyResolverUsesOffloadPool' -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw -pl client -Dtest='AddressResolverGroupTest#defaultNameResolverOnRedirectUsesOffloadPool+defaultNameResolverOnRedirectKeepsInlineBehavior,AsyncHttpClientDefaultsTest#testDefaultFallbackNameResolverOffloadEnabled,NameResolverOffloadTest' test`
- `./mvnw -pl client -DskipTests -Dgpg.skip=true verify`

## Attribution


pavel-ptashyts:
Closing by decision; performance-plan item 1.1 will be marked skipped.

PR #2275, opened 2026-07-20
Validate converted HTTP/2 headers and trailers

pavel-ptashyts:
## Summary

* retain complete HTTP token and value validation when converting HTTP/2 response headers and trailers
* drop the trailer-prohibited fields (Content-Length, Transfer-Encoding, Trailer) instead of rejecting them
* centralize pseudo-header filtering without a per-frame capturing lambda
* cover malformed names and values, empty names, and prohibited trailer fields

## Motivation

The original implementation assumed the HTTP/2 decoder had already performed complete header-name validation. Netty 4.2.15 validates HTTP/2 casing and pseudo-header rules, but it does not enforce the complete HTTP token syntax, and decoded values are not validated by default. Disabling name validation therefore removed the only complete check before headers reached user code.

This revision keeps Netty's normal header validation on both the response and trailer paths, so names are token-checked and values are CR/LF-checked exactly as they were before.

Trailers additionally drop the fields that cannot be processed after the content, matching HttpObjectDecoder.isPermittedTrailingHeader. Dropping rather than rejecting is deliberate. RFC 9110 section 6.5.1 puts the constraint on the sender, and the only recipient-level requirement in that section concerns merging trailers into the header section, which this code does not do. Rejecting would also fail a response whose body has already been delivered to the AsyncHandler, and would make HTTP/2 behave differently from HTTP/1 for the same server response, since HttpObjectDecoder drops these fields silently.

The shared enhanced-for copy path avoids a capturing lambda allocation for each response and trailer frame.

## Validation

* JDK 11 focused Http2HandlerTest: 5 tests passed
* JDK 11 `./mvnw clean verify`: build, tests, Javadocs, signing, and Revapi passed

## Attribution

Codex on behalf of Pavel Ptashyt

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
@@ -19,9 +19,10 @@ import io.netty.buffer.ByteBuf;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandler.Sharable;
 import io.netty.channel.ChannelHandlerContext;
-import io.netty.handler.codec.http.DefaultHttpHeaders;
+import io.netty.handler.codec.http.DefaultHttpHeadersFactory;
 import io.netty.handler.codec.http.DefaultHttpResponse;
 import io.netty.handler.codec.http.HttpHeaders;
+import io.netty.handler.codec.http.HttpHeadersFactory;
 import io.netty.handler.codec.http.HttpResponse;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.HttpVersion;
@@ -56,6 +57,8 @@ import java.io.IOException;
 public final class Http2Handler extends AsyncHttpClientHandler {
 
     private static final HttpVersion HTTP_2 = new HttpVersion("HTTP", 2, 0, true);
+    private static final HttpHeadersFactory RESPONSE_HEADERS_FACTORY =
+            DefaultHttpHeadersFactory.headersFactory().withNameValidation(false);
 
     public Http2Handler(AsyncHttpClientConfig config, ChannelManager channelManager, NettyRequestSender requestSender) {
         super(config, channelManager, requestSender);
@@ -146,14 +149,7 @@ public final class Http2Handler extends AsyncHttpClientHandler {
         }
         HttpResponseStatus nettyStatus = HttpResponseStatus.valueOf(statusCode);
 
-        // Build HTTP/1.1-style headers, skipping HTTP/2 pseudo-headers (start with ':')
-        HttpHeaders responseHeaders = new DefaultHttpHeaders();
-        h2Headers.forEach(entry -> {
-            CharSequence name = entry.getKey();
-            if (name.length() > 0 && name.charAt(0) != ':') {
-                responseHeaders.add(name, entry.getValue());
-            }
-        });
+        HttpHeaders responseHeaders = copyHttp2Headers(h2Headers);
 
         // Build a synthetic HttpResponse so the existing interceptor chain can be reused unchanged
         HttpResponse syntheticResponse = new DefaultHttpResponse(HTTP_2, nettyStatus, responseHeaders);
@@ -222,13 +218,7 @@ public final class Http2Handler extends AsyncHttpClientHandler {
                                                   NettyResponseFuture<?> future, AsyncHandler<?> handler) throws Exception {
         Http2Headers h2Headers = headersFrame.headers();
 
-        HttpHeaders trailingHeaders = new DefaultHttpHeaders();
-        h2Headers.forEach(entry -> {
-            CharSequence name = entry.getKey();
-            if (name.length() > 0 && name.charAt(0) != ':') {
-                trailingHeaders.add(name, entry.getValue());
-            }
-        });
+        HttpHeaders trailingHeaders = copyHttp2Headers(h2Headers);
 
         boolean abort = false;
         if (!trailingHeaders.isEmpty()) {
@@ -238,6 +228,19 @@ public final class Http2Handler extends AsyncHttpClientHandler {
         if (abort || headersFrame.isEndStream()) {
             finishUpdate(future, channel, false);
         }
+    }
+
+    static HttpHeaders copyHttp2Headers(Http2Headers h2Headers) {
+        // The HTTP/2 decoder validates names but not values by default. Avoid repeating the name scan while
+        // retaining the HTTP/1 value validator that rejects CR/LF and other prohibited characters.
+        HttpHeaders headers = RESPONSE_HEADERS_FACTORY.newHeaders();
+        h2Headers.forEach(entry -> {
+            CharSequence name = entry.getKey();
+            if (name.length() > 0 && name.charAt(0) != ':') {
+                headers.add(name, entry.getValue());
+            }
+        });
+        return headers;
     }
 
     /**
--- a/client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
@@ -42,6 +42,7 @@ import org.asynchttpclient.netty.channel.Channels;
 import org.asynchttpclient.netty.request.NettyRequestSender;
 
 import java.io.IOException;
+import java.util.Map;
 
 /**
  * HTTP/2 channel handler for stream child channels created by {@link io.netty.handler.codec.http2.Http2MultiplexHandler}.
@@ -57,8 +58,8 @@ import java.io.IOException;
 public final class Http2Handler extends AsyncHttpClientHandler {
 
     private static final HttpVersion HTTP_2 = new HttpVersion("HTTP", 2, 0, true);
-    private static final HttpHeadersFactory RESPONSE_HEADERS_FACTORY =
-            DefaultHttpHeadersFactory.headersFactory().withNameValidation(false);
+    private static final HttpHeadersFactory RESPONSE_HEADERS_FACTORY = DefaultHttpHeadersFactory.headersFactory();
+    private static final HttpHeadersFactory TRAILING_HEADERS_FACTORY = DefaultHttpHeadersFactory.trailersFactory();
 
     public Http2Handler(AsyncHttpClientConfig config, ChannelManager channelManager, NettyRequestSender requestSender) {
         super(config, channelManager, requestSender);
@@ -218,7 +219,7 @@ public final class Http2Handler extends AsyncHttpClientHandler {
                                                   NettyResponseFuture<?> future, AsyncHandler<?> handler) throws Exception {
         Http2Headers h2Headers = headersFrame.headers();
 
-        HttpHeaders trailingHeaders = copyHttp2Headers(h2Headers);
+        HttpHeaders trailingHeaders = copyHttp2Trailers(h2Headers);
 
         boolean abort = false;
         if (!trailingHeaders.isEmpty()) {
@@ -231,15 +232,22 @@ public final class Http2Handler extends AsyncHttpClientHandler {
     }
 
     static HttpHeaders copyHttp2Headers(Http2Headers h2Headers) {
-        // The HTTP/2 decoder validates names but not values by default. Avoid repeating the name scan while
-        // retaining the HTTP/1 value validator that rejects CR/LF and other prohibited characters.
-        HttpHeaders headers = RESPONSE_HEADERS_FACTORY.newHeaders();
-        h2Headers.forEach(entry -> {
+        return copyHttp2Headers(h2Headers, RESPONSE_HEADERS_FACTORY);
+    }
+
+    static HttpHeaders copyHttp2Trailers(Http2Headers h2Headers) {
+        return copyHttp2Headers(h2Headers, TRAILING_HEADERS_FACTORY);
+    }
+
+    private static HttpHeaders copyHttp2Headers(Http2Headers h2Headers, HttpHeadersFactory headersFactory) {
+        // HTTP/2 validation does not enforce HTTP token syntax or header values, so validate both while converting.
+        HttpHeaders headers = headersFactory.newHeaders();
+        for (Map.Entry<CharSequence, CharSequence> entry : h2Headers) {
             CharSequence name = entry.getKey();
             if (name.length() > 0 && name.charAt(0) != ':') {
                 headers.add(name, entry.getValue());
             }
-        });
+        }
         return headers;
     }
 
--- a/client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
@@ -21,6 +21,7 @@ import io.netty.channel.ChannelHandler.Sharable;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.http.DefaultHttpHeadersFactory;
 import io.netty.handler.codec.http.DefaultHttpResponse;
+import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpHeadersFactory;
 import io.netty.handler.codec.http.HttpResponse;
@@ -58,8 +59,7 @@ import java.util.Map;
 public final class Http2Handler extends AsyncHttpClientHandler {
 
     private static final HttpVersion HTTP_2 = new HttpVersion("HTTP", 2, 0, true);
-    private static final HttpHeadersFactory RESPONSE_HEADERS_FACTORY = DefaultHttpHeadersFactory.headersFactory();
-    private static final HttpHeadersFactory TRAILING_HEADERS_FACTORY = DefaultHttpHeadersFactory.trailersFactory();
+    private static final HttpHeadersFactory HEADERS_FACTORY = DefaultHttpHeadersFactory.headersFactory();
 
     public Http2Handler(AsyncHttpClientConfig config, ChannelManager channelManager, NettyRequestSender requestSender) {
         super(config, channelManager, requestSender);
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
```diff
@@ -56,6 +57,8 @@
 public final class Http2Handler extends AsyncHttpClientHandler {
 
     private static final HttpVersion HTTP_2 = new HttpVersion("HTTP", 2, 0, true);
+    private static final HttpHeadersFactory RESPONSE_HEADERS_FACTORY =
```
hyperxpro: This is the core issue. Turning off name validation here is not removing a duplicate check, it is removing the only check. Netty's HTTP/2 name validation rejects uppercase and nothing else, so everything the HTTP/1 token validator was catching now passes straight through into the headers we hand to user code.

I suggest dropping withNameValidation(false) and using DefaultHttpHeadersFactory.headersFactory() as is, or just going back to new DefaultHttpHeaders(). Keeping the factory as a hoisted static is fine and thread safe under Sharable, so that part can stay if you prefer it.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
```diff
@@ -240,6 +230,19 @@ private void handleHttp2TrailingHeadersFrame(Http2HeadersFrame headersFrame, Cha
         }
     }
 
+    static HttpHeaders copyHttp2Headers(Http2Headers h2Headers) {
+        // The HTTP/2 decoder validates names but not values by default. Avoid repeating the name scan while
```
hyperxpro: This comment is the part I would most want fixed even after the code is sorted out. The second clause is correct, the decoder genuinely does not validate values. The first clause is not, and left in the tree it will convince the next person reading this method that the copy is safe when it is not. Comments that assert an invariant tend to outlive the reasoning behind them, so it is worth rewriting rather than deleting.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
```diff
@@ -222,13 +218,7 @@ private void handleHttp2TrailingHeadersFrame(Http2HeadersFrame headersFrame, Cha
...
         Http2Headers h2Headers = headersFrame.headers();
 
-        HttpHeaders trailingHeaders = new DefaultHttpHeaders();
-        h2Headers.forEach(entry -> {
-            CharSequence name = entry.getKey();
-            if (name.length() > 0 && name.charAt(0) != ':') {
-                trailingHeaders.add(name, entry.getValue());
-            }
-        });
+        HttpHeaders trailingHeaders = copyHttp2Headers(h2Headers);
```
hyperxpro: Not introduced by this PR, but this is the natural moment to raise it since you are centralising the code. The trailers path uses the same response factory, so it never gets Netty's trailer specific restrictions. trailersFactory() additionally rejects Content-Length, Transfer-Encoding and Trailer in trailers, per RFC 9110 section 6.5.1, and I confirmed those three are rejected by that factory and accepted by the one used here.

Worth knowing that the H2 layer does not cover this either. I checked HttpHeaderValidationUtil.isConnectionHeader and it returns true for transfer-encoding, connection, te, keep-alive and upgrade, but false for content-length. So nothing anywhere currently rejects Content-Length in an HTTP/2 trailer. Happy for this to be a follow up rather than scope creep here, but note that trailersFactory().withNameValidation(false) would not help, since that flag disables the trailer restrictions too. The flag has to go first.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
```diff
@@ -240,6 +230,19 @@ private void handleHttp2TrailingHeadersFrame(Http2HeadersFrame headersFrame, Cha
         }
     }
 
+    static HttpHeaders copyHttp2Headers(Http2Headers h2Headers) {
+        // The HTTP/2 decoder validates names but not values by default. Avoid repeating the name scan while
+        // retaining the HTTP/1 value validator that rejects CR/LF and other prohibited characters.
+        HttpHeaders headers = RESPONSE_HEADERS_FACTORY.newHeaders();
+        h2Headers.forEach(entry -> {
```
hyperxpro: Small aside while this loop is being touched. The lambda captures headers, so it cannot be cached and a fresh instance is allocated per response and again per trailer frame. A plain enhanced for loop over the entries removes that allocation at no cost to safety, which is a cleaner win than the one this PR is going for.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
```diff
@@ -240,6 +230,19 @@ private void handleHttp2TrailingHeadersFrame(Http2HeadersFrame headersFrame, Cha
...
+            CharSequence name = entry.getKey();
+            if (name.length() > 0 && name.charAt(0) != ':') {
```
hyperxpro: Worth being aware that this guard is now the only thing stopping empty names. Previously the default validator rejected them as well. With name validation off the replacement is a not null check, so an empty name would be accepted if it ever reached the add call. Behaviour is unchanged today because this line filters them, but the backstop is gone and nothing tests this line.

client/src/test/java/org/asynchttpclient/netty/handler/Http2HandlerTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class Http2HandlerTest {
+
+    @Test
+    void copiesDecoderValidatedHeaderNamesWithoutRevalidation() {
```
hyperxpro: This test pins the regression as intended behaviour, and the method name states the claim that does not hold. The name used here is not decoder validated. Passing false to DefaultHttp2Headers looks like it is there to construct something the decoder would have rejected, but DefaultHttp2Headers with validation on accepts this name too, because it contains no uppercase. A space is in the forbidden range in RFC 9113 section 8.2.1.

I suggest inverting it, so it asserts that a name containing CRLF is rejected. That is the property worth locking down, and it is the one that regressed. If it helps I have a reproduction that goes through the real encoder and decoder rather than constructing headers by hand, and I am happy to contribute it.

client/src/test/java/org/asynchttpclient/netty/handler/Http2HandlerTest.java
```diff
@@ -0,0 +1,48 @@
...
+                .add("invalid name", "value");
+
+        HttpHeaders copy = Http2Handler.copyHttp2Headers(source);
+
+        assertEquals("value", copy.get("invalid name"));
+        assertFalse(copy.contains(":status"));
+    }
+
+    @Test
+    void stillValidatesHeaderValues() {
```
hyperxpro: This one is good and I would keep it regardless of what happens to the rest of the PR. Value validation on this path was not covered before and it is the property that actually needs a guard, since the decoder does not check values at all.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
pavel-ptashyts: Added a focused test for the empty-name guard. Empty names are skipped before insertion, while the destination validator remains the backstop for malformed non-empty names.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
pavel-ptashyts: Agreed. Removed withNameValidation(false). Converted response headers now use Netty's default name and value validators.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
pavel-ptashyts: Replaced the capturing lambda with an enhanced for loop shared by response headers and trailers.

client/src/test/java/org/asynchttpclient/netty/handler/Http2HandlerTest.java
pavel-ptashyts: Kept and renamed the value-validation test. It continues to verify that CRLF in a decoded header value is rejected.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
pavel-ptashyts: Rewrote the comment to state the actual boundary: HTTP/2 validation does not enforce complete HTTP token syntax or decoded values, so both are validated during conversion.

client/src/test/java/org/asynchttpclient/netty/handler/Http2HandlerTest.java
pavel-ptashyts: Agreed and inverted the test. It now verifies that a CRLF-containing name is rejected during conversion; the previous invalid-name acceptance claim has been removed.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
pavel-ptashyts: Applied the trailer-specific factory in this revision. Content-Length, Transfer-Encoding, and Trailer are now rejected before trailing headers reach the handler.

pavel-ptashyts:
Addressed all review feedback in 0411239fe:

- restored complete response-header name and value validation
- applied trailer-specific restrictions
- replaced the capturing copy lambda with an enhanced for loop
- added coverage for malformed and empty names, malformed values, and prohibited trailers
- corrected the PR title and description to reflect the revised scope

The focused tests and the full JDK 11 ./mvnw clean verify pass.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
```diff
@@ -56,6 +58,8 @@
 public final class Http2Handler extends AsyncHttpClientHandler {
 
     private static final HttpVersion HTTP_2 = new HttpVersion("HTTP", 2, 0, true);
+    private static final HttpHeadersFactory RESPONSE_HEADERS_FACTORY = DefaultHttpHeadersFactory.headersFactory();
+    private static final HttpHeadersFactory TRAILING_HEADERS_FACTORY = DefaultHttpHeadersFactory.trailersFactory();
```
hyperxpro: This is the one thing blocking me. trailersFactory throws IllegalArgumentException on Content-Length, Transfer-Encoding and Trailer, and that exception has nowhere good to go. It propagates out of copyHttp2Trailers into the catch in handleRead and ends up in streamFailed, so a response whose body we already handed to the user fails at the very last moment.

I suggest dropping this constant and using the response factory for both paths, then skipping the prohibited names in the trailer copy instead of failing on them. That keeps the full name and value validation this PR is actually for, and matches what HttpObjectDecoder already does on the HTTP/1 side.

It's worth knowing that only two of the three names actually matter here. I checked and Netty's HTTP/2 decoder rejects transfer-encoding earlier as a connection specific header, so it never reaches this code. Still worth listing it for parity with the HTTP/1 helper, just do not expect it to fire.

With one factory left, the RESPONSE prefix on line 61 stops earning its keep. HEADERS_FACTORY reads better.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
```diff
@@ -240,6 +231,26 @@ private void handleHttp2TrailingHeadersFrame(Http2HeadersFrame headersFrame, Cha
         }
     }
 
+    static HttpHeaders copyHttp2Headers(Http2Headers h2Headers) {
+        return copyHttp2Headers(h2Headers, RESPONSE_HEADERS_FACTORY);
+    }
+
+    static HttpHeaders copyHttp2Trailers(Http2Headers h2Headers) {
```
hyperxpro, line 238: Following on from the factory comment. If you take a boolean here rather than a factory, the two entry points stay as thin as they are now and the shared loop gets one extra skip. Something like this, which I have run against decoder produced names.

private static HttpHeaders copyHttp2Headers(Http2Headers h2Headers, boolean trailers) with the loop body doing continue when the name is empty or starts with a colon, then continue when trailers is true and the name is one of the three prohibited ones, then add.

The prohibited check itself mirrors HttpObjectDecoder.isPermittedTrailingHeader, which is a private static in Netty so we cannot call it directly. Three contentEqualsIgnoreCase calls against HttpHeaderNames.CONTENT_LENGTH, TRANSFER_ENCODING and TRAILER does the same job.

Results from running exactly that. Trailers of content-length alone give an empty set, grpc-status plus content-length gives just grpc-status, x-checksum alone survives untouched, and nothing throws. Response path still returns content-length and x-ok while skipping the status pseudo-header. Bad names and bad values are still rejected on both paths.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
```diff
@@ -240,6 +231,26 @@ private void handleHttp2TrailingHeadersFrame(Http2HeadersFrame headersFrame, Cha
...
+        return copyHttp2Headers(h2Headers, TRAILING_HEADERS_FACTORY);
+    }
+
+    private static HttpHeaders copyHttp2Headers(Http2Headers h2Headers, HttpHeadersFactory headersFactory) {
+        // HTTP/2 validation does not enforce HTTP token syntax or header values, so validate both while converting.
```
hyperxpro, line 243: This comment is accurate now and I am glad it is here. One addition worth making once the trailer handling changes. It explains what we validate but not why trailers are treated differently, which is the surprising part for anyone reading this later. A line saying prohibited trailer fields are dropped rather than rejected, because the RFC constraint is on the sender and the body has already been delivered by this point, would save the next person the round trip I just did.

client/src/test/java/org/asynchttpclient/netty/handler/Http2HandlerTest.java
```diff
@@ -0,0 +1,79 @@
...
+    @Test
+    void rejectsInvalidHeaderValues() {
+        Http2Headers source = new DefaultHttp2Headers(false)
+                .add("x-test", "value\r\ninjected");
+
+        assertThrows(IllegalArgumentException.class, () -> Http2Handler.copyHttp2Headers(source));
+    }
+
+    @Test
+    void rejectsProhibitedTrailingHeaders() {
```
hyperxpro: This pins the behaviour I am asking you to change, so it will need inverting. Under the fix it becomes something like dropsProhibitedTrailingHeaders, asserting the prohibited names are absent rather than that the call throws.

One suggestion on the fixture. Rather than looping one name at a time, put a good field in alongside the bad ones, for example grpc-status equals 0 with content-length and trailer. Then assert grpc-status survives and the other two are gone. That pins the thing that actually regressed, which is not that we reject a bad trailer but that rejecting one takes the rest of the trailer section with it.

You can drop TRANSFER_ENCODING from the list while you are in here. The HTTP/2 decoder rejects it upstream so it never reaches this method, and the Arrays import goes with it.

client/src/test/java/org/asynchttpclient/netty/handler/Http2HandlerTest.java
```diff
@@ -0,0 +1,79 @@
...
+import static io.netty.handler.codec.http.HttpHeaderNames.TRANSFER_ENCODING;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class Http2HandlerTest {
+
+    @Test
+    void copiesRegularHeadersAndSkipsPseudoHeaders() {
```
hyperxpro, line 33: The gap I would most like closed. Nothing asserts that the response path still accepts content-length, and it must, since only trailers restrict it. It works correctly today, I checked. But the whole design now rests on the two paths using different rules, and if someone later tidied them into one shared factory then every HTTP/2 response carrying a content-length would start failing and no test here would notice. Cheap to add next to this one.

Two smaller ones if you are adding tests anyway. Nothing covers trailers accepting an ordinary field, so every trailer test is currently a rejection test, and nothing covers trailers skipping pseudo-headers.

client/src/test/java/org/asynchttpclient/netty/handler/Http2HandlerTest.java
```diff
@@ -0,0 +1,79 @@
...
+                .add("x-test", "value");
+
+        HttpHeaders copy = Http2Handler.copyHttp2Headers(source);
+
+        assertEquals("value", copy.get("x-test"));
+        assertFalse(copy.contains(":status"));
+    }
+
+    @Test
+    void rejectsInvalidHeaderNames() {
```
hyperxpro, line 47: This is the right regression guard for the problem in the first round and I am glad it is here. Minor hardening suggestion. It only covers CRLF, so a hypothetical future validator that checked for CRLF and nothing else would still pass it, which is close to the mistake we just got out of. Adding a name with a space or an opening parenthesis would pin the token rule rather than one character pair. Both are rejected today.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
pavel-ptashyts: Agreed. I removed the trailer-specific factory and now use the validating header factory for both paths. Prohibited trailer fields are dropped before insertion, so a late invalid field no longer fails the stream after the body has been delivered.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
pavel-ptashyts: Implemented this shape. The shared copy method now takes a trailer flag, skips empty and pseudo-header names, drops the three prohibited trailer fields with case-insensitive checks, and still validates every retained name and value.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
pavel-ptashyts: Added the rationale next to the trailer-specific branch: the restriction is on the sender, so prohibited fields are dropped instead of failing after the response body has already been delivered.

client/src/test/java/org/asynchttpclient/netty/handler/Http2HandlerTest.java
pavel-ptashyts: Inverted the test to cover the intended behavior. It now combines grpc-status with Content-Length and Trailer, verifies that grpc-status survives, the prohibited fields are removed, and no exception is thrown. Transfer-Encoding is no longer part of the fixture.

client/src/test/java/org/asynchttpclient/netty/handler/Http2HandlerTest.java
pavel-ptashyts: Added coverage for all three boundaries: Content-Length remains valid on the response path, an ordinary trailer survives, and a pseudo-header is skipped on the trailer path.

client/src/test/java/org/asynchttpclient/netty/handler/Http2HandlerTest.java
pavel-ptashyts: Strengthened the regression guard with names containing a space and an opening parenthesis, in addition to CRLF. Both response and trailer conversion paths now assert the full token validation.

hyperxpro (approved):

PR #2279, opened 2026-07-20
Redact sensitive headers from debug logs

### LINKED ISSUE #1739: authorization headers exposed in log
when log level is set to debug, the request object is logged.
this prints all headers, including authorization header.
in case a user has implemented their own credentials mechanism, for example, since Realm object was not sufficient for the authorization scheme (or any other reason) sensitive information can leak.
Bearer scheme authorization is one such example.

a simple fix is to not log authorization headers: when iterating the headers, if the header key is authorization, print something like * authorization data skipped *

this is a simple straighforward solution that is very robust, no matter why the user set headers themselves, no authorization header leak is possible (on the client side), assuming underlying implementation (such as netty) doesn't also log all headers.

### LINKED ISSUE #1740: authorization headers sensitive data leaks in debug logs
when log level is set to debug, the request object is logged.
this prints all headers, including authorization header.
in case a user has implemented their own credentials mechanism, for example, since Realm object was not sufficient for the authorization scheme (or any other reason) sensitive information can leak.
Bearer scheme authorization is one such example.

a simple fix is to not log authorization headers: when iterating the headers, if the header key is authorization, print something like * authorization data skipped *

this is a simple straighforward solution that is very robust, no matter why the user set headers themselves, no authorization header leak is possible (on the client side), assuming underlying implementation (such as netty) doesn't also log all headers.

pavel-ptashyts:
## Summary

- redact authentication and cookie values in HTTP and WebSocket debug logs
- share the redaction policy through a neutral utility used by request and Netty logging
- allow sensitive logging through an explicit startup-only JVM property or environment variable
- warn once at startup when sensitive logging is enabled
- avoid rendering complete request objects in retry and replay logs by default
- add regression coverage for Netty message formatting, request output, and every startup configuration

## Root cause

Debug logging passed complete Netty request and response objects to SLF4J. Their
string representations included every header value, so user-supplied
`Authorization` headers and other credentials could be written to application
logs.

## Sensitive logging

Sensitive values remain redacted by default. Full header values can be enabled
before the sensitive logging policy is initialized with either:

- JVM property: `-Dorg.asynchttpclient.enableSensitiveLogging=true`
- environment variable: `AHC_ENABLE_SENSITIVE_LOGGING=true`

The JVM property takes precedence over the environment variable. The setting is
read once and remains immutable for the process. A warning is logged once when
sensitive logging is enabled because credentials and session data may be exposed.

## Scope

This change addresses the header disclosure reported in #1739 and #1740.
Request form, query, multipart, and arbitrary body values do not have a reliable
generic sensitivity classification and remain outside this change.

## Validation

- JDK 11 focused `HttpMessageFormatterTest` and `RequestBuilderTest`: 33 tests passed, 0 skipped
- JDK 11 system-property, environment-variable, and precedence paths execute in isolated JVMs
- JDK 11 `./mvnw clean verify`: passed, including Javadocs and Revapi
- GitHub Actions matrix: passed on Ubuntu, macOS, and Windows with JDK 11, 17, 21, and 25

Closes #1739
Closes #1740

### DIFF
--- a/client/src/main/java/org/asynchttpclient/DefaultRequest.java
+++ b/client/src/main/java/org/asynchttpclient/DefaultRequest.java
@@ -37,6 +37,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
+import static org.asynchttpclient.netty.handler.HttpMessageFormatter.REDACTED;
+import static org.asynchttpclient.netty.handler.HttpMessageFormatter.isSensitiveHeader;
 import static org.asynchttpclient.util.MiscUtils.isNonEmpty;
 
 public class DefaultRequest implements Request {
@@ -293,7 +295,7 @@ public class DefaultRequest implements Request {
                 sb.append('\t');
                 sb.append(header.getKey());
                 sb.append(':');
-                sb.append(header.getValue());
+                sb.append(isSensitiveHeader(header.getKey()) ? REDACTED : header.getValue());
             }
         }
 
--- a/client/src/main/java/org/asynchttpclient/DefaultRequest.java
+++ b/client/src/main/java/org/asynchttpclient/DefaultRequest.java
@@ -37,9 +37,9 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
-import static org.asynchttpclient.netty.handler.HttpMessageFormatter.REDACTED;
-import static org.asynchttpclient.netty.handler.HttpMessageFormatter.isSensitiveHeader;
 import static org.asynchttpclient.util.MiscUtils.isNonEmpty;
+import static org.asynchttpclient.util.SensitiveLoggingUtils.REDACTED;
+import static org.asynchttpclient.util.SensitiveLoggingUtils.isSensitiveHeader;
 
 public class DefaultRequest implements Request {
 
@@ -294,7 +294,7 @@ public class DefaultRequest implements Request {
             for (Map.Entry<String, String> header : headers) {
                 sb.append('\t');
                 sb.append(header.getKey());
-                sb.append(':');
+                sb.append(": ");
                 sb.append(isSensitiveHeader(header.getKey()) ? REDACTED : header.getValue());
             }
         }
--- a/client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
@@ -1,0 +1,69 @@
+/*
+ *    Copyright (c) 2026 AsyncHttpClient Project. All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package org.asynchttpclient.netty.handler;
+
+import io.netty.handler.codec.http.HttpHeaderNames;
+import io.netty.handler.codec.http.HttpHeaders;
+import io.netty.handler.codec.http.HttpRequest;
+import io.netty.handler.codec.http.HttpResponse;
+
+import java.util.Map;
+
+/** Formats HTTP messages for logging without rendering sensitive header values. */
+public final class HttpMessageFormatter {
+
+    /** The value used in place of sensitive header values. */
+    public static final String REDACTED = "<redacted>";
+
+    private HttpMessageFormatter() {
+    }
+
+    static String format(HttpRequest request) {
+        StringBuilder value = new StringBuilder()
+                .append(request.method()).append(' ')
+                .append(request.uri()).append(' ')
+                .append(request.protocolVersion());
+        return appendHeaders(value, request.headers()).toString();
+    }
+
+    static String format(HttpResponse response) {
+        StringBuilder value = new StringBuilder()
+                .append(response.protocolVersion()).append(' ')
+                .append(response.status());
+        return appendHeaders(value, response.headers()).toString();
+    }
+
+    private static StringBuilder appendHeaders(StringBuilder value, HttpHeaders headers) {
+        for (Map.Entry<String, String> header : headers) {
+            value.append('\n').append(header.getKey()).append(": ")
+                    .append(isSensitiveHeader(header.getKey()) ? REDACTED : header.getValue());
+        }
+        return value;
+    }
+
+    /**
+     * Returns whether a header value must be redacted from logs.
+     *
+     * @param name the header name
+     * @return {@code true} for authentication and cookie headers
+     */
+    public static boolean isSensitiveHeader(CharSequence name) {
+        return HttpHeaderNames.AUTHORIZATION.contentEqualsIgnoreCase(name)
+                || HttpHeaderNames.PROXY_AUTHORIZATION.contentEqualsIgnoreCase(name)
+                || HttpHeaderNames.COOKIE.contentEqualsIgnoreCase(name)
+                || HttpHeaderNames.SET_COOKIE.contentEqualsIgnoreCase(name);
+    }
+}
--- a/client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
@@ -22,8 +22,17 @@ import io.netty.handler.codec.http.HttpResponse;
 
 import java.util.Map;
 
-/** Formats HTTP messages for logging without rendering sensitive header values. */
+/**
+ * Formats HTTP messages for logging. Sensitive header values are redacted unless the
+ * {@code org.asynchttpclient.enableSensitiveLogging} system property or {@code AHC_ENABLE_SENSITIVE_LOGGING} environment
+ * variable is set to {@code true} when this class is initialized. Enabling sensitive logging may expose credentials and
+ * session data.
+ */
 public final class HttpMessageFormatter {
+
+    private static final String ENABLE_SENSITIVE_LOGGING_PROPERTY = "org.asynchttpclient.enableSensitiveLogging";
+    private static final String ENABLE_SENSITIVE_LOGGING_ENVIRONMENT_VARIABLE = "AHC_ENABLE_SENSITIVE_LOGGING";
+    private static final boolean SENSITIVE_LOGGING_ENABLED = isSensitiveLoggingEnabled();
 
     /** The value used in place of sensitive header values. */
     public static final String REDACTED = "<redacted>";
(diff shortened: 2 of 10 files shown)


hyperxpro:
I would like to have a system property / environment variable to enable non-redacted logs. It is helpful in debugging at CI pipeline.

Probably like AHC_ENABLE_SENSITIVE_LOGGING=true?

Of course feel free to choose a better key.

pavel-ptashyts:
@hyperxpro Implemented in [`2a7f11b84`](https://github.com/AsyncHttpClient/async-http-client/commit/2a7f11b847b7828beb4b175600b454edb2cbdd53).

Sensitive values remain redacted by default. Unredacted logging can be enabled before `HttpMessageFormatter` is initialized with either:

- JVM property: `-Dorg.asynchttpclient.enableSensitiveLogging=true`
- environment variable: `AHC_ENABLE_SENSITIVE_LOGGING=true`

The value is calculated once and remains immutable for the process. The JVM property takes precedence over the environment variable, including allowing an explicit `false` to keep redaction enabled. Full JDK 11 verification passes.

Thanks for the suggestion.

client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
```diff
@@ -0,0 +1,84 @@
...
+ * Formats HTTP messages for logging. Sensitive header values are redacted unless the
+ * {@code org.asynchttpclient.enableSensitiveLogging} system property or {@code AHC_ENABLE_SENSITIVE_LOGGING} environment
+ * variable is set to {@code true} when this class is initialized. Enabling sensitive logging may expose credentials and
+ * session data.
+ */
+public final class HttpMessageFormatter {
+
+    private static final String ENABLE_SENSITIVE_LOGGING_PROPERTY = "org.asynchttpclient.enableSensitiveLogging";
+    private static final String ENABLE_SENSITIVE_LOGGING_ENVIRONMENT_VARIABLE = "AHC_ENABLE_SENSITIVE_LOGGING";
+    private static final boolean SENSITIVE_LOGGING_ENABLED = isSensitiveLoggingEnabled();
```
hyperxpro: This is read once when the class loads and never rechecked. Surefire in this repo reuses a single JVM across the whole client module by default, so whichever test touches this class first locks the value for the rest of the run. That means the EnabledIfSystemProperty and EnabledIfEnvironmentVariable tests further down just get skipped in a normal build instead of actually running, so the enabled path has no real coverage right now. Might be worth resolving this per call or behind a seam that a test can override.

client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
```diff
@@ -0,0 +1,84 @@
...
+        StringBuilder value = new StringBuilder()
+                .append(response.protocolVersion()).append(' ')
+                .append(response.status());
+        return appendHeaders(value, response.headers()).toString();
+    }
+
+    private static StringBuilder appendHeaders(StringBuilder value, HttpHeaders headers) {
+        for (Map.Entry<String, String> header : headers) {
+            value.append('\n').append(header.getKey()).append(": ")
+                    .append(isSensitiveHeader(header.getKey()) ? REDACTED : header.getValue());
```
hyperxpro, line 57: This uses a colon followed by a space between header name and value, but DefaultRequest toString uses a colon with no space for the same kind of dump. Small thing, but the two renderings will drift if they stay separate.

client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
```diff
@@ -0,0 +1,84 @@
...
+        return Boolean.parseBoolean(value);
+    }
+
+    /**
+     * Returns whether a header value must be redacted from logs.
+     *
+     * @param name the header name
+     * @return {@code true} for authentication and cookie headers when sensitive logging is disabled
+     */
+    public static boolean isSensitiveHeader(CharSequence name) {
```
hyperxpro, line 68: DefaultRequest in the base package now depends on this method and REDACTED. Since the base Request model has not reached into netty handler before, it might be cleaner to move this predicate and the redacted constant into a neutral utility package and have this class consume it instead.

client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private static StringBuilder appendHeaders(StringBuilder value, HttpHeaders headers) {
+        for (Map.Entry<String, String> header : headers) {
+            value.append('\n').append(header.getKey()).append(": ")
+                    .append(isSensitiveHeader(header.getKey()) ? REDACTED : header.getValue());
+        }
+        return value;
+    }
+
+    private static boolean isSensitiveLoggingEnabled() {
```
hyperxpro: This reads from an environment variable that can be inherited from a container or orchestration setup. Worth considering a one time warn log when this ends up enabled, so it shows up in the logs instead of only being discoverable after something has already leaked.

client/src/main/java/org/asynchttpclient/DefaultRequest.java
```diff
@@ -37,6 +37,8 @@
 import java.util.List;
 import java.util.Map;
 
+import static org.asynchttpclient.netty.handler.HttpMessageFormatter.REDACTED;
```
hyperxpro: This pulls the base Request model into a dependency on the netty handler package, which has been treated as transport level internal up to now. See the note on HttpMessageFormatter about moving the shared predicate somewhere neutral instead.

client/src/main/java/org/asynchttpclient/DefaultRequest.java
```diff
@@ -293,7 +295,7 @@ public String toString() {
                 sb.append('\t');
                 sb.append(header.getKey());
                 sb.append(':');
-                sb.append(header.getValue());
+                sb.append(isSensitiveHeader(header.getKey()) ? REDACTED : header.getValue());
```
hyperxpro, line 298: Headers are redacted here but formParams a bit further down in this same method still get appended in full. If a password ever comes through as a form field it would still show up in this output untouched, worth checking if that is in scope for the issues this closes.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -1413,7 +1414,7 @@ public void replayRequest(final NettyResponseFuture<?> future, FilterContext fc,
         future.setChannelState(ChannelState.NEW);
         future.touch();
 
-        LOGGER.debug("\n\nReplaying Request {}\n for Future {}\n", newRequest, future);
+        LOGGER.debug("\n\nReplaying request '{}' to '{}'\n for Future {}\n", newRequest.getMethod(), newRequest.getUri(), future);
```
hyperxpro, line 1417: Checked this one, NettyResponseFuture toString does not render the underlying NettyRequest headers since NettyRequest has no toString override, so this falls back to identity hash and does not leak anything today. 

Might be worth a short comment here noting that this is load bearing, so a future toString added to NettyRequest does not quietly reopen this.

client/src/test/java/org/asynchttpclient/netty/handler/HttpMessageFormatterTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+        String value = HttpMessageFormatter.format(response);
+
+        assertFalse(value.contains("response-secret"));
+        assertTrue(value.contains("Set-Cookie: <redacted>"));
+        assertTrue(value.contains("X-Request-Id: request-id"));
+    }
+
+    @Test
+    @EnabledIfSystemProperty(named = "org.asynchttpclient.enableSensitiveLogging", matches = "(?i)true")
```
hyperxpro: Same concern as on HttpMessageFormatter, this test and the one below it get silently skipped in a normal run rather than exercised, so the enabled path is effectively untested in CI as configured.

hyperxpro (changes requested):

hyperxpro:
Also please don't rebase your PRs as it spams my inbox :)

client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
pavel-ptashyts: Aligned DefaultRequest with the formatter colon-space spelling. Both paths now use the same neutral redaction policy.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Added the load-bearing comment directly in NettyResponseFuture.toString, where NettyRequest is rendered, explaining that NettyRequest must keep identity-only rendering so request headers are not exposed.

client/src/test/java/org/asynchttpclient/netty/handler/HttpMessageFormatterTest.java
pavel-ptashyts: Replaced the skipped conditional tests with isolated JVM probes. Both enabled modes and property-over-environment precedence now execute in the regular CI test suite.

client/src/main/java/org/asynchttpclient/DefaultRequest.java
pavel-ptashyts: Checked the scope against #1739 and #1740: both report Authorization and header disclosure. Form, query, multipart, and arbitrary body fields have no reliable generic sensitivity classification, so this PR intentionally leaves request-body rendering unchanged and now documents that scope.

client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
pavel-ptashyts: Added a one-time WARN during policy initialization whenever sensitive logging is enabled, regardless of whether it came from the system property or inherited environment.

client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
pavel-ptashyts: Kept the startup value immutable, but replaced the conditional tests with isolated JVM probes. The system-property, environment-variable, and precedence paths now run in every normal test build.

client/src/main/java/org/asynchttpclient/netty/handler/HttpMessageFormatter.java
pavel-ptashyts: Moved the shared REDACTED value and predicate to SensitiveLoggingUtils in the neutral util package. HttpMessageFormatter retains its public facade and delegates to that policy.

client/src/main/java/org/asynchttpclient/DefaultRequest.java
pavel-ptashyts: Removed the dependency from DefaultRequest to the Netty handler package. It now consumes SensitiveLoggingUtils from the neutral util package.

pavel-ptashyts:
Addressed the review feedback in 3b572a6c7:

- moved the shared redaction policy to the neutral util package
- added a one-time warning when sensitive logging is enabled
- replaced skipped startup-mode tests with isolated JVM coverage
- aligned request header formatting and documented the NettyRequest rendering invariant
- clarified that this PR covers header disclosure, not arbitrary request-body fields

Full JDK 11 clean verify passes, including Javadocs and Revapi.

No rebase was used for this update, and I will keep future updates as regular commits.

hyperxpro:
There are test failures, can you have a look?

pavel-ptashyts:
Fixed the Windows CI timeouts in 12474ee7f.

The isolated probe JVMs wrote Logback and Netty startup output to a pipe while the parent process waited for them to exit. The smaller Windows pipe buffer filled first, which blocked each child until the 30-second timeout. Probe output is now redirected to a JUnit temporary file and read after the child exits.

Validation:
- focused HttpMessageFormatterTest: 5 tests passed, 0 skipped
- JDK 11 ./mvnw clean verify: passed, including Javadocs and Revapi

No rebase was used.

pavel-ptashyts:
The follow-up CI run is fully green: compile-and-check plus Ubuntu, macOS, and Windows on JDK 11, 17, 21, and 25 all pass.

hyperxpro (approved):

PR #2289, opened 2026-07-25
Fix idle pool cleaner generation race

### LINKED ISSUE #2284: DefaultChannelPool cleaner can close a freshly re-offered channel
## Summary

In 3.0.11, `DefaultChannelPool` can close a channel that was leased and then freshly returned to the pool while the idle-channel cleaner was inspecting it.

The cleaner calculates that the channel is idle-expired using the old idle timestamp. If the channel is leased and re-offered before the cleaner claims it, `offer()` resets the shared `IdleState` with a fresh timestamp.

However, the cleaner captures `startSnapshot` only after calculating the expiry result. It can therefore capture the new timestamp, successfully claim the new idle generation, and close it using the expiry result calculated from the old generation.

## Version

Reproduced against the official `async-http-client-project-3.0.11` tag.

The same deterministic reproduction also fails on current `main` as of `f01188629` (`3.0.12-SNAPSHOT`).

The relevant shared `IdleState` behavior was introduced by #2183. The same reproduction does not occur with the per-offer idle holder used in 3.0.10.

## Problematic ordering

The current cleaner flow is effectively:

```java
boolean isIdleTimeoutExpired = isIdleTimeoutExpired(idleState, now);

// Other checks can allow the channel to be leased and re-offered here.

long startSnapshot = idleState.start();

if (!idleState.takeOwnership()) {
    continue;
}

if (idleState.start() != startSnapshot) {
    idleState.releaseOwnership();
    continue;
}

close(channel);
```

The expiry decision and the ownership validation can refer to different idle generations.

## Deterministic reproduction

I created a unit test that forces this interleaving:

1. Offer a channel and allow its idle timeout to expire.
2. Start the cleaner.
3. Pause the cleaner after it has calculated that the old idle generation is expired.
4. Poll the channel and offer it back, resetting its `IdleState`.
5. Resume the cleaner.
6. Verify that the freshly re-offered channel remains active and leasable.

On unmodified 3.0.11, the assertion fails. The cleaner closes the channel and logs:

```text
isIdleTimeoutExpired=true isRemotelyClosed=false isTtlExpired=false
```

## Why reordering the timestamp read is not sufficient

Moving `startSnapshot` before the expiry calculation prevents the stale close in the reproduction above, but it still leaves a smaller race:

1. The cleaner evaluates old generation A.
2. The channel is leased and re-offered as fresh generation B.
3. The cleaner claims generation B.
4. Before the cleaner notices the generation changed and releases the claim, another `poll()` removes B from the deque but cannot take ownership.
5. The cleaner releases ownership, leaving the active channel absent from the pool.

I reproduced this second interleaving with a deterministic test that pauses the cleaner immediately after the claim. This is a fix-safety concern: the observed 3.0.11 failure is the stale close described above, while this interleaving shows why a timestamp-only change would not be a complete fix.

## Fix requirement

The cleaner needs to atomically claim the exact idle generation that it evaluated. If the channel has been leased and re-offered since the snapshot was taken, the claim must fail without temporarily owning the fresh generation.

Conceptually (pseudocode, not a proposed public API), the cleaner flow should be:

```java
IdleSnapshot snapshot = idleState.snapshot();

if (!isExpired(snapshot, channel, now)) {
    continue;
}

if (!idleState.tryTakeOwnership(snapshot)) {
    continue;
}

close(channel);
```

`tryTakeOwnership(snapshot)` must use a compare-and-set operation that succeeds only when the state is still the same unowned generation represented by `snapshot`. One possible implementation is a stamped atomic state containing both the generation identity and ownership flag.

Another option is to restore a distinct per-offer idle holder, where the timestamp and ownership flag belong to one deque entry. That is simpler to reason about and avoids sharing ownership across idle generations, but it restores the per-offer allocation that #2183 was intended to remove.

The exact representation is open for discussion; the correctness requirement is that generation validation and ownership acquisition form one atomic transition.

## Regression coverage

A fix should include deterministic tests for both cases:

1. A stale idle-expiry decision must not close a channel that was leased and re-offered.
2. A poll racing with the cleaner's generation validation must not leave a live, freshly re-offered channel absent from the pool.

I’m happy to contribute the implementation and regression tests once the preferred state representation is clear.

pavel-ptashyts:
## Summary

- claim the exact idle generation evaluated by the channel pool cleaner
- keep generation and ownership in one atomic state without restoring per-offer allocations
- cover stale close, fresh entry loss, and equal-timestamp generations with regression tests

Fixes #2284

## Testing

- `./mvnw -pl client -Dtest=DefaultChannelPoolTest test` (19 tests passed)
- full JDK 11 `./mvnw clean verify` attempt: 1,326 tests passed and one unrelated timing assertion failed in `SemaphoreTest.combinedCheckAcquireTime`
- immediate isolated `SemaphoreTest` rerun passed (10/10)

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
@@ -52,10 +52,11 @@ public final class DefaultChannelPool implements ChannelPool {
     private static final AttributeKey<IdleState> IDLE_STATE_ATTRIBUTE_KEY = AttributeKey.valueOf("channelIdleState");
 
     // The partition deques hold the bare Channel; per-checkout idle state (start timestamp + the
-    // generation/ownership CAS state) lives on the channel's IDLE_STATE_ATTRIBUTE_KEY attribute,
+    // generation/ownership/reset CAS state) lives on the channel's IDLE_STATE_ATTRIBUTE_KEY attribute,
     // which is allocated once per physical connection and reused across every pool cycle.
     private final ConcurrentHashMap<Object, ConcurrentLinkedDeque<Channel>> partitions = new ConcurrentHashMap<>();
     private final AtomicBoolean isClosed = new AtomicBoolean(false);
+    private final AtomicBoolean duplicateOfferLogged = new AtomicBoolean(false);
     private final Timer nettyTimer;
     private final long connectionTtl;
     private final boolean connectionTtlEnabled;
@@ -119,19 +120,10 @@ public final class DefaultChannelPool implements ChannelPool {
             return false;
         }
 
-        boolean offered = offer0(channel, partitionKey, now);
-        if (connectionTtlEnabled && offered) {
-            registerChannelCreation(channel, partitionKey, now);
-        }
-
-        return offered;
+        return offer0(channel, partitionKey, now);
     }
 
     private boolean offer0(Channel channel, Object partitionKey, long now) {
-        ConcurrentLinkedDeque<Channel> partition = partitions.get(partitionKey);
-        if (partition == null) {
-            partition = partitions.computeIfAbsent(partitionKey, pk -> new ConcurrentLinkedDeque<>());
-        }
         // Reuse the channel's IdleState instead of allocating a holder per offer; reset() transfers an
         // owned generation to a new leasable generation before offerFirst publishes the channel.
         Attribute<IdleState> idleStateAttribute = channel.attr(IDLE_STATE_ATTRIBUTE_KEY);
@@ -142,9 +134,18 @@ public final class DefaultChannelPool implements ChannelPool {
             idleState = existingIdleState == null ? newIdleState : existingIdleState;
         }
         if (!idleState.reset(now)) {
-            return false;
+            if (LOGGER.isDebugEnabled() && duplicateOfferLogged.compareAndSet(false, true)) {
+                LOGGER.debug("Ignoring duplicate pool offer for channel {}", channel);
+            }
+            return true;
         }
-        return partition.offerFirst(channel);
+        ConcurrentLinkedDeque<Channel> partition =
+                partitions.computeIfAbsent(partitionKey, pk -> new ConcurrentLinkedDeque<>());
+        boolean offered = partition.offerFirst(channel);
+        if (connectionTtlEnabled && offered) {
+            registerChannelCreation(channel, partitionKey, now);
+        }
+        return offered;
     }
 
     private static void registerChannelCreation(Channel channel, Object partitionKey, long now) {
@@ -295,19 +296,21 @@ public final class DefaultChannelPool implements ChannelPool {
      * {@link #IDLE_STATE_ATTRIBUTE_KEY} attribute, then reused across every pool checkout so no holder
      * is allocated per offer.
      *
-     * <p>The low bit of {@code state} is the ownership flag; the remaining bits advance by two on every
-     * offer because bit zero is reserved. A successful {@code poll()} lease and a {@code removeAll()}
-     * tombstone both claim the current generation. The cleaner can therefore claim only the exact
-     * generation whose expiry it evaluated, without temporarily claiming a newer entry while checking
-     * whether {@code start} changed.
+     * <p>The low bit of {@code state} is the ownership flag and the next bit marks a reset in progress;
+     * the remaining bits advance by four on every offer. A successful {@code poll()} lease, a
+     * {@code removeAll()} tombstone, and the cleaner's pre-close claim all own the current generation.
+     * The cleaner can therefore claim only the exact generation whose expiry it evaluated, without
+     * temporarily claiming a newer entry while checking whether {@code start} changed.
      *
-     * <p>A generation can be reset only while owned. New state starts owned, and {@code poll()} claims
-     * a generation before returning it, so {@code offer()} transfers ownership instead of performing a
-     * non-atomic update from a leasable generation.
+     * <p>A generation can be reset only while owned and not already being reset. New state starts owned,
+     * and reset first claims the generation's reset bit before publishing the timestamp and next leasable
+     * generation. Concurrent or delayed resets therefore cannot transfer a generation they did not claim.
      */
     static final class IdleState {
 
         private static final long OWNED_MASK = 1L;
+        private static final long RESETTING_MASK = 2L;
+        private static final long GENERATION_INCREMENT = 4L;
         private static final AtomicLongFieldUpdater<IdleState> STATE_UPDATER =
                 AtomicLongFieldUpdater.newUpdater(IdleState.class, "state");
 
@@ -330,6 +333,10 @@ public final class DefaultChannelPool implements ChannelPool {
             return (stateSnapshot & OWNED_MASK) != 0;
         }
 
+        private static boolean isResetting(long stateSnapshot) {
+            return (stateSnapshot & RESETTING_MASK) != 0;
+        }
+
         /** Atomically claim the current generation. */
         boolean takeOwnership() {
             return tryTakeOwnership(state);
@@ -341,15 +348,20 @@ public final class DefaultChannelPool implements ChannelPool {
                     && STATE_UPDATER.compareAndSet(this, stateSnapshot, stateSnapshot | OWNED_MASK);
         }
 
-        /** Transfer an owned generation to a new leasable generation. Called on every offer. */
-        synchronized boolean reset(long now) {
-            long stateSnapshot = state;
-            if (!isOwned(stateSnapshot)) {
+        /** Attempt to transfer the current owned generation to a new leasable generation. */
+        boolean reset(long now) {
+            return tryReset(state, now);
+        }
+
+        /** Transfer {@code stateSnapshot} if it is still owned, current, and not already being reset. */
+        boolean tryReset(long stateSnapshot, long now) {
+            if (!isOwned(stateSnapshot) || isResetting(stateSnapshot)
+                    || !STATE_UPDATER.compareAndSet(this, stateSnapshot, stateSnapshot | RESETTING_MASK)) {
                 return false;
             }
-            long nextGeneration = (stateSnapshot & ~OWNED_MASK) + 2;
             start = now;
-            return STATE_UPDATER.compareAndSet(this, stateSnapshot, nextGeneration);
+            state = (stateSnapshot & ~(OWNED_MASK | RESETTING_MASK)) + GENERATION_INCREMENT;
+            return true;
         }
     }
 
@@ -430,10 +442,10 @@ public final class DefaultChannelPool implements ChannelPool {
 
                 long stateSnapshot = idleState.snapshot();
                 if (IdleState.isOwned(stateSnapshot)) {
-                    // In-deque + owned ==> a removeAll(Channel) tombstone, or a node a concurrent poll()
-                    // has already leased and unlinked. Either way: unlink, never close — the owner of the
-                    // claim is responsible for closing it. Unlinking an already-unlinked node through the
-                    // iterator is a harmless no-op.
+                    // In-deque + owned ==> a removeAll(Channel) tombstone, a node a concurrent poll()
+                    // already leased and unlinked, or an old node whose generation is being reset.
+                    // Either way: unlink, never close — the owner of the claim handles the channel.
+                    // Unlinking an already-unlinked node through the iterator is a harmless no-op.
                     it.remove();
                     continue;
                 }
--- a/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
+++ b/client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
@@ -37,7 +37,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import java.util.function.Predicate;
 
 import static org.asynchttpclient.util.DateUtils.unpreciseMillisTime;
@@ -52,8 +52,8 @@ public final class DefaultChannelPool implements ChannelPool {
     private static final AttributeKey<IdleState> IDLE_STATE_ATTRIBUTE_KEY = AttributeKey.valueOf("channelIdleState");
 
     // The partition deques hold the bare Channel; per-checkout idle state (start timestamp + the
-    // owned/tombstone CAS flag) lives on the channel's IDLE_STATE_ATTRIBUTE_KEY attribute, which is
-    // allocated once per physical connection and reused across every pool cycle (no per-offer holder).
+    // generation/ownership CAS state) lives on the channel's IDLE_STATE_ATTRIBUTE_KEY attribute,
+    // which is allocated once per physical connection and reused across every pool cycle.
     private final ConcurrentHashMap<Object, ConcurrentLinkedDeque<Channel>> partitions = new ConcurrentHashMap<>();
     private final AtomicBoolean isClosed = new AtomicBoolean(false);
     private final Timer nettyTimer;
@@ -132,9 +132,8 @@ public final class DefaultChannelPool implements ChannelPool {
         if (partition == null) {
             partition = partitions.computeIfAbsent(partitionKey, pk -> new ConcurrentLinkedDeque<>());
         }
-        // Reuse the channel's IdleState instead of allocating a holder per offer; reset() stamps the
-        // idle start and clears the owned flag (must happen-before offerFirst publishes the channel,
-        // so any thread that observes it in the deque also observes owned == 0).
+        // Reuse the channel's IdleState instead of allocating a holder per offer; reset() publishes a
+        // new idle generation before offerFirst publishes the channel.
         Attribute<IdleState> idleStateAttribute = channel.attr(IDLE_STATE_ATTRIBUTE_KEY);
         IdleState idleState = idleStateAttribute.get();
         if (idleState == null) {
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -293,52 +292,62 @@ private static final class ChannelCreation {
...
+        /** Atomically claim {@code stateSnapshot}, provided it is still the current generation. */
+        boolean takeOwnership(long stateSnapshot) {
+            return !isOwned(stateSnapshot)
+                    && STATE_UPDATER.compareAndSet(this, stateSnapshot, stateSnapshot | OWNED_MASK);
         }
 
-        /** Stamp the idle start and mark the channel leasable again. Called on every offer. */
+        /** Stamp a new idle generation and mark the channel leasable again. Called on every offer. */
         void reset(long now) {
+            long nextGeneration = (state & ~OWNED_MASK) + 2;
```
hyperxpro: `reset()` needs its precondition written down (DefaultChannelPool.java:337). The non-atomic RMW of state is safe only while reset() never runs on an unowned state. If it does: reset() reads G unowned → cleaner CAS(G, G|1) succeeds → reset()'s plain stores clobber the claim → cleaner closes the channel while offerFirst publishes it as leasable. Unreachable in production today (every path goes through 

`ChannelManager.tryToOfferChannelToPool`, which setDiscards first, and poll() only returns after a successful claim), but ChannelPool is a public SPI that documents no such rule, and the PR removed the paragraph that documented the old design's equivalent invariant. Either restore an explicit invariant paragraph, or make reset an ownership-transfer CAS (compareAndSet(s, (s & ~OWNED_MASK) + 2) on an isOwned(s) snapshot) so
IdleState is safe under any caller. Worth noting concurrentOfferPollRemoveAllIsConsistent already offers without lease discipline — it only survives because it runs with the cleaner effectively disabled.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -293,52 +292,62 @@ private static final class ChannelCreation {
...
      * is allocated per offer.
      *
-     * <p>{@code owned} is a single CAS flag with two roles, both meaning "this idle entry is claimed,
-     * do not lease it": a successful {@code poll()} lease, or a {@code removeAll()} tombstone. The pool
-     * upholds the invariant that a channel sitting in a partition deque has {@code owned == 0} unless it
-     * was tombstoned, because {@link #reset(long)} clears the flag before {@code offerFirst} publishes
-     * the channel and {@code poll()} unlinks a channel from the deque before claiming it. {@code start}
-     * doubles as a generation token: it changes on every offer, letting the cleaner detect a channel
-     * that was leased and re-offered between its expiry check and its claim.
+     * <p>The low bit of {@code state} is the ownership flag; the remaining bits are incremented on
```
hyperxpro: The remaining bits are incremented by two on every offer, not by one, since bit zero is the ownership flag. Worth saying explicitly so the plus two in reset is not read as a typo.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -293,52 +292,62 @@ private static final class ChannelCreation {
...
+            return isOwned(state);
         }
 
-        /** Atomically claim this entry; returns true only for the caller that transitions 0 -> 1. */
+        static boolean isOwned(long stateSnapshot) {
+            return (stateSnapshot & OWNED_MASK) != 0;
+        }
+
+        /** Atomically claim the current generation. */
         boolean takeOwnership() {
```
hyperxpro, line 341: This overload pair is easy to misread at a call site, because the no argument form silently re reads the field while the one argument form validates a caller supplied snapshot, and both are named takeOwnership. The tryTakeOwnership name suggested in the issue for the snapshot form reads better and makes the failure mode obvious.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -293,52 +292,62 @@ private static final class ChannelCreation {
...
+        boolean takeOwnership(long stateSnapshot) {
+            return !isOwned(stateSnapshot)
+                    && STATE_UPDATER.compareAndSet(this, stateSnapshot, stateSnapshot | OWNED_MASK);
         }
 
-        /** Stamp the idle start and mark the channel leasable again. Called on every offer. */
+        /** Stamp a new idle generation and mark the channel leasable again. Called on every offer. */
         void reset(long now) {
+            long nextGeneration = (state & ~OWNED_MASK) + 2;
+            // Keep the new generation unavailable until its timestamp is published.
```
hyperxpro: This comment does not describe what the store actually does. The JMM already keeps the new generation from being observed with a stale timestamp without this store. Volatile accesses are totally ordered in the synchronization order, per JLS 17.4.4 and 17.4.7, so a reader that observes state equal to nextGeneration must also observe start as now or newer, because the store to start precedes the store to state in program order.

What this extra store really does is narrow, but not close, the window described in the comment above on the generation read, and it costs an extra StoreLoad fence per pooled request on x86 64. Either drop it, or keep it and relabel it honestly as defence in depth.

client/src/test/java/org/asynchttpclient/netty/channel/DefaultChannelPoolTest.java
```diff
@@ -245,6 +246,63 @@ public void channelReofferedAfterExpiryIsNotReaped() throws Exception {
...
+        cleanerRun.get(5, TimeUnit.SECONDS);
+
+        assertTrue(channel.isActive(), "cleaner must not close a freshly re-offered channel");
+        assertSame(channel, pool.poll(KEY), "freshly re-offered channel must remain leasable");
+
+        pool.destroy();
+    }
+
+    @Test
+    public void staleCleanerClaimDoesNotHideReofferedChannelFromPoll() throws Exception {
```
hyperxpro: This test never runs the cleaner. It calls takeOwnership with a stale snapshot directly and then asserts that a failed CAS had no side effect, which restates the contract of compareAndSet rather than exercising the pool. Race b from the issue therefore has no end to end coverage.

I would fold it into the BlockingActiveChannel test and assert that the partition still holds one entry after the cleaner pass, since that is the assertion that would actually catch the cleaner hiding the entry.

Related gap: the soak test concurrentOfferPollRemoveAllIsConsistent runs with idle timeout off and a one hour TTL, so the cleaner only ever takes the isOwned then remove branch and never claims or closes anything.

client/src/test/java/org/asynchttpclient/netty/channel/DefaultChannelPoolTest.java
```diff
@@ -490,14 +548,51 @@ public void concurrentOfferPollRemoveAllIsConsistent() throws Exception {
...
 
+    private static final class BlockingActiveChannel extends EmbeddedChannel {
+
+        private final AtomicBoolean blockNextActiveCheck = new AtomicBoolean();
+        private final CountDownLatch activeCheck = new CountDownLatch(1);
+        private final CountDownLatch resumeActiveCheck = new CountDownLatch(1);
+
+        @Override
+        public boolean isActive() {
+            AtomicBoolean block = blockNextActiveCheck;
```
hyperxpro, line 598: This null guard is load bearing and deserves a one line comment saying why. EmbeddedChannel's constructor calls the subclass isActive once while these fields are still null, so without the guard you get an NPE. The trap is that the NPE is swallowed by a DefaultPromise listener, so removing this check would fail silently rather than loudly, and an IDE will flag the check as always true on a final field and invite someone to delete it.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
pavel-ptashyts: Agreed. I made reset an ownership transfer instead of a non-atomic update. IdleState now starts owned, reset rejects a leasable generation, publishes the timestamp, and CASes the owned snapshot to the next generation. offer also installs the per-channel state with setIfAbsent and returns false when the transfer fails. I added a regression test for duplicate offers.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
pavel-ptashyts: Updated the description to say that the generation bits advance by two on every offer because bit zero is reserved for ownership.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
pavel-ptashyts: Renamed the snapshot form to tryTakeOwnership. The no-argument takeOwnership method now clearly delegates to the conditional snapshot operation.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
pavel-ptashyts: Removed the extra volatile state store. reset now writes the timestamp and then CASes the owned snapshot directly to the next leasable generation; synchronized reset calls prevent concurrent publishers from corrupting the winning timestamp.

client/src/test/java/org/asynchttpclient/netty/channel/DefaultChannelPoolTest.java
pavel-ptashyts: Removed the standalone CAS-contract test and folded the observable guarantee into the BlockingActiveChannel test. It now asserts that the partition still contains exactly one fresh entry after the cleaner pass, before polling it. I also enabled idle expiry in the soak test and added a dedicated expiring channel with a close latch, so the test proves that the cleaner ownership and close path actually runs.

client/src/test/java/org/asynchttpclient/netty/channel/DefaultChannelPoolTest.java
pavel-ptashyts: Added the one-line constructor-dispatch explanation above the null guard so the EmbeddedChannel initialization trap is explicit.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -132,16 +132,18 @@ private boolean offer0(Channel channel, Object partitionKey, long now) {
         if (partition == null) {
             partition = partitions.computeIfAbsent(partitionKey, pk -> new ConcurrentLinkedDeque<>());
```
hyperxpro: Minor one. This runs before the reset guard now, so a rejected offer leaves an empty deque sitting in partitions and the cleaner walks it every tick. Was not possible before, since offer0 could not return false. Probably just move the lookup below the reset call.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -132,16 +132,18 @@ private boolean offer0(Channel channel, Object partitionKey, long now) {
...
         Attribute<IdleState> idleStateAttribute = channel.attr(IDLE_STATE_ATTRIBUTE_KEY);
         IdleState idleState = idleStateAttribute.get();
         if (idleState == null) {
-            idleState = new IdleState();
-            idleStateAttribute.set(idleState);
+            IdleState newIdleState = new IdleState();
+            IdleState existingIdleState = idleStateAttribute.setIfAbsent(newIdleState);
+            idleState = existingIdleState == null ? newIdleState : existingIdleState;
+        }
+        if (!idleState.reset(now)) {
```
hyperxpro, line 136: Heads up that false here means the caller closes the channel. ChannelManager line 426 calls closeChannel on a false return, and that is really what false has always meant here, NoopChannelPool returns false every time exactly so nothing gets pooled.

So a duplicate offer now kills a live pooled channel. I tried it: second offer returns false, the close leaves it inactive but still in the deque, and the next poll on that partition comes back null. Nothing in tree can actually hit this since setDiscard runs first, but it does turn a harmless no op into quietly dropping a good keep alive.

Might be simpler to treat a duplicate as a no op, debug log and return true.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -293,52 +295,68 @@ private static final class ChannelCreation {
...
-     * the channel and {@code poll()} unlinks a channel from the deque before claiming it. {@code start}
-     * doubles as a generation token: it changes on every offer, letting the cleaner detect a channel
-     * that was leased and re-offered between its expiry check and its claim.
+     * <p>The low bit of {@code state} is the ownership flag; the remaining bits advance by two on every
+     * offer because bit zero is reserved. A successful {@code poll()} lease and a {@code removeAll()}
+     * tombstone both claim the current generation. The cleaner can therefore claim only the exact
+     * generation whose expiry it evaluated, without temporarily claiming a newer entry while checking
+     * whether {@code start} changed.
+     *
+     * <p>A generation can be reset only while owned. New state starts owned, and {@code poll()} claims
```
hyperxpro: Tiny thing, this says two producers of owned state but there are four. The removeAll tombstone and the cleaner's pre close claim leave it owned too, and offer will reset from either. That part is pre existing and harmless since poll drops inactive channels, it just reads like a complete list.

Separately, the guard returns false with nothing logged, so if the invariant ever does break, all you would see is a connection closing for no apparent reason.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -293,52 +295,68 @@ private static final class ChannelCreation {
...
+        /** Atomically claim {@code stateSnapshot}, provided it is still the current generation. */
+        boolean tryTakeOwnership(long stateSnapshot) {
+            return !isOwned(stateSnapshot)
+                    && STATE_UPDATER.compareAndSet(this, stateSnapshot, stateSnapshot | OWNED_MASK);
         }
 
-        /** Stamp the idle start and mark the channel leasable again. Called on every offer. */
-        void reset(long now) {
+        /** Transfer an owned generation to a new leasable generation. Called on every offer. */
+        synchronized boolean reset(long now) {
```
hyperxpro: I do not think the synchronized is buying what it looks like it is buying. If a duplicate offer loses the race it parks here, and when it wakes up it re reads state fresh. So if a poll grabbed ownership in the meantime, the loser sails straight through the isOwned check and stamps a timestamp that is now well out of date, onto a generation that is not its own.

Walking through it. reset(1000) publishes gen 2 with start 1000, poll takes gen 2, a second thread calls reset(1000) and parks. First thread offers again, so reset(5000) publishes gen 4 with start 5000, and poll takes gen 4. Now the parked thread wakes, sees gen 4 is owned, passes the check, writes start back to 1000, and moves gen 4 to gen 6 leasable. Idle clock jumps back four seconds, and the entry goes leasable via a thread that never owned it. I did reproduce this.

With a plain CAS and no monitor the loser just fails and nothing happens. The lock is what turns that into a quiet wrong answer.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
```diff
@@ -293,52 +295,68 @@ private static final class ChannelCreation {
...
+            long stateSnapshot = state;
+            if (!isOwned(stateSnapshot)) {
+                return false;
+            }
+            long nextGeneration = (stateSnapshot & ~OWNED_MASK) + 2;
             start = now;
-            owned = 0;
+            return STATE_UPDATER.compareAndSet(this, stateSnapshot, nextGeneration);
```
hyperxpro: Small thing, this CAS cannot fail. Only reset takes state out of owned and the monitor serializes those, and tryTakeOwnership needs an unowned snapshot so it can never match. I ran four threads at it for eight seconds and got 58 million resets with zero CAS failures, so the boolean is always true unless the guard above catches it.

The lock is not free though. On JDK 25 a take plus reset cycle went from 20.7 ns to 52.9 ns, roughly 22 ns of that being the monitor, which more than eats the fence you saved by dropping the intermediate store.

Guard, then start = now, then a plain volatile store of the next generation gets you the same thing for two stores.

client/src/test/java/org/asynchttpclient/netty/channel/DefaultChannelPoolTest.java
```diff
@@ -245,6 +262,46 @@ public void channelReofferedAfterExpiryIsNotReaped() throws Exception {
...
+            assertTrue(channel.awaitActiveCheck(), "cleaner must reach the active-channel check");
+            assertSame(channel, pool.poll(KEY));
+            assertTrue(pool.offer(channel, KEY));
+        } finally {
+            channel.resumeActiveCheck();
+        }
+        cleanerRun.get(5, TimeUnit.SECONDS);
+
+        assertTrue(channel.isActive(), "cleaner must not close a freshly re-offered channel");
+        assertEquals(1, partitionSize(pool, KEY), "cleaner must not unlink the fresh generation");
```
hyperxpro: I do not think this one can fail. There is only one node in the deque, poll unlinks it, and the re offer makes a fresh node the cleaner's cursor never reaches, so the iterator remove is a no op on something already unlinked. I checked by making the tryTakeOwnership failure branch call it.remove as well, and the test still passed.

If you want it to actually bite, put a second untouched channel ahead of the blocked one so the iterator is still on a live node when it resumes.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
pavel-ptashyts: Agreed. I moved partition creation after the idle-generation transfer and simplified it to a direct computeIfAbsent. A duplicate offer now returns before touching partitions, so it cannot leave an empty deque for the cleaner.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
pavel-ptashyts: Agreed, and I reproduced this ordering. I removed synchronized and added a reset-in-progress bit. reset now CASes the exact owned snapshot to the resetting state before writing the timestamp, then publishes the next leasable generation. A delayed caller with a stale snapshot fails without retimestamping or transferring the newer generation.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
pavel-ptashyts: Removed the monitor. I kept a meaningful CAS by using it to reserve the reset-in-progress state; unlike the CAS under the monitor, it can fail for a duplicate or stale caller. This avoids the monitor cost while preventing competing writers from changing the timestamp before winning the transfer.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
pavel-ptashyts: Agreed. I now treat a duplicate as an accepted no-op and return true, so ChannelManager will not close a live pooled channel. The generation and original partition remain unchanged, and the test verifies that the channel stays active and leasable.

client/src/main/java/org/asynchttpclient/netty/channel/DefaultChannelPool.java
pavel-ptashyts: Updated the ownership documentation to cover the initial state, poll leases, removeAll tombstones, cleaner pre-close claims, and the reset-in-progress state. A rejected generation transfer now emits a rate-limited debug message, while the public offer is accepted as a no-op instead of causing a close.

client/src/test/java/org/asynchttpclient/netty/channel/DefaultChannelPoolTest.java
pavel-ptashyts: Added a second untouched channel ahead of the blocked channel and switched this test pool to FIFO, so the cleaner cursor traverses a live neighbor while the blocked channel is leased and re-offered. The test now verifies that both entries remain and are leasable in the expected order.

hyperxpro:
This is going to be a lot of back and forth since this is a complicated - I will close and handle it.

PR #2301, opened 2026-08-03
Skip redirect set lookup for non-3xx responses

pavel-ptashyts:
## Problem

`Interceptors.exitAfterIntercept` runs for every HTTP response and tested the
redirect status like this:

```java
if (Redirect30xInterceptor.REDIRECT_STATUSES.contains(statusCode)) {
```

`REDIRECT_STATUSES` is a `Set<Integer>` and `statusCode` is an `int`, so the
call autoboxes. `Integer.valueOf` caches -128..127, which covers 100..103 but
nothing else a response carries, so every 2xx / 4xx / 5xx response allocated a
fresh `Integer` and hashed it, only to answer `false`. Those are almost all of
the traffic.

## Change

Move the test into a package-private `Redirect30xInterceptor.isRedirect(int)`,
next to the set it guards, and gate the set lookup behind Netty's
`HttpStatusClass.REDIRECTION.contains(int)`. That check takes a primitive, so
only a genuine 3xx pays for the boxing.

The set is still consulted rather than inlined as a `switch` over the five
known codes. `REDIRECT_STATUSES` is `public` and a mutable `HashSet`; changing
what it means is out of scope for a performance change, so any 3xx a caller
registered keeps working. Behaviour differs only for a non-3xx code added to
the set, which would not be a redirect status.

No public API change.

## Scope

One predicate, its two call sites, and a unit test. Related per-response
allocations found in the same method (`responseHeaders.getAll(SET_COOKIE)`
allocates a `LinkedList` per response even with no `Set-Cookie` present) are
left for a separate PR.

## Tests

`Redirect30xInterceptorTest` covers `isRedirect(int)` directly: the five
followed statuses, the 3xx that are not followed (300, 304, 305, 306, 399 -
304 above all, which must reach the normal response path), and non-3xx codes.

The redirect paths themselves are already covered by `Relative302Test`,
`PerRequestRelative302Test`, `PostRedirectGetTest`, `RedirectBodyTest`,
`HttpToHttpsRedirectTest`, `RedirectCredentialSecurityTest`,
`RedirectConnectionUsageTest`, `Head302Test`,
`StripAuthorizationOnRedirectHttpTest` and `ws.RedirectTest`.

## Verification

Follows the same review pass as #2300.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
@@ -125,11 +125,7 @@ public class Interceptors {
             return continue100Interceptor.exitAfterHandling100(channel, future);
         }
 
-        // Range check first: REDIRECT_STATUSES is a Set<Integer>, so contains(statusCode) boxed a fresh
-        // Integer on every response (status codes are outside Integer's valueOf cache). The set is public
-        // and mutable, so the lookup is kept rather than inlined as a switch, and a caller that registered
-        // an extra 3xx status still has it honoured.
-        if (statusCode >= 300 && statusCode < 400 && Redirect30xInterceptor.REDIRECT_STATUSES.contains(statusCode)) {
+        if (Redirect30xInterceptor.isRedirect(statusCode)) {
             return redirect30xInterceptor.exitAfterHandlingRedirect(channel, future, response, request, statusCode, realm);
         }
 
--- a/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
@@ -125,7 +125,11 @@ public class Interceptors {
             return continue100Interceptor.exitAfterHandling100(channel, future);
         }
 
-        if (Redirect30xInterceptor.REDIRECT_STATUSES.contains(statusCode)) {
+        // Range check first: REDIRECT_STATUSES is a Set<Integer>, so contains(statusCode) boxed a fresh
+        // Integer on every response (status codes are outside Integer's valueOf cache). The set is public
+        // and mutable, so the lookup is kept rather than inlined as a switch, and a caller that registered
+        // an extra 3xx status still has it honoured.
+        if (statusCode >= 300 && statusCode < 400 && Redirect30xInterceptor.REDIRECT_STATUSES.contains(statusCode)) {
             return redirect30xInterceptor.exitAfterHandlingRedirect(channel, future, response, request, statusCode, realm);
         }
 
--- a/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
+++ b/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
@@ -18,6 +18,7 @@ package org.asynchttpclient.netty.handler.intercept;
 import io.netty.channel.Channel;
 import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpResponse;
+import io.netty.handler.codec.http.HttpStatusClass;
 import io.netty.handler.codec.http.HttpUtil;
 import io.netty.handler.codec.http.cookie.Cookie;
 import org.asynchttpclient.AsyncHttpClientConfig;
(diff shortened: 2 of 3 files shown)


client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
```diff
@@ -125,7 +125,11 @@ public boolean exitAfterIntercept(Channel channel, NettyResponseFuture<?> future
             return continue100Interceptor.exitAfterHandling100(channel, future);
         }
 
-        if (Redirect30xInterceptor.REDIRECT_STATUSES.contains(statusCode)) {
+        // Range check first: REDIRECT_STATUSES is a Set<Integer>, so contains(statusCode) boxed a fresh
+        // Integer on every response (status codes are outside Integer's valueOf cache). The set is public
+        // and mutable, so the lookup is kept rather than inlined as a switch, and a caller that registered
+        // an extra 3xx status still has it honoured.
```
hyperxpro: Can we drop this one? It describes the change rather than the code, and once the old version is out of memory it just reads as noise. It is also four lines of prose for a one line predicate, where the rest of the file only comments invariants. If you want something here, one line does it: only a 3xx can be a redirect, so the range check keeps the boxed lookup off the common path.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
```diff
@@ -125,7 +125,11 @@ public boolean exitAfterIntercept(Channel channel, NettyResponseFuture<?> future
...
+        if (statusCode >= 300 && statusCode < 400 && Redirect30xInterceptor.REDIRECT_STATUSES.contains(statusCode)) {
```
hyperxpro: Netty already has this one. HttpStatusClass.REDIRECTION.contains(statusCode) takes an int and uses the same 300/400 bounds. We do not use HttpStatusClass anywhere else yet, so take it or leave it, the explicit range reads fine.

hyperxpro: Minor: the range now lives here while the set lives in Redirect30xInterceptor, so the knowledge is split across two classes. A static isRedirect(int) next to the set would keep it in one place and give you something to test. Fine to skip if you want the diff minimal.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
pavel-ptashyts: Dropped. It was describing the change rather than the code, and with the predicate moved into Redirect30xInterceptor the call site reads on its own, so there is nothing left worth saying there.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Interceptors.java
pavel-ptashyts: Taken. REDIRECTION is built with 300/400 bounds and contains takes an int, so it keeps the boxing off non-3xx responses exactly as the open-coded range did, and it names what the bounds mean instead of leaving two magic numbers. First use of HttpStatusClass in the codebase, but that seems like a reason to start rather than not to.

pavel-ptashyts: Done. Package-private isRedirect(int) next to the set, so Interceptors just asks the question and the range and the statuses stay together.

Added Redirect30xInterceptorTest for it. The case that earns the test is 304: in the 3xx class, not a followed redirect, so it has to be rejected exactly like a non-3xx status. Neither half of the predicate says that on its own.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -70,6 +71,16 @@ public class Redirect30xInterceptor {
         REDIRECT_STATUSES.add(PERMANENT_REDIRECT_308);
     }
 
+    /**
+     * Whether {@code statusCode} is a redirect this interceptor follows. Only a 3xx can be, and the class
+     * check takes an {@code int}, so it keeps the {@link #REDIRECT_STATUSES} lookup, which boxes, off the
+     * responses that make up almost all traffic. The set is still consulted rather than inlined here because
+     * it is public and mutable, so an extra 3xx status a caller registered stays honoured.
+     */
```
hyperxpro, line 79: Optional. The "still consulted rather than inlined here" phrasing argues against the version we did not write, rather than describing the method. Reads a bit cleaner as plain contract:

```suggestion
    /**
     * Whether {@code statusCode} is a redirect this interceptor follows. Only a 3xx can be, and the class
     * check takes an {@code int}, so the boxing {@link #REDIRECT_STATUSES} lookup stays off the 2xx, 4xx
     * and 5xx responses that carry almost all traffic. The set is consulted rather than inlined because it
     * is public and mutable, so an extra 3xx a caller registered is honoured.
     */
```

hyperxpro (changes requested):

hyperxpro (approved):

PR #2303, opened 2026-08-03
Decode a lone response body part in place

pavel-ptashyts:
## Problem

For the default handler (`executeRequest(request)` -> `AsyncCompletionHandlerBase`
-> `Response`) the body is copied three times:

1. `HttpHandler.handleChunk` -> `EagerResponseBodyPart` copies each chunk out of
   the network buffer into a heap `byte[]` (needed: `channelRead` releases the
   message afterwards);
2. `NettyResponse.getResponseBodyAsByteBuffer` concatenates every part into a
   freshly allocated array;
3. `getResponseBody(charset)` decodes that array.

Step 2 is pure waste when there is only one part, which is the case for any body
that lands in a single socket read: it copies a single part into a new array with
nothing to concatenate it with.

## Change

`getResponseBody(Charset)` decodes straight from the part when there is exactly
one of them. The array does not escape the method, so the part's own array can be
decoded in place.

Several parts are still concatenated before decoding, never decoded one at a
time, because a multi-byte character can straddle a part boundary.

`getResponseBodyAsBytes` and `getResponseBodyAsByteBuffer` are deliberately left
untouched: they hand the array to the caller, so they keep making a defensive
copy rather than expose a part's own array. There is no aliasing change anywhere
in this PR.

## Measurements

Rough probe on JDK 17 over a single-part ASCII body, concatenate-then-decode
versus decode-in-place. Not JMH, so read the shape rather than the digits:

| body | before | after |
|------|--------|-------|
| 512 B | 496 ns | 47 ns |
| 4 KB | 2277 ns | 373 ns |
| 16 KB | 2963 ns | 1450 ns |
| 128 KB | 24913 ns | 12248 ns |

Plus one fewer whole-body allocation per response. The percentages look large
partly because a pure-ASCII body decodes through a JDK intrinsic, which makes the
removed copy a big share of what is left.

## Tests

Two added to `NettyAsyncResponseTest`:

* `testGetResponseBodyDecodesOnePartAndSplitPartsIdentically` splits the two-byte
  UTF-8 encoding of U+00E9 across two parts and asserts one-part and split-part
  bodies decode alike. This pins the constraint the comment states: it fails if
  anyone later makes the multi-part path decode part by part.
* `testGetResponseBodyAsBytesDoesNotShareTheBodyPartArray` pins that
  `getResponseBodyAsBytes` still returns a fresh array and never the part's own.

The body bytes are built as an explicit `byte[]` rather than a string literal to
keep the source ASCII per `AGENTS.md`.

## Verification

`mvnw clean verify` - BUILD SUCCESS, 1373 tests (1371 before, plus these two),
0 failures, 0 errors, 19 skipped. Error Prone, NullAway and Revapi clean.
`LargeResponseTest`, `NoNullResponseTest`, `BodyDeferringAsyncHandlerTest` and
`RedirectBodyTest`, which exercise the multi-part path, are green.

## Not in scope

The multi-part case still concatenates. A `CompositeByteBuf.toString(charset)`
variant measured faster there (Netty decodes a multi-component buffer through a
recycled, un-zeroed thread-local array instead of a fresh `byte[]`), but it
regressed at high part counts in the same probe, so it needs proper benchmarking
before it becomes a change. Removing copy 1 would mean retaining network buffers
and giving `Response` a lifecycle, which is public API and wants a design
discussion first.

No public API change here. Same review pass as #2300, #2301 and #2302.

### DIFF
--- a/client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
+++ b/client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
@@ -241,8 +241,7 @@ public class NettyResponse implements Response {
 
     @Override
     public InputStream getResponseBodyAsStream() {
-        // ByteArrayInputStream never exposes the array it wraps, so it can read the part's own.
-        return new ByteArrayInputStream(sharedBodyBytes());
+        return new ByteArrayInputStream(getResponseBodyAsBytes());
     }
 
     @Override
--- a/client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
+++ b/client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
@@ -226,6 +226,14 @@ public class NettyResponse implements Response {
 
     @Override
     public String getResponseBody(Charset charset) {
+        // Decode a lone body part straight from its own bytes. getResponseBodyAsBytes concatenates every part
+        // into a fresh array first, so a body that arrived in a single read was copied twice, once to
+        // concatenate and once to decode. The array does not escape this method, so decoding the part's own
+        // one is safe. Several parts are still concatenated before decoding rather than decoded one at a
+        // time, because a multi-byte character can straddle a part boundary.
+        if (bodyParts.size() == 1) {
+            return new String(bodyParts.get(0).getBodyPartBytes(), charset);
+        }
         return new String(getResponseBodyAsBytes(), charset);
     }
 
--- a/client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
+++ b/client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
@@ -224,22 +224,25 @@ public class NettyResponse implements Response {
         return getResponseBody(withDefault(extractContentTypeCharsetAttribute(getContentType()), UTF_8));
     }
 
+    /**
+     * The body as bytes, for callers that keep the array to themselves. A lone part's own array is returned
+     * rather than a copy of it, so a caller that let it out would let the part's buffer be mutated through it;
+     * {@link #getResponseBodyAsBytes()} is the copying variant for those. Several parts are concatenated
+     * because a multi-byte character can straddle a part boundary.
+     */
+    private byte[] sharedBodyBytes() {
+        return bodyParts.size() == 1 ? bodyParts.get(0).getBodyPartBytes() : getResponseBodyAsBytes();
+    }
+
     @Override
     public String getResponseBody(Charset charset) {
-        // Decode a lone body part straight from its own bytes. getResponseBodyAsBytes concatenates every part
-        // into a fresh array first, so a body that arrived in a single read was copied twice, once to
-        // concatenate and once to decode. The array does not escape this method, so decoding the part's own
-        // one is safe. Several parts are still concatenated before decoding rather than decoded one at a
-        // time, because a multi-byte character can straddle a part boundary.
-        if (bodyParts.size() == 1) {
-            return new String(bodyParts.get(0).getBodyPartBytes(), charset);
-        }
-        return new String(getResponseBodyAsBytes(), charset);
+        return new String(sharedBodyBytes(), charset);
     }
 
     @Override
     public InputStream getResponseBodyAsStream() {
-        return new ByteArrayInputStream(getResponseBodyAsBytes());
+        // ByteArrayInputStream never exposes the array it wraps, so it can read the part's own.
+        return new ByteArrayInputStream(sharedBodyBytes());
     }
 
     @Override
--- a/client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
+++ b/client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
@@ -22,6 +22,7 @@ import org.asynchttpclient.HttpResponseBodyPart;
 import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
+import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
 import java.util.Date;
@@ -138,5 +139,28 @@ public class NettyAsyncResponseTest {
         // caller, so it must keep copying rather than expose the part's own array.
         assertNotSame(response.getResponseBodyAsBytes(), response.getResponseBodyAsBytes());
         assertNotSame(bodyParts.get(0).getBodyPartBytes(), response.getResponseBodyAsBytes());
+    }
+
+    @Test
+    public void testGetResponseBodyAsStreamDoesNotShareTheBodyPartArray() throws IOException {
+        List<HttpResponseBodyPart> bodyParts = new LinkedList<>();
+        bodyParts.add(new EagerResponseBodyPart(Unpooled.wrappedBuffer("Hello World".getBytes(StandardCharsets.UTF_8)), true));
+        NettyResponse response = new NettyResponse(new NettyResponseStatus(null, null, null), null, bodyParts);
+
+        // On JDK 11 ByteArrayInputStream.transferTo passes its own array to the OutputStream, so a stream over
+        // a part's array would put that array in the caller's hands.
+        byte[][] handedOut = new byte[1][];
+        response.getResponseBodyAsStream().transferTo(new OutputStream() {
+            @Override
+            public void write(int b) {
+            }
+
+            @Override
+            public void write(byte[] b, int off, int len) {
+                handedOut[0] = b;
+            }
+        });
+
+        assertNotSame(bodyParts.get(0).getBodyPartBytes(), handedOut[0]);
     }
 }
--- a/client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
+++ b/client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
@@ -21,6 +21,7 @@ import io.netty.handler.codec.http.cookie.Cookie;
 import org.asynchttpclient.HttpResponseBodyPart;
 import org.junit.jupiter.api.Test;
 
+import java.io.IOException;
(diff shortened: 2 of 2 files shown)


client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
```diff
@@ -226,6 +226,14 @@ public String getResponseBody() {
 
     @Override
     public String getResponseBody(Charset charset) {
+        // Decode a lone body part straight from its own bytes. getResponseBodyAsBytes concatenates every part
+        // into a fresh array first, so a body that arrived in a single read was copied twice, once to
+        // concatenate and once to decode. The array does not escape this method, so decoding the part's own
+        // one is safe. Several parts are still concatenated before decoding rather than decoded one at a
+        // time, because a multi-byte character can straddle a part boundary.
```
hyperxpro: Five lines of comment in a file that has none. The middle sentence is about what the old code did, which is commit message material. Two lines carry everything a reader needs:

    // A lone part's array does not escape this method, so it can be decoded in place. Several parts are
    // concatenated first because a multi-byte character can straddle a part boundary.

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
```diff
@@ -226,6 +226,14 @@ public String getResponseBody() {
...
+        if (bodyParts.size() == 1) {
```
hyperxpro: getResponseBodyAsStream a few lines below hands the same array to a ByteArrayInputStream, which never exposes what it wraps, so it can take this shortcut too. Fine to leave for a follow up. If you do both, put the guard in a small private helper instead of repeating it.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
```diff
@@ -90,4 +92,36 @@ public void testGetResponseBodyAsByteBuffer() {
...
+
+    @Test
+    public void testGetResponseBodyDecodesOnePartAndSplitPartsIdentically() {
+        byte[] utf8 = {'c', 'a', 'f', (byte) 0xC3, (byte) 0xA9, ' ', (byte) 0xC3, (byte) 0xBC, 'b', 'e', 'r'};
+        String expected = new String(utf8, StandardCharsets.UTF_8);
+        // 0xC3 0xA9 encodes U+00E9; split between its two bytes so neither half decodes on its own
+        int split = 4;
+
+        List<HttpResponseBodyPart> onePart = new LinkedList<>();
+        onePart.add(new EagerResponseBodyPart(Unpooled.wrappedBuffer(utf8), true));
```
hyperxpro, line 106: Both new tests use Eager parts. Lazy is the case where this is least obvious, since its getBodyPartBytes returns only the readable region rather than the whole backing array. Worth one more case with a Lazy part over a slice, say Unpooled.wrappedBuffer(backing, 3, 11), so that a later change to getBodyByteBuf().array() cannot slip through. testGetResponseBodyAsByteBuffer above already uses Lazy parts, so it fits the file.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
```diff
@@ -90,4 +92,36 @@ public void testGetResponseBodyAsByteBuffer() {
...
+
+    @Test
+    public void testGetResponseBodyAsBytesDoesNotShareTheBodyPartArray() {
+        List<HttpResponseBodyPart> bodyParts = new LinkedList<>();
+        bodyParts.add(new EagerResponseBodyPart(Unpooled.wrappedBuffer("Hello World".getBytes(StandardCharsets.UTF_8)), true));
+        NettyResponse response = new NettyResponse(new NettyResponseStatus(null, null, null), null, bodyParts);
+
+        // getResponseBody may decode a lone part in place, but getResponseBodyAsBytes hands the array to the
+        // caller, so it must keep copying rather than expose the part's own array.
+        assertNotSame(response.getResponseBodyAsBytes(), response.getResponseBodyAsBytes());
```
hyperxpro, line 140: Good one to have. This is what catches the tempting wrong version of this change, pushing the fast path down into getResponseBodyAsBytes, because Eager hands back the same array every time.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
pavel-ptashyts: Cut to two lines. The sentences about what the old code did went into the commit message, and what remains sits on the private helper, since that is where the branch and the constraint on it now live.

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
pavel-ptashyts: Did both. The guard is in a private sharedBodyBytes(), named to carry the constraint to the call sites: the array it returns may be a part's own, so only callers that keep it to themselves may use it. getResponseBodyAsBytes and getResponseBodyAsByteBuffer still copy.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
pavel-ptashyts: Added, with your wrappedBuffer(backing, 3, 11). Good catch. A version reaching for getBodyByteBuf().array() would read the surrounding bytes too, and nothing else in the suite would have noticed. The case asserts the stream path as well, now that it shares the shortcut.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
pavel-ptashyts: Kept for exactly that reason. Eager handing back the same array every time makes the wrong version of this look free.

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
```diff
@@ -224,14 +224,25 @@ public String getResponseBody() {
...
     public String getResponseBody(Charset charset) {
-        return new String(getResponseBodyAsBytes(), charset);
+        return new String(sharedBodyBytes(), charset);
     }
 
     @Override
     public InputStream getResponseBodyAsStream() {
-        return new ByteArrayInputStream(getResponseBodyAsBytes());
+        // ByteArrayInputStream never exposes the array it wraps, so it can read the part's own.
+        return new ByteArrayInputStream(sharedBodyBytes());
```
hyperxpro: This is the part to undo. transferTo on 11 will pass this array straight to a caller supplied OutputStream, so the stream cannot read the part's own.

```suggestion
        return new ByteArrayInputStream(getResponseBodyAsBytes());
```

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
```diff
@@ -19,7 +19,9 @@
 import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.cookie.Cookie;
 import org.asynchttpclient.HttpResponseBodyPart;
+import org.junit.jupiter.api.Test;
 
+import java.io.IOException;
```
hyperxpro, line 24: Needed by the test below.

```suggestion
import java.io.IOException;
import java.io.OutputStream;
```

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
```diff
@@ -90,4 +93,50 @@ public void testGetResponseBodyAsByteBuffer() {
...
+        List<HttpResponseBodyPart> bodyParts = new LinkedList<>();
+        bodyParts.add(new EagerResponseBodyPart(Unpooled.wrappedBuffer("Hello World".getBytes(StandardCharsets.UTF_8)), true));
+        NettyResponse response = new NettyResponse(new NettyResponseStatus(null, null, null), null, bodyParts);
+
+        // getResponseBody may decode a lone part in place, but getResponseBodyAsBytes hands the array to the
+        // caller, so it must keep copying rather than expose the part's own array.
+        assertNotSame(response.getResponseBodyAsBytes(), response.getResponseBodyAsBytes());
+        assertNotSame(bodyParts.get(0).getBodyPartBytes(), response.getResponseBodyAsBytes());
+    }
 }
```
hyperxpro, line 166: A companion to the getResponseBodyAsBytes one, so the stream path cannot quietly start sharing later. Worth knowing that it only bites on 11: from 17 or so the JDK copies through a temp buffer for an untrusted OutputStream, so the assertion passes there whether the code is right or not. 11 is in the matrix, which is what matters.

```suggestion
    }

    @Test
    public void testGetResponseBodyAsStreamDoesNotShareTheBodyPartArray() throws IOException {
        List<HttpResponseBodyPart> bodyParts = new LinkedList<>();
        bodyParts.add(new EagerResponseBodyPart(Unpooled.wrappedBuffer("Hello World".getBytes(StandardCharsets.UTF_8)), true));
        NettyResponse response = new NettyResponse(new NettyResponseStatus(null, null, null), null, bodyParts);

        // On JDK 11 ByteArrayInputStream.transferTo passes its own array to the OutputStream, so a stream over
        // a part's array would put that array in the caller's hands.
        byte[][] handedOut = new byte[1][];
        response.getResponseBodyAsStream().transferTo(new OutputStream() {
            @Override
            public void write(int b) {
            }

            @Override
            public void write(byte[] b, int off, int len) {
                handedOut[0] = b;
            }
        });

        assertNotSame(bodyParts.get(0).getBodyPartBytes(), handedOut[0]);
    }
}
```

hyperxpro (changes requested):

hyperxpro (approved):

PR #2313, opened 2026-08-17
Arm request timeouts on an event loop

pavel-ptashyts:
## Problem

Request and read timeouts are armed on the client's `HashedWheelTimer`. That has two
properties that only show up on short deadlines:

* **A wheel quantizes.** It fires on the first tick at or after the deadline, so a
  deadline near or below `hashedWheelTimerTickDuration` is rounded up to it.
* **One thread carries every expiry** for the whole client, and `HashedWheelTimer`'s
  default `taskExecutor` is `ImmediateExecutor`, so each expiry runs inline on the wheel
  thread - including `future.completeExceptionally(...)` and therefore whatever the caller
  chained onto the response future.

On a one-second budget the first costs 0.3% and nobody notices. On a budget of tens of
milliseconds a tick is a large fraction of it, and a burst of expiries has no headroom to
absorb before the wheel starts running late.

## Measured

2000 timeouts armed as one burst on Netty 4.2.16, JDK 17, tasks doing nothing but
recording their own lag. This is the floor; real work on the firing thread only adds to it.

| instrument | deadline | mean | p50 | p99 | max |
|---|---:|---:|---:|---:|---:|
| wheel, tick 5 ms | 20 ms | +2.7 | +2 | +5 | +5 |
| wheel, tick 1 ms | 20 ms | +1.3 | +1 | +2 | +2 |
| `EventLoop.schedule` | 20 ms | +0.0 | +0 | +0 | +0 |
| wheel, tick 5 ms | 1000 ms | +3.0 | +3 | +3 | +3 |
| wheel, tick 1 ms | 1000 ms | +1.9 | +2 | +2 | +2 |
| `EventLoop.schedule` | 1000 ms | +1.6 | +2 | +2 | +2 |

An event loop shows zero overshoot because it schedules by deadline and derives its own
`select()` timeout from the nearest one. There is no quantum to round to.

This was a throwaway probe rather than JMH - `client/src/jmh/java` is not currently wired
into the build, so its benchmarks do not compile. Happy to add a proper benchmark if that
is fixed first, or as part of this.

## Change

`AsyncHttpClientConfig#isUseEventLoopTimeouts()`, **off by default**, arms the request and
read timeouts on an event loop instead of the timer.

* **Always the channel's own loop.** On the pooled path the channel is already in hand, so
  its loop is used and the timeout expires on the thread that would have to close it. On the
  connect path there is no channel yet - deliberately, so that the timeout also bounds
  address resolution and the connect - so it is armed on the timer and moved onto the loop
  once the connect succeeds. No other loop is ever used; see the review round below for why
  that matters.
* **Arming allocates nothing extra.** The cancellation handle lives on the task rather than
  in a wrapper, and the existing `done` flag stands in for the scheduler's already-expired
  flag, which the two schedulers spell differently.
* **Shutdown race closed.** `isShuttingDown()` can return false and `schedule` reject
  immediately after. Netty answers a rejected timeout with a logged warning rather than an
  exception, which would leave the exchange with nothing to end it, so a rejection falls
  back to the timer.
* The connection-pool cleaner stays on the timer either way.

Off by default because the expiry - and so whatever the caller chained onto the future -
then runs on an I/O thread, and blocking one stalls every connection it serves. The javadoc
says so and points callers at `handleAsync`.

## Why not a wheel per event loop

That is how the Aerospike client solves the same problem: `EventLoopBase` owns a
`HashedWheelTimer` that is a `Runnable` the loop ticks itself. Deliberately not copied here.
A wheel arms in O(1) against O(log n) for a deadline queue, but at a few thousand timeouts
per loop that is a dozen comparisons, while the quantization it reintroduces costs
milliseconds on a 20 ms budget - the third row above is the whole point. A wheel also has to
be ticked forever, waking every loop even with nothing armed. Aerospike wrote its own because
its `EventLoop` abstracts over NIO, Netty and direct NIO and needed one timer; AHC is
Netty-only and gets a per-loop deadline queue for free.

## Review round 1

Most of the substance of this PR changed in review, so the sections above describe the
current shape rather than what was first pushed. Two things are worth calling out here
because they were design errors, not polish:

**The loop is now only ever the channel's own.** It used to come from
`EventLoopGroup#next()`, which is almost never the loop the channel ends up on:
`initAndRegister` draws from the same chooser, so the two agreed about one time in N. Every
completion then cancelled an entry on a foreign loop, and until the original deadline that
entry sat in a queue whose loop it would wake for a request that had long finished. Drawing
from the chooser also shifted which loops connections land on. The pooled path has the
channel in hand; the connect path arms on the timer, as before this branch, and
`NettyConnectListener` moves the timeouts onto the loop once the connect succeeds, next to
the `attachChannel` that publishes it on the future. That listener already runs on the
channel's loop, so the move costs a same-thread schedule and no wakeup.

**Arming left the `TimeoutsHolder` constructor.** The task holds the holder and can run the
moment it is armed, and an event loop does not round a short deadline up to a tick, so the
expiry could reach a holder whose fields were not yet frozen, a future that had not been
handed the holder, and on the pooled path a future with no channel attached - which aborted
with `null` and left the pooled socket open. The caller now publishes the holder, attaches
the channel, and calls `start()` last.

Also from the review: `arm` re-checks `cancelled` after recording its handle, so an exchange
that finishes mid-arming cannot leave behind an entry nobody will cancel; `cancelArmed`
catches the `RejectedExecutionException` that Netty's off-loop cancellation path can raise on
a closing client, which had never escaped `ListenableFuture#cancel` before; the cancellation
handle is two typed fields rather than an `Object` and `instanceof`; `requestTimeoutArmed` is
gone in favour of a null test on the task; the rationale lives on
`isUseEventLoopTimeouts()` alone; and the new option sits in the `// timeouts` group
everywhere rather than splitting the two `failedIpCooldown` entries.

## API compatibility

No `revapi` entries. `implements Runnable` and `run()` live on the two subclasses rather than
on `TimeoutTimerTask`, which leaves that class's surface unchanged: both subclasses already
declared `run(Timeout)` without a throws clause, so inside a subclass `run()` calls its own
override and has nothing to catch. No dead handler, and nothing narrowed.

The knock-on is that only the concrete classes are both a `TimerTask` and a `Runnable`, so
`TimeoutsHolder#arm` takes that intersection as a type parameter. Everything else is
additive: the existing `TimeoutsHolder` constructor is kept and delegates, and nothing is
removed.

`start()` is called from `NettyResponseFuture#setTimeoutsHolder` rather than by the sender.
`TimeoutsHolder` has a public constructor in an exported package, and splitting the arming
out of it would otherwise leave an outside caller free to install a holder and get an
exchange with no request timeout at all.

## Tests

Four cases in `EventLoopTimeoutTest`, asserting where an expiry is delivered from rather than
what it does. They hand the config their own `Timer` and `EventLoopGroup` so the assertions
are against those objects and not against thread names, which a pool name containing `timer`
or a configured thread factory would have broken with no bug present:

- the timer default, by identity against the timer's own thread;
- a connecting exchange, on the loop of the channel `onTcpConnectSuccess` reported;
- an exchange on a pooled channel, on the loop of the channel `onConnectionPooled` reported,
  which is also what says it reused the connection rather than opening one of its own;
- a read timeout under the new mode, which is armed after the request is written and so
  exercises a different arming path.

The group has eight loops. With two, a timeout armed on the wrong loop is on the right one
half the time, and these assertions would have passed about half the runs against the bug they
exist to catch; at eight, reverting `timeoutExecutor` to `next()` fails the pooled case.

The connecting case, and the pooled case's first request, get a one second budget on purpose.
A deadline reached before connecting would be delivered from the timer quite correctly, there
being no channel to deliver it from, and would prove nothing either way; and the pooled case's
first request is the cold one - class loading, the connect, the server's first response - and
is meant to succeed.

One gap, called out rather than papered over: the `RejectedExecutionException` fallback in
`arm` has no test. Reaching it needs a loop that answers `isShuttingDown()` with `false` and
then rejects the schedule, and the executor comes from the channel, so there is no way in
through the config. A test double for `EventExecutor` would do it if that is acceptable.

## Verification

`mvnw clean verify` - BUILD SUCCESS, 1468 tests, 0 failures, 0 errors, 21 skipped. Error
Prone, NullAway clean; `revapi` clean with the one scoped entry above.

installed on this machine, so it was run on **JDK 17** (also in the CI matrix). The JDK 11
leg of CI on this PR is the real gate.


client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutTimerTask.java
```diff
@@ -15,31 +15,84 @@
...
+     * Cancels the scheduled entry this task was armed on, if any. Never interrupts: on the event-loop path the
+     * task may be running on the very thread this is called from, and nothing in it answers interruption.
+     */
+    void cancelArmed() {
+        Object handle = armed;
+        armed = null;
+        if (handle instanceof Timeout) {
+            ((Timeout) handle).cancel();
+        } else if (handle instanceof Future) {
+            ((Future<?>) handle).cancel(false);
```
hyperxpro: arm() catches RejectedExecutionException, this doesn't. Netty's ScheduledFutureTask.cancel goes through removeScheduled, which for an off-loop caller lazyExecutes a removal task, and offerTask rejects once the loop is shut down. So after a client.close() any late future.cancel(true) or abort() throws out of ListenableFuture.cancel(), which never threw before. The same try/catch arm() already has would settle it.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -60,10 +91,12 @@ public TimeoutsHolder(Timer nettyTimer, NettyResponseFuture<?> nettyResponseFutu
 
         if (requestTimeoutInMs > -1) {
             requestTimeoutMillisTime = unpreciseMillisTime() + requestTimeoutInMs;
-            requestTimeout = newTimeout(new RequestTimeoutTimerTask(nettyResponseFuture, requestSender, this, requestTimeoutInMs), requestTimeoutInMs);
+            requestTimeoutTask = new RequestTimeoutTimerTask(nettyResponseFuture, requestSender, this, requestTimeoutInMs);
+            requestTimeoutArmed = arm(requestTimeoutTask, requestTimeoutInMs);
```
hyperxpro: This arms a task holding this from inside the constructor, so before requestTimeoutTask and requestTimeoutArmed are frozen and before the future has been handed the holder. At the pooled call site it is also before attachChannel. The wheel's 100 ms tick hid that window, an event loop won't: with a small enough request timeout the task can run first, find channel null in expire() and leave the pooled socket open, and setTimeoutsHolder then installs a holder that has already fired. Arming after the constructor returns closes it.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -1080,12 +1081,39 @@ private static void configureTransferAdapter(AsyncHandler<?> handler, HttpReques
...
+     * the whole client and a tick the deadline is rounded up to, not the identity of the thread.
+     */
+    private @Nullable EventExecutor timeoutExecutor(@Nullable Channel channel) {
+        if (!config.isUseEventLoopTimeouts()) {
+            return null;
+        }
+        if (channel != null) {
+            return channel.eventLoop();
+        }
+        return channelManager.getEventLoopGroup().next();
```
hyperxpro: This is almost never the loop the channel ends up on. initAndRegister calls next() again, so for an N loop group the affinity the javadoc above claims holds about one time in N. Every completion then cancels a scheduled entry on a foreign loop, and that entry sits in its queue waking it at the original request timeout deadline long after the request finished. Read timeout re-arms cross loops too. NettyConnectListener.onSuccess already has both the channel and the holder, it sets the resolved address there, so re-homing at that point would make the claim true on the connect path as well.

hyperxpro: Separately: calling next() only to pick a timeout thread advances the group's round robin counter, and that is the same counter that assigns channels to loops. With a power of two group and a fixed number of next() calls per request, registrations can settle onto a subset of the loops. Whether that helps or hurts depends on whether an address resolver group is configured, so it is action at a distance either way. Picking with ThreadLocalRandom over the group's executors, or re-homing once the channel exists, leaves the chooser alone.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -98,24 +133,47 @@ void startReadTimeout(ReadTimeoutTimerTask task) {
...
         // cases where a channel's remote address wasn't available. In such cases
         // avoid scheduling any timeouts rather than throwing a NPE.
-        if (requestSender == null || nettyTimer == null || requestSender.isClosed()) {
-            return null;
+        if (requestSender == null || requestSender.isClosed()) {
+            return false;
+        }
+        if (eventExecutor != null && !eventExecutor.isShuttingDown()) {
+            try {
+                task.armedOn(eventExecutor.schedule(task, delay, TimeUnit.MILLISECONDS));
```
hyperxpro: The handle is recorded after schedule() returns. If the exchange completes in that gap, cancel() wins the CAS, release() cancels the previous handle, and then armedOn writes the live one with nobody left to cancel it. cancelled is one shot, so nobody ever will. No wrong abort, since the task no-ops on done, but the entry stays in that loop's queue until the full deadline. Re-checking cancelled after arming and cancelling there closes it.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -81,14 +114,16 @@ public void startReadTimeout() {
         }
     }
 
-    void startReadTimeout(ReadTimeoutTimerTask task) {
-        if (requestTimeout == null || !requestTimeout.isExpired() && readTimeoutValue < requestTimeoutMillisTime - unpreciseMillisTime()) {
+    void startReadTimeout(@Nullable ReadTimeoutTimerTask task) {
+        if (!requestTimeoutArmed
+                || !requestTimeoutTask.isClaimed() && readTimeoutValue < requestTimeoutMillisTime - unpreciseMillisTime()) {
```
hyperxpro, line 151: requestTimeoutTask is declared nullable and this dereferences it, with only requestTimeoutArmed making that safe. And the case the field comment describes cannot happen: arm() returns false only when the sender is closed or there is nothing to arm on, and the read timeout's own arm() then fails identically. So requestTimeoutArmed is always requestTimeoutTask being non null. Testing for null instead keeps the old shape and drops both the field and the unchecked deref. Worth noting NullAway is switched off for this package, so it won't flag it for you.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutTimerTask.java
```diff
@@ -15,31 +15,84 @@
...
     protected final AtomicBoolean done = new AtomicBoolean();
     protected final NettyRequestSender requestSender;
     final TimeoutsHolder timeoutsHolder;
     volatile NettyResponseFuture<?> nettyResponseFuture;
+    /**
+     * The scheduled entry this task is armed on: an {@link Timeout} from a {@link io.netty.util.Timer}, or a
+     * {@link Future} from an event loop. Held here rather than in a wrapper so arming a timeout allocates
+     * nothing beyond what the scheduler itself needs.
+     */
+    private volatile @Nullable Object armed;
```
hyperxpro: Object plus instanceof means an unrecognized handle makes cancellation a silent no-op instead of a compile error. It works today only because Netty's ScheduledFuture happens to extend java.util.concurrent.Future, and Netty 5 already split that hierarchy once. If it ever moves, every cancel quietly stops working and each request leaks a scheduled entry until its deadline, with no signal. Two typed fields, or a Runnable canceller captured at arm time, costs the same allocation and keeps the compiler in the loop.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -98,24 +133,47 @@ void startReadTimeout(ReadTimeoutTimerTask task) {
...
+            task.cancelArmed();
+            task.clean();
+        }
+    }
+
+    /**
+     * Arms {@code task} to run after {@code delay} milliseconds, recording the scheduled entry on the task so it
+     * can cancel itself later.
+     *
+     * @return whether the task was armed. It is not when the client is shutting down, in which case there is no
```
hyperxpro: This isn't what it returns false for. False comes back when the sender is null or closed, or when there is no timer and no usable executor. During shutdown it returns true whenever the executor rejects and the timer fallback below takes it. requestTimeoutArmed is the only reader and startReadTimeout keys off it, so this is the precondition the next reader will trust.

client/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
```diff
@@ -270,6 +270,30 @@ default Duration getFailedIpCooldownPeriod() {
         return Duration.ofSeconds(10);
     }
 
+    /**
+     * Whether request and read timeouts are armed on an event loop rather than on {@link #getNettyTimer()}.
```
hyperxpro: This rationale is now spelled out four times: here, the `TimeoutsHolder` class doc, `timeoutExecutor`'s Javadoc, and the class doc on `TimeoutTimerTask`. Keeping this as the canonical explanation and linking to it from the other three says it once and leaves a single place to update.

client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClientConfig.java
```diff
@@ -63,6 +63,7 @@
 import static org.asynchttpclient.config.AsyncHttpClientConfigDefaults.defaultEnabledProtocols;
 import static org.asynchttpclient.config.AsyncHttpClientConfigDefaults.defaultExpiredCookieEvictionDelay;
 import static org.asynchttpclient.config.AsyncHttpClientConfigDefaults.defaultFailedIpCooldownEnabled;
+import static org.asynchttpclient.config.AsyncHttpClientConfigDefaults.defaultUseEventLoopTimeouts;
```
hyperxpro: This block is alphabetical, so it belongs just before defaultUseInsecureTrustManager, not between the two failedIpCooldown entries. The same split repeats in the field list, the constructor parameters and assignments, the Builder field and copy constructor, the build() call, AsyncHttpClientConfigDefaults, and ahc-default.properties, and the getter lands between isFailedIpCooldownEnabled and getFailedIpCooldownPeriod. The constructor is positional with a lot of adjacent booleans, so order is worth more here than tidiness.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutTimerTask.java
```diff
@@ -15,31 +15,84 @@
...
         this.requestSender = requestSender;
         this.timeoutsHolder = timeoutsHolder;
     }
 
+    @Override
+    public void run() {
+        try {
+            run(null);
+        } catch (Exception e) {
+            // TimerTask#run is declared to throw, and on this entry point the caller is an event loop, where an
```
hyperxpro: The subclass this defends against cannot exist. The only constructor is package private, so nothing outside this package can extend TimeoutTimerTask, and both subclasses are in it and neither throws. Which also means run(Timeout) could just drop its throws clause and run() could call it straight, no catch needed.

client/src/test/java/org/asynchttpclient/EventLoopTimeoutTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @Test
+    public void byDefaultTheTimeoutIsDeliveredFromTheTimerThread() throws Throwable {
+        String thread = threadDeliveringRequestTimeout(false);
+
+        assertTrue(thread.contains(TIMER_MARKER),
+                "expected the timer thread by default, got " + thread);
+    }
+
+    @Test
+    public void withEventLoopTimeoutsTheTimeoutIsDeliveredFromAnEventLoop() throws Throwable {
```
hyperxpro: Both tests use a fresh client's first request, so timeoutExecutor always takes the channel null branch. The channel branch, the whole reason scheduleRequestTimeout grew a third parameter and the only place affinity actually exists, is never exercised: drop the channel argument at the pooled call site and both tests stay green. The RejectedExecutionException fallback in arm() has no test either, and neither does a read timeout under the new mode, which is where isClaimed replaced isExpired.

client/src/test/java/org/asynchttpclient/EventLoopTimeoutTest.java
```diff
@@ -0,0 +1,126 @@
...
+ * Where a request timeout is delivered from, which is what
+ * {@link AsyncHttpClientConfig#isUseEventLoopTimeouts()} changes. Off, every expiry in the client runs on the
+ * timer's single thread; on, it runs on an event loop. The thread a timeout is delivered on is observable
+ * through {@link AsyncHandler#onThrowable}, so these assert the switch rather than its side effects.
+ */
+public class EventLoopTimeoutTest extends HttpTest {
+
+    private static final String IO_THREAD_POOL = "ahc-timeout-test";
+    // Netty derives the timer's thread names from this, so a timer thread is the one carrying "timer".
+    private static final String TIMER_MARKER = "timer";
```
hyperxpro: Both assertions rest on this substring. The timer factory is the pool name plus a timer suffix and the I/O factory is the pool name alone, so a pool name containing timer makes the first test pass for the wrong reason and the second fail, and setting a thread factory on the config drops the suffix entirely and breaks the default case with no bug present. Asserting on the executor rather than the thread name would be steadier. Minor: each test leaves the echo handler sleeping 5s while the server closes, so the class pays that stop timeout twice.

hyperxpro (changes requested):
Round 1

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutTimerTask.java
pavel-ptashyts: Right, and the path checks out: ScheduledFutureTask.cancel goes through AbstractScheduledEventExecutor.removeScheduled, which for an off-loop caller has to enqueue the removal. cancelArmed now catches RejectedExecutionException and reports it as not-cancelled, so cancel() and abort() keep the contract they had before this branch. Debug rather than warn: the entry dies with the loop anyway, and a caller cancelling a request on a closing client has no use for the news.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: Agreed, and this is the one that actually bites. Arming has moved out of the constructor into TimeoutsHolder#start(), which scheduleRequestTimeout calls only after setTimeoutsHolder, so the task can no longer reach a half-built holder or a future that has not been handed one.

The pooled path also had attachChannel after the arming, and those two are now swapped: an expiry that lands immediately finds a channel on the future and closes it, instead of aborting with null and leaving the socket open.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: You are right, and the javadoc was claiming something the code did not do. timeoutExecutor no longer calls next(): it returns the channel's loop when there is a channel and null otherwise, which leaves the connect path on the timer exactly as it was. NettyConnectListener.onSuccess then moves it across, right after attachChannel as you suggested, through a new TimeoutsHolder#rehomeOn. That listener already runs on the channel's own loop, so the move is a same-thread schedule, and the read timeout armed later re-arms on that loop too. The holder keeps the switch rather than every caller consulting the config, which also keeps the listener from needing the request sender it has no other use for at that point.

The invariant is now simply: the loop that owns the exchange's channel, or the timer for as long as there is no channel.

pavel-ptashyts: The same change covers this one: nothing draws from the group's chooser any more, so the counter is left to initAndRegister alone. That seemed better than picking with ThreadLocalRandom, which would still have been the wrong loop, only less predictably so.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: Agreed. arm re-reads cancelled after recording the handle and releases the task there, so either the arming thread sees the flag or cancel() sees the handle.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: Correct on both counts, and the field is gone. startReadTimeout tests requestTimeoutTask == null, which is the shape the code had before this branch, and the unchecked deref goes with it.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutTimerTask.java
pavel-ptashyts: Fair, and cheap to fix. Two typed fields now, io.netty.util.Timeout and io.netty.util.concurrent.ScheduledFuture, with an armedOn overload each. Same allocation as before, and a hierarchy change becomes a compile error rather than a cancellation that quietly stops working.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: Yes, that description was wrong. It went with the boolean: requestTimeoutArmed was its only reader, so arm is void now and there is nothing left to describe.

client/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
pavel-ptashyts: Agreed. isUseEventLoopTimeouts() is the only place that carries the reasoning; the other three say what they do and link here.

I did add one paragraph to it, about the timer carrying the timeout until a channel exists and the exchange moving it afterwards. After the change discussed on the next() thread that is part of the contract rather than an implementation detail, so it belongs in the canonical place.

client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClientConfig.java
pavel-ptashyts: Agreed, and splitting that pair was careless. The whole set has moved into the // timeouts group next to requestTimeout and readTimeout, which is where it belongs on its own merits: field, constructor parameter and assignment, getter, builder field, copy constructor, setter, the build() argument, AsyncHttpClientConfigDefaults and ahc-default.properties. The interface method moved next to getRequestTimeout for the same reason, and the import is alphabetical, before defaultUseInsecureTrustManager.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutTimerTask.java
pavel-ptashyts: You are right, that constructor makes the subclass impossible. run(Timeout) is redeclared without throws, run() calls it directly, and the catch is gone.

One thing worth flagging: revapi objects to it, java.method.exception.checkedRemoved, so this needed an ignore entry scoped to that one method. I tried it without: keeping the inherited signature and letting each subclass own run() instead trades it for abstractMethodAdded, which is the worse of the two, since that one would break a subclass rather than a caller catching a narrower checked exception around a direct call. Only leaving the dead catch in place avoids both. Say the word if you would rather have the catch than the entry and I will put it back.

client/src/test/java/org/asynchttpclient/EventLoopTimeoutTest.java
pavel-ptashyts: Agreed, and the tests were covering only the branch that has since been removed. There are four now: the timer default, a connecting exchange, a pooled one, and a read timeout under the new mode.

The pooled case takes the channel from onConnectionPooled and waits for onConnectionOffer before firing the second request, so it cannot quietly open a connection of its own and test nothing. The connecting case takes the channel from onTcpConnectSuccess and gets a one second budget: a deadline it could reach before connecting would be delivered from the timer quite correctly, there being no channel to deliver it from, and would prove nothing either way.

The RejectedExecutionException fallback in arm still has no test, and I would rather say so than leave you to find out. To reach it I need a loop that answers isShuttingDown() with false and then rejects the schedule, and the executor comes from the channel, so there is no way in through the config. If a test double for EventExecutor is acceptable to you I will add one.

client/src/test/java/org/asynchttpclient/EventLoopTimeoutTest.java
pavel-ptashyts: Agreed on both. The tests now build their own Timer and EventLoopGroup and hand them to the config, so the default case asserts assertSame against the timer's own thread and the event-loop cases assert channel.eventLoop().inEventLoop(...). No name matching left, and setting a thread factory can no longer break them.

The 5 s delay is gone as well: the endpoint blocks on a latch that @AfterEach releases before closing the server, so teardown pays nothing for it.

pavel-ptashyts:
Round 1 is addressed; thanks, the two structural ones changed the shape of this for the better.

What moved:

* Timeouts are only ever armed on the loop that owns the exchange's channel. `next()` is gone, so the group's chooser is left to `initAndRegister`, and the connect path arms on the timer and is moved onto the loop by `NettyConnectListener` once the connect succeeds.
* Arming left the `TimeoutsHolder` constructor for a `start()` the caller invokes after publishing the holder, and the pooled path attaches the channel before arming rather than after.
* `arm` re-checks `cancelled` after recording the handle; `cancelArmed` no longer throws out of `cancel()` during shutdown, and holds two typed handles instead of an `Object`.
* `requestTimeoutArmed` and the `throws` on `run(Timeout)` are gone, the rationale lives in `isUseEventLoopTimeouts()` alone, and the new option sits in the `// timeouts` group everywhere instead of splitting the failedIpCooldown pair.
* The tests assert against the timer and the loops themselves rather than thread names, and cover the pooled branch and a read timeout. The one gap left is the `RejectedExecutionException` fallback in `arm`, noted inline.

Dropping the `throws` on `run(Timeout)` needed a revapi entry, `java.method.exception.checkedRemoved`, scoped to that method; the justification is in the pom and the reasoning is on the inline thread. It is the only way to keep the dead catch out, and the alternative shapes trade it for a worse difference.

`./mvnw clean verify` is green. Not on JDK 11 as the contract asks, only 17 - no 11 on this machine.

client/src/test/java/org/asynchttpclient/EventLoopTimeoutTest.java
```diff
@@ -0,0 +1,232 @@
...
+        assertNull(recorder.pooledChannel.get(), "this request was meant to open its own connection");
+        assertDeliveredOnTheLoopOf(recorder.connectedChannel.get(), recorder);
+    }
+
+    @Test
+    public void anExchangeOnAPooledChannelExpiresOnThatChannelsLoop() throws Throwable {
+        Recorder first = new Recorder();
+        Recorder second = new Recorder();
+
+        withClient(baseConfig(true)).run(client -> withServer(server).run(server -> {
```
hyperxpro: Only test that keeps the 200 ms from baseConfig, and its first request is the cold path the other two buy headroom for. Class loading plus connect plus Jetty's first init will blow that on a loaded runner, and it fails as a timeout on a request meant to succeed. Give it the same second the sibling uses.

client/src/test/java/org/asynchttpclient/EventLoopTimeoutTest.java
```diff
@@ -0,0 +1,232 @@
...
+    private HashedWheelTimer timer;
+    private final AtomicReference<Thread> timerThread = new AtomicReference<>();
+    // Released before the server is closed, so a request left hanging on purpose never delays teardown.
+    private final CountDownLatch released = new CountDownLatch(1);
+
+    @BeforeEach
+    public void start() throws Throwable {
+        server = new HttpServer();
+        server.start();
+        eventLoopGroup = new NioEventLoopGroup(2, new DefaultThreadFactory("ahc-timeout-test"));
```
hyperxpro: Two loops means the wrong loop is the right one half the time. Revert timeoutExecutor to next() and these assertions still pass about half the runs. Four to eight.

pom.xml
```diff
@@ -501,6 +501,12 @@
...
                                                 "justification": "The public computeRspAuth(Realm) was removed by the Digest mutual-authentication fix; revapi pairs it with the new private helper of the same name and reports the removal as a visibility reduction. It computed the expected rspauth from the Realm carried on the response future, whose uri is stale after a redirect and whose cnonce is regenerated on every build() - i.e. never the values actually sent on the wire - so it could only ever produce a value that disagrees with a conformant server. Verification now derives every parameter from the request's own Authorization header. The method had no caller inside the library; leaving it would invite it to be wired back in."
                                             },
+                                            {
+                                                "code": "java.method.exception.checkedRemoved",
```
hyperxpro: I think you can skip this. Both subclasses already declare run(Timeout) without throws on main, so moving implements Runnable and run() down into them leaves TimeoutTimerTask's surface byte identical: no entry and no dead catch. Nothing abstract is added to the base, so abstractMethodAdded doesn't apply. I have not run revapi:check on it, so try it; if it doesn't hold, the entry as written is fine.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -15,37 +15,63 @@
...
     private final long requestTimeoutMillisTime;
     private final long readTimeoutValue;
-    private volatile Timeout readTimeout;
+    private final boolean useEventLoopTimeouts;
+    private final @Nullable RequestTimeoutTimerTask requestTimeoutTask;
+    private volatile @Nullable ReadTimeoutTimerTask readTimeoutTask;
     private final NettyResponseFuture<?> nettyResponseFuture;
     private volatile InetSocketAddress remoteAddress;
 
     public TimeoutsHolder(Timer nettyTimer, NettyResponseFuture<?> nettyResponseFuture, NettyRequestSender requestSender,
```
hyperxpro, line 58: Public constructor in an exported package, and it went from construct-and-arm to construct-only. Compatible, so revapi says nothing, but an outside caller who installs a holder now gets an exchange with no request timeout. Calling start() from setTimeoutsHolder would make that impossible.

client/src/test/java/org/asynchttpclient/EventLoopTimeoutTest.java
```diff
@@ -0,0 +1,232 @@
...
+        Recorder second = new Recorder();
+
+        withClient(baseConfig(true)).run(client -> withServer(server).run(server -> {
+            server.enqueueOk();
+            client.prepareGet(server.getHttpUrl() + "/foo/bar").execute(first);
+            first.awaitCompletion();
+            // Waited for, not assumed: the connection is offered to the pool around the same time as the future
+            // completes, and a second request that overtook the offer would open its own connection and test
+            // the wrong branch.
+            assertTrue(first.offered.await(10, TimeUnit.SECONDS), "the first connection was never pooled");
```
hyperxpro: This can't fail. finishUpdate offers to the pool before future.done(), and done() is what releases settled, so awaitCompletion above already covers it. The comment describes a race that can't happen.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutTimerTask.java
```diff
@@ -15,31 +15,105 @@
...
+
+    /**
+     * Cancels the scheduled entry this task was armed on, if any. Never interrupts: on the event-loop path the
+     * task may be running on the very thread this is called from, and nothing in it answers interruption.
+     *
+     * @return whether an entry was taken back out of its scheduler before it could run
+     */
+    boolean cancelArmed() {
+        Timeout timer = timerHandle;
+        if (timer != null) {
```
hyperxpro, line 79: Latent, nothing reaches it today. A stale timer handle would mask a live loop handle, and that entry sits in the queue holding the future until its deadline. Each armedOn nulling the other field closes it.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -98,24 +157,64 @@ void startReadTimeout(ReadTimeoutTimerTask task) {
...
-        if (requestSender == null || nettyTimer == null || requestSender.isClosed()) {
-            return null;
+        if (requestSender == null || requestSender.isClosed()) {
+            return;
+        }
+        EventExecutor executor = eventExecutor;
+        if (executor != null && !executor.isShuttingDown()) {
+            try {
+                task.armedOn(executor.schedule(task, delay, TimeUnit.MILLISECONDS));
+                cancelIfRaced(task);
```
hyperxpro: cancelIfRaced is inside the try for schedule(). Nothing throws here today, but if cancelArmed ever stops swallowing the rejection, it lands in the schedule's handler and re-arms on the timer for a finished exchange. Move it below the catch.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -98,24 +157,64 @@ void startReadTimeout(ReadTimeoutTimerTask task) {
...
+        if (task != null) {
+            task.cancelArmed();
+            task.clean();
+        }
+    }
+
+    private long remainingRequestTimeout() {
+        // A deadline already behind us is armed at zero rather than negative, so the task still runs and still
+        // cancels its read-timeout sibling, which is bookkeeping only it does.
+        return Math.max(requestTimeoutMillisTime - unpreciseMillisTime(), 0L);
```
hyperxpro, line 183: Not for this PR. Two wall clock reads where main used the configured duration, so an NTP step moves the deadline. Widens an existing exposure rather than adding one.

hyperxpro (changes requested):

client/src/test/java/org/asynchttpclient/EventLoopTimeoutTest.java
pavel-ptashyts: Agreed - and it is the first request there that would have failed, the one meant to succeed. It takes the same second the connecting case does now, hoisted into a constant the two share so the next reader can see why either has it.

client/src/test/java/org/asynchttpclient/EventLoopTimeoutTest.java
pavel-ptashyts: This is the better of the two test notes, and it was right: with two loops the assertion agreed with the bug half the time. Eight now.

Checked rather than assumed - reverting timeoutExecutor to next() fails the pooled case on both runs I tried, where before it would have been a coin toss.

pom.xml
pavel-ptashyts: It holds. `revapi:check` passes with the entry removed: moving `implements Runnable` and `run()` down into the two subclasses leaves `TimeoutTimerTask` untouched, and inside a subclass `run()` calls its own override of `run(Timeout)`, which already declared no throws - so there is nothing to catch and nothing to narrow. Entry is out of the pom.

One knock-on: only the concrete classes are both a `TimerTask` and a `Runnable`, so `arm` takes `<T extends TimeoutTimerTask & Runnable>`. Slightly exotic for a private method, but it keeps the two schedulers honest at the call site rather than casting.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: Agreed, and your fix is better than what I had. `start()` is called from `setTimeoutsHolder`, so installing a holder is what arms it and there is no order for an outside caller to get wrong. The sender no longer arms by hand.

The reason the arming left the constructor still holds under this: the holder is reachable from the future before its task can run, and on the pooled path the channel is attached before the install.

client/src/test/java/org/asynchttpclient/EventLoopTimeoutTest.java
pavel-ptashyts: You are right, I checked: `finishUpdate` offers to the pool before `future.done()`, and `done()` is what releases the latch `awaitCompletion` is already waiting on. It was a wait for something that had by then happened, described by a comment about a race that cannot occur. Both gone; the assertion on `onConnectionPooled` still says the second request actually took the connection.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutTimerTask.java
pavel-ptashyts: Agreed. Each `armedOn` clears the other field now, so the order the two are ever called in stops mattering.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: Agreed. Only `schedule()` is inside the try now; the handle is recorded and unwound after the catch, so nothing but a rejected schedule can reach the timer fallback.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: Noted, and taken as far as it goes without turning into another change: `start()` arms with the configured duration, the way main did, since it runs within microseconds of the constructor and the second clock read bought nothing. The remaining-time arithmetic is left only where subtracting elapsed time is the whole point, on the re-homing path. So this is no wider than main except there.

`unpreciseMillisTime` being `System.currentTimeMillis()` is a separate matter and I agree it is not this PR's.

pavel-ptashyts:
Round 2 addressed. The revapi hypothesis was right, and so was the two-loop one - thanks for both, they were not obvious from where I was sitting.

* `implements Runnable` moved down into the two subclasses, so `TimeoutTimerTask` is byte identical and the revapi entry is out of the pom. No dead catch either: inside a subclass `run()` calls its own override, which already declared no throws. `arm` takes `<T extends TimeoutTimerTask & Runnable>` as the knock-on.
* `start()` is called from `NettyResponseFuture#setTimeoutsHolder`, so an outside caller cannot install a holder and get an exchange with no request timeout.
* `armedOn` clears the other handle; `cancelIfRaced` moved out of the try that guards `schedule()`; `start()` arms with the configured duration rather than reading the clock a second time.
* Tests: eight loops instead of two, and reverting `timeoutExecutor` to `next()` now fails the pooled case on both runs tried. The pooled case gets the cold-path budget its first request needs, and the wait for `onConnectionOffer` is gone - `finishUpdate` offers before `future.done()`, so it was waiting for something already done.

`./mvnw clean verify` green: 1468 tests, 0 failures, API checks clean with no revapi entries added. JDK 17 locally again, no 11 on this machine; the JDK 11 legs of CI are the real gate and were green on the previous push.

hyperxpro (approved):

hyperxpro:
Thanks a lot!

PR #2314, opened 2026-08-17
Bound an exchange by one request timeout

pavel-ptashyts:
## Problem

`TimeoutsHolder` anchors the request deadline on its own construction:

```java
requestTimeoutMillisTime = unpreciseMillisTime() + requestTimeoutInMs;
```

A redirect, a retry and an auth replay all continue the same exchange on the same
`NettyResponseFuture`, but each builds a new holder for it. Every hop therefore starts the
budget again, so with `maxRedirects=5` a chain can legitimately run for six times the
configured `requestTimeout`. Nothing carries an absolute deadline across hops:
`NettyResponseFuture#getStart()` exists but is only read for a diagnostic `age` in a log line.

The `getRequestTimeout()` javadoc says it is "the maximum time an AsyncHttpClient waits until
the response is completed", which is not what happens.

## Change

`AsyncHttpClientConfig#isUseAbsoluteRequestDeadline()`, **off by default**, anchors the
deadline on when the exchange was submitted instead, so a later hop gets whatever is left of
the budget rather than a fresh one.

Off by default because turning it on shortens exchanges that rely on the per-attempt
behaviour, which is a behaviour change even if the current one contradicts the docs. The
`getRequestTimeout()` javadoc now describes what actually happens and points at the flag, so
it stops being wrong either way.

Settable **per request** as well as per client, following the existing `followRedirect`
pattern: a nullable `Boolean` on `Request` that overrides the config value.

```java
client.prepareGet(url).setUseAbsoluteRequestDeadline(true).execute(handler);
```

## Where the flag lives, and why not on the request

It is resolved once, in `newNettyResponseFuture`, and kept on the `NettyResponseFuture`.

The first attempt kept it only on `Request` and the two override tests failed in opposite
directions. `Redirect30xInterceptor` rebuilds the request for the next hop from a hand-picked
set of fields, so the override was silently dropped mid-exchange and the config value took
over - precisely in the case the setting exists for. Anything carried only on the request has
that problem, and every future site that rebuilds a request would have to remember it.

Keeping it on the exchange also says the right thing: the deadline describes the exchange, and
a redirect target cannot change it, because the budget belongs to the caller.

## Review round 1

Two of the nine were bugs rather than polish, and both are worth reading before the rest:

**An attempt the deadline has no time for is no longer sent.** Clamping a spent deadline to zero
only delayed the abort: the attempt still took a connection permit, took a connection and wrote
the request, and the timeout arrived a tick later - so a 307 put its body on the redirect target
while the caller was handed a `TimeoutException` that reads as though nothing had been sent.
`scheduleRequestTimeout` now fails the exchange before the write. Every attempt passes through
it, first or otherwise, and it is the last point before the request goes out; it returns whether
the attempt may go ahead and its four call sites stop when it may not. Disabling that guard
while keeping the new test shows the old behaviour was not even a late abort: on a wheel too
coarse to expire the exchange in time, it ran both hops and **completed successfully**, the
deadline exceeded and nothing reported.

The budget is a static on `TimeoutsHolder`, asked of the future rather than of a holder, because
a first attempt has no holder to ask - and under `ROUND_ROBIN` the up-front resolve asks for no
timeout at all, so a slow resolver could spend the whole budget before one existed.
`Long.MAX_VALUE` stands for an exchange that is not bounded as a whole, so a per-attempt timeout
needs no special case at any call site.

**The anchor is monotonic.** It was `getStart()`, which is `currentTimeMillis`. A per-attempt
timeout can only be distorted by a clock step for the length of one hop; an anchor spanning a
whole exchange carries the step to every hop after it, so a correction backwards hands a later
hop a budget it never had and one forwards aborts it on a healthy connection. The future now
records `System.nanoTime()` at submission and the budget is netted off that. Wall clock is left
only where main already had it - `requestTimeoutMillisTime` is still recomputed per hop, for the
read-timeout comparison.

Also from the review: two fields were being dropped by hand-copied lists, which is the same
fault this change exists to fix. `Redirect30xInterceptor` rebuilds the next request from a
hand-picked set and carried neither the read timeout - reverting a per-request value to the
config default on every hop after the first - nor the deadline flag, which left the `Request`
disagreeing with what the exchange was being held to. `RequestBuilderBase`'s
signature-calculator copy block dropped the read timeout the same way. And the comment
justifying the clamp claimed the task still cancels its read-timeout sibling; there is no
sibling at that point, the read timeout being armed after the write.

## Behaviour change outside the flag

Carrying the read timeout across a redirect applies whether or not anyone turns the deadline on.
A caller with a short per-request read timeout and `followRedirect` was getting the config
default - 60 s unless configured otherwise - on every hop after the first, and now gets the
value they set. That is a fix, but it is a visible one and belongs in the release notes. There is
no release notes file in the repo, so it is called out here.

## API compatibility

`revapi` passes with no entries. `DefaultRequest` keeps its existing public constructor, and
the widened one taking the flag is **package private**: public, it would be pinned by revapi at
twenty-seven arguments, the next per-request option would make it twenty-eight, and the two
parameter lists would have to be kept in step by hand with a tail that is all reference types
for the compiler to confuse. `RequestBuilderBase#build` is the only caller.

`Request#getUseAbsoluteRequestDeadline()` is a `default` method returning null, so existing
implementations are unaffected.

`AsyncHttpClientConfig#isUseAbsoluteRequestDeadline()` returns a literal rather than reading
`org.asynchttpclient.useAbsoluteRequestDeadline`, as every other option on that interface does.
The javadoc now says so, so that an implementation setting the property and getting nothing is
documented rather than surprising.

## Tests

`AbsoluteRequestDeadlineTest` runs two hops of 400 ms against a 600 ms budget, so each hop fits
on its own and the pair does not. Six cases:

* default: both hops run and the exchange ends on the second (per-attempt behaviour preserved)
* config on: the chain times out
* config off + request override on: times out
* config on + request override off: ends on the second hop
* config on, single hop: completes, so the first hop is not handed a shortened budget
* config on, first hop answering after the budget is gone: the second hop is never sent

The cases that pass assert the final 200 and which hop it came from, not merely that nothing was
thrown - a dropped `Location` header or `followRedirect` turned off would satisfy that having run
a single hop. The last case pins its client to a wheel too coarse to expire anything, which is
the only way to land reliably in the window where the deadline has passed and the timeout has not
yet run; it was checked by mutation rather than by assumption.

`TimeoutsHolderTest` covers the anchor and the clamp without a wall clock. Given no timer and no
request sender, a holder computes its deadline and arms nothing, so what a second holder makes of
the same exchange is the whole of the difference between the two modes: the anchor holding across
hops, a per-attempt timeout starting a fresh budget, a spent deadline reporting itself as passed,
and a per-attempt exchange never reporting that however long it has run.

`AsyncHttpClientDefaultsTest` asserts this default and the property that sets it, along with the
event-loop one merged alongside, which had the same gap.

Timing-based, so `@RepeatedIfExceptionsTest`, matching the neighbouring timeout tests.

## Verification

`mvnw clean verify` - BUILD SUCCESS, 1485 tests, 0 failures, 0 errors, 26 skipped. Error
Prone, NullAway and Revapi all clean.

installed on this machine, so it was run on **JDK 17** (also in the CI matrix). The JDK 11 leg
of CI on this PR is the real gate.

## Relationship to #2313

Resolved: #2313 merged first and `main` is merged in here. A merge rather than a rebase, since
the branch is published and `AGENTS.md` rules out force-pushing a shared one.

The conflict was the four lines of the `TimeoutsHolder` constructor both changes touch, plus the
two config methods landing in the same place. `start()` now arms with the remaining budget when
the deadline is absolute and with the configured duration when it is per attempt, which is what
#2313 left it doing.

## Noticed while running the suite

Two timing tests are flaky under load and unrelated to this change, mentioned only so a red
run is not mistaken for this PR:

* `SemaphoreTest.checkAcquireTime` (three methods) allow 400 ms for a 100 ms timeout and use
  `@RepeatedTest(10)`, which does not retry, unlike the `checkRelease` tests beside them. This
  failed one cell of thirteen on #2313's CI (macOS, JDK 21) at 420 ms.
* `NettyRequestThrottleTimeoutTest.testRequestTimeout` takes ~31 s against a 30 s latch even on
  an idle machine, and releases its throttle permit only from `onThrowable`, so a single request
  completing instead of timing out deadlocks the remaining threads.

Happy to send a separate PR for both.


client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -59,8 +59,16 @@ public TimeoutsHolder(Timer nettyTimer, NettyResponseFuture<?> nettyResponseFutu
...
+            // A redirect, a retry or an auth replay builds a new holder for the same future. Anchoring the
+            // deadline here hands each of those hops a fresh budget, so a chain of n hops runs for n times the
+            // configured timeout; anchoring it on the future bounds the exchange as a whole instead. Which one
+            // applies is the caller's choice, per request or per client.
+            requestTimeoutMillisTime = (nettyResponseFuture.isUseAbsoluteRequestDeadline()
+                    ? nettyResponseFuture.getStart() : unpreciseMillisTime()) + requestTimeoutInMs;
+            // A deadline already behind us is scheduled at zero rather than negative, so the task still runs and
+            // still cancels its read-timeout sibling, which is bookkeeping only it does.
+            requestTimeout = newTimeout(new RequestTimeoutTimerTask(nettyResponseFuture, requestSender, this, requestTimeoutInMs),
+                    Math.max(requestTimeoutMillisTime - unpreciseMillisTime(), 0L));
```
hyperxpro: Clamping to zero does not stop the hop, it only delays the abort. The wheel timer fires on tick boundaries, so with a deadline already behind us we still take a permit, still take a connection and still write the request, then fail it a tick later. On a 307 that means the POST body reached the redirect target while the caller is handed a TimeoutException that reads like nothing was sent. I think this wants a remaining budget check in sendNextRequest that fails the exchange before the write.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -59,8 +59,16 @@ public TimeoutsHolder(Timer nettyTimer, NettyResponseFuture<?> nettyResponseFutu
         }
 
         if (requestTimeoutInMs > -1) {
-            requestTimeoutMillisTime = unpreciseMillisTime() + requestTimeoutInMs;
-            requestTimeout = newTimeout(new RequestTimeoutTimerTask(nettyResponseFuture, requestSender, this, requestTimeoutInMs), requestTimeoutInMs);
+            // A redirect, a retry or an auth replay builds a new holder for the same future. Anchoring the
+            // deadline here hands each of those hops a fresh budget, so a chain of n hops runs for n times the
+            // configured timeout; anchoring it on the future bounds the exchange as a whole instead. Which one
+            // applies is the caller's choice, per request or per client.
+            requestTimeoutMillisTime = (nettyResponseFuture.isUseAbsoluteRequestDeadline()
```
hyperxpro: getStart is currentTimeMillis, but the timer that actually fires runs on nanoTime. Until now a clock step could only distort the hop in flight. With an absolute anchor, a backward NTP correction of a few seconds effectively removes a 600ms deadline for the rest of the chain, and a forward one aborts every later hop instantly on a healthy connection. An absolute anchor really wants a monotonic source.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -59,8 +59,16 @@ public TimeoutsHolder(Timer nettyTimer, NettyResponseFuture<?> nettyResponseFutu
...
+                    ? nettyResponseFuture.getStart() : unpreciseMillisTime()) + requestTimeoutInMs;
+            // A deadline already behind us is scheduled at zero rather than negative, so the task still runs and
+            // still cancels its read-timeout sibling, which is bookkeeping only it does.
```
hyperxpro: This justification does not hold. readTimeout is still null at this point, it is only set once scheduleReadTimeout runs after the write, and requestTimeout is not assigned yet either because the constructor registers with the timer first. In the case you are describing the task cancels nothing.

client/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
```diff
@@ -109,11 +109,35 @@ public interface AsyncHttpClientConfig {
...
+     * this anchors it on when the exchange was submitted instead, so a later hop gets whatever is left and the
+     * caller's total wait is bounded by the one value.
+     * <p>
+     * Off by default because turning it on shortens exchanges that rely on the per-attempt behaviour. A caller
+     * working to an end-to-end budget wants it on; {@link Request#getUseAbsoluteRequestDeadline()} sets it for a
+     * single request.
+     *
+     * @return {@code true} to treat the request timeout as a deadline for the whole exchange
+     */
+    default boolean isUseAbsoluteRequestDeadline() {
```
hyperxpro, line 169: Hardcoding false here means the property only ever works through the builder. Someone on a custom config or a wrapper sets org.asynchttpclient.useAbsoluteRequestDeadline, gets nothing, and nothing is logged, while a sibling service on the builder honours it. Either delegate to defaultUseAbsoluteRequestDeadline or say in the javadoc that the property is builder only.

client/src/main/java/org/asynchttpclient/DefaultRequest.java
```diff
@@ -99,6 +100,45 @@ public DefaultRequest(String method,
...
+                channelPoolPartitioning, nameResolver, null);
+    }
+
+    /**
+     * @param useAbsoluteRequestDeadline whether {@code requestTimeout} bounds the whole exchange rather than
+     *                                   each attempt within it, or null to defer to the client config. Trailing
+     *                                   rather than beside {@code followRedirect} so the original signature
+     *                                   stays intact for callers that build a request without the builder.
+     */
+    public DefaultRequest(String method,
```
hyperxpro: This is the part we cannot take back later. A 27 arg public constructor gets pinned by revapi, the next per request option makes it 28, and the two parameter lists have to stay in lockstep with no compiler help if they drift, since the tail is all reference types. Only RequestBuilderBase.build calls it. Package private, or a small options holder, costs nothing today.

client/src/main/java/org/asynchttpclient/DefaultRequest.java
```diff
@@ -99,6 +100,45 @@ public DefaultRequest(String method,
...
                           ChannelPoolPartitioning channelPoolPartitioning,
                           NameResolver<InetAddress> nameResolver) {
+        this(method, uri, address, localAddress, headers, cookies, byteData, compositeByteData, stringData,
+                byteBufferData, byteBufData, streamData, bodyGenerator, formParams, bodyParts, virtualHost,
+                proxyServer, realm, file, followRedirect, requestTimeout, readTimeout, rangeOffset, charset,
+                channelPoolPartitioning, nameResolver, null);
+    }
+
+    /**
+     * @param useAbsoluteRequestDeadline whether {@code requestTimeout} bounds the whole exchange rather than
```
hyperxpro: AGENTS.md asks us not to explain a change to reviewers in the code. This one is revapi rationale rather than API contract, and it ends up in the published javadoc of a public constructor. Same for the block in NettyResponseFuture.

client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
```diff
@@ -685,6 +698,7 @@ private RequestBuilderBase<?> executeSignatureCalculator() {
         rb.realm = realm;
         rb.file = file;
         rb.followRedirect = followRedirect;
+        rb.useAbsoluteRequestDeadline = useAbsoluteRequestDeadline;
         rb.requestTimeout = requestTimeout;
```
hyperxpro, line 702: Since you are already in this copy block, readTimeout is missing from it. Set a signature calculator and a per request read timeout silently falls back to the config 60s. Exactly the hand copied field list problem this PR is about.

client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
```diff
@@ -726,6 +729,21 @@ public void acquirePartitionLockLazily(boolean nonBlocking) throws IOException {
...
     }
 
+    /**
+     * Whether this exchange's request timeout is a deadline for the exchange as a whole. Resolved once, from the
+     * request the caller submitted and the client config, and then kept here rather than re-read per hop: a
+     * redirect rebuilds the request from a hand-picked set of fields, so anything carried only on the request
+     * would silently revert to the config value partway through the exchange, which is exactly the case this
+     * setting exists for.
+     */
+    public boolean isUseAbsoluteRequestDeadline() {
```
hyperxpro, line 756: Holding it on the future is the right call, but the Request now disagrees with the behaviour. After a redirect, getTargetRequest().getUseAbsoluteRequestDeadline() is null, so a filter, a signature calculator or a handler reading the request thinks we are on per attempt timeouts while the future is enforcing a deadline. Redirect30xInterceptor can carry it across for free. While in there, it also drops readTimeout, which resets a 500ms read timeout to the config default on every hop after the first.

client/src/test/java/org/asynchttpclient/AbsoluteRequestDeadlineTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @RepeatedIfExceptionsTest(repeats = 5)
+    public void byDefaultEachHopGetsItsOwnBudget() throws Throwable {
+        // Two hops of 400 ms against a 600 ms budget. Each hop on its own fits, the pair does not, so with a
+        // per-attempt timeout the exchange completes.
+        enqueueTwoDelayedHops();
+
+        Throwable cause = runAndAwait(baseConfig(), null);
+
+        assertNull(cause, "per-attempt timeouts should let both hops run, got " + cause);
```
hyperxpro: This one and the two other assertNull tests only check that nothing threw. Drop the Location header or break followRedirect and they all still pass, having exercised a single hop. Assert the final 200 or the second URI so they can actually fail.

client/src/main/java/org/asynchttpclient/config/AsyncHttpClientConfigDefaults.java
```diff
@@ -183,6 +184,10 @@ public static boolean defaultFailedIpCooldownEnabled() {
         return AsyncHttpClientConfigHelper.getAsyncHttpClientConfig().getBoolean(ASYNC_CLIENT_CONFIG_ROOT + FAILED_IP_COOLDOWN_ENABLED_CONFIG);
     }
 
+    public static boolean defaultUseAbsoluteRequestDeadline() {
```
hyperxpro, line 192: Nothing asserts this default. AsyncHttpClientDefaultsTest covers the other flags, and if the key is ever dropped or misspelled getBoolean returns false quietly, which is also the intended default, so the regression stays invisible until someone flips it. A direct TimeoutsHolder test for the anchor and the clamp would help too, the five new tests are all wall clock.

hyperxpro (changes requested):
Round 1

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: Agreed, and it turns out to be worse than late. sendNextRequest, which every redirect, auth replay and retry funnels through, now fails the exchange before the write when the deadline has passed.

I kept the test and disabled the guard to see what the old behaviour actually was: on a wheel too coarse to expire the exchange in time, it did not abort late at all - it ran both hops and completed successfully, the deadline exceeded and nothing reported. So the clamp was hiding two failures, not one.

The abort passes future.channel(), which is the same channel a request timeout firing at that instant would have closed, so the connection's fate is no different from the timer path.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: Agreed, and this is the one that made the anchor unsafe rather than merely imprecise. The future now records System.nanoTime() at submission and the budget is netted off that, so a correction landing mid-chain cannot move the deadline in either direction.

Wall clock is left only where main already had it: requestTimeoutMillisTime is still computed per hop, for the read-timeout comparison, so that exposure is unchanged rather than widened.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: You are right, there is no sibling at that point - the read timeout is armed after the write. The comment now gives the actual reason: a scheduler has no use for a negative delay, and the task has to run either way, running being what fails the exchange.

client/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
pavel-ptashyts: Took the second option, the javadoc. Delegating would make this one option behave unlike every other on the interface: they all return a literal, including isUseEventLoopTimeouts which merged alongside this. The javadoc now says the property is read by the builder and that an implementation which does not override gets false whatever the property says.

If you would rather the interface read the properties, I agree it would be better, but it is a change worth making across the board rather than for one option.

client/src/main/java/org/asynchttpclient/DefaultRequest.java
pavel-ptashyts: Agreed. It is package private now, RequestBuilderBase#build being the only caller, and the public twenty-six argument constructor is untouched, so revapi has nothing to pin. The javadoc records why, so that the next option added does not quietly make it public again.

client/src/main/java/org/asynchttpclient/RequestBuilderBase.java
pavel-ptashyts: Fixed, and thank you for spotting it - it is the same fault this change exists to fix, one file over. `rb.readTimeout = readTimeout;` added to the block.

client/src/main/java/org/asynchttpclient/netty/NettyResponseFuture.java
pavel-ptashyts: Agreed on both. Redirect30xInterceptor now carries the read timeout and the deadline flag across, so the Request agrees with what the exchange is being held to.

The flag only when it was actually set: the setter takes a primitive and null means defer to the client config. Worth saying that the existing tests caught my first attempt at this on the spot - passing the nullable Boolean straight into the primitive setter NPEs on every redirect that has no override, which is most of them.

client/src/test/java/org/asynchttpclient/AbsoluteRequestDeadlineTest.java
pavel-ptashyts: Agreed, they were asserting that nothing threw and calling it a pass. They now assert the final 200 and which hop it came from, through a small Outcome type, so a dropped Location header or followRedirect turned off fails them.

client/src/main/java/org/asynchttpclient/config/AsyncHttpClientConfigDefaults.java
pavel-ptashyts: Both added to AsyncHttpClientDefaultsTest: this option, and the event-loop one that merged alongside it, which had the same gap. Say the word if you would rather that second one went in separately.

There is a TimeoutsHolderTest now as well. With no timer and no request sender the holder computes its deadline and arms nothing, which makes the anchor and the clamp testable directly rather than through a request: what a second holder makes of the same exchange is the whole of the difference between the two modes. Four cases - the anchor holding across hops, a per-attempt timeout starting a fresh budget, a spent deadline reporting itself as passed, and a per-attempt exchange never reporting that however long it has run.

pavel-ptashyts:
Round 1 addressed, and #2313 is merged in - as a merge rather than a rebase, since the branch is pushed and `AGENTS.md` rules out force-pushing a shared one. The conflict was the four lines of the `TimeoutsHolder` constructor both changes touch.

The two that were bugs rather than polish:

* **A hop with no budget left is no longer sent.** `sendNextRequest` fails the exchange before the write. Disabling that guard and keeping the new test shows the old behaviour was not a late abort at all: on a coarse wheel it ran both hops and completed successfully, the deadline exceeded with nothing reported.
* **The anchor is monotonic.** The future records `System.nanoTime()` at submission and the budget is netted off that, so a clock step mid-chain cannot move a deadline that spans the whole exchange. Wall clock is left only where main already had it, per hop.

The rest: `DefaultRequest`'s widened constructor is package private, so revapi has nothing to pin and the parameter lists cannot drift into API; `Redirect30xInterceptor` carries the read timeout and the flag across a hop, and `RequestBuilderBase`'s copy block carries the read timeout, both of which were dropping fields by hand-copied list - the same fault this PR exists to fix; the clamp comment says what is actually true; and the interface javadoc admits the property is read by the builder rather than here.

Tests: the three cases that only checked "nothing threw" now assert the final status and which hop it came from; `TimeoutsHolderTest` covers the anchor and the clamp without a wall clock; `AsyncHttpClientDefaultsTest` covers this default and the event-loop one alongside it. The new fail-fast case was checked by mutation rather than by assumption.

`./mvnw clean verify` green. JDK 17 locally, no 11 on this machine, so the JDK 11 legs of CI remain the real gate.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -1210,9 +1216,25 @@ public boolean applyIoExceptionFiltersAndReplayRequest(NettyResponseFuture<?> fu
     }
 
     public <T> void sendNextRequest(final Request request, final NettyResponseFuture<T> future) {
+        TimeoutsHolder timeoutsHolder = future.getTimeoutsHolder();
+        if (timeoutsHolder != null && timeoutsHolder.isDeadlinePassed()) {
```
hyperxpro: This only gates the hops. The first attempt goes through sendRequest and never reaches here. With ROUND_ROBIN the up front resolve passes scheduleTimeout=false, so a slow resolver can eat the whole budget before any holder exists, and a pooled hit then arms at zero and writes anyway. Should we check the budget before scheduleRequestTimeout instead, off the future rather than the holder?

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -108,12 +116,37 @@ public TimeoutsHolder(Timer nettyTimer, @Nullable EventExecutor eventExecutor, N
...
+     * Whether the exchange has run out of time to start another hop with. Always {@code false} when the timeout
+     * is per attempt, where a hop is given the configured timeout of its own by definition.
+     *
+     * @see org.asynchttpclient.AsyncHttpClientConfig#isUseAbsoluteRequestDeadline()
+     */
+    public boolean isDeadlinePassed() {
+        return absoluteDeadline && requestTimeoutValue > -1 && remainingBudget() <= 0L;
+    }
+
+    /**
```
hyperxpro, line 152: I think this javadoc belongs on remainingBudget. requestTimeoutMillisTime returns a wall clock instant, not what is left, and the line comment right below it says exactly that.

client/src/test/java/org/asynchttpclient/netty/timeout/TimeoutsHolderTest.java
```diff
@@ -0,0 +1,114 @@
...
+        assertTrue(holder(future, Duration.ofMillis(1)).isDeadlinePassed(),
+                "a spent deadline should report itself as passed");
+    }
+
+    @RepeatedIfExceptionsTest(repeats = 5)
+    public void aPerAttemptExchangeNeverRunsOutOfBudgetBetweenHops() throws Exception {
+        NettyResponseFuture<?> future = exchange(false);
+        Thread.sleep(ELAPSED_MS);
+
+        assertFalse(holder(future, Duration.ofMillis(1)).isDeadlinePassed(),
```
hyperxpro: absoluteDeadline is false here, so isDeadlinePassed returns on the first conjunct and remainingBudget never runs. This passes whatever the arithmetic does. Can we assert the per attempt case some other way, maybe on the deadline the second holder computes?

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -87,8 +88,15 @@ public TimeoutsHolder(Timer nettyTimer, @Nullable EventExecutor eventExecutor, N
...
         requestTimeoutValue = requestTimeoutInMs;
+        absoluteDeadline = nettyResponseFuture.isUseAbsoluteRequestDeadline();
         if (requestTimeoutInMs > -1) {
-            requestTimeoutMillisTime = unpreciseMillisTime() + requestTimeoutInMs;
+            // A redirect, a retry or an auth replay builds a new holder for the same future. Giving each of
+            // those hops the configured timeout lets a chain of n hops run for n times it; netting off what the
+            // exchange has already spent bounds it as a whole instead. Which one applies is the caller's
+            // choice, per request or per client. May be negative, and is deliberately left so: a deadline
+            // already behind us has to read as behind us, so that startReadTimeout does not arm a sibling and
+            // isDeadlinePassed can say the exchange is over.
```
hyperxpro: isDeadlinePassed does not read this field, it calls remainingBudget directly. Only startReadTimeout depends on it staying negative, so maybe just say that.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -139,7 +139,19 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
                         .setNameResolver(request.getNameResolver())
                         .setProxyServer(request.getProxyServer())
                         .setRealm(stripAuth ? null : request.getRealm())
-                        .setRequestTimeout(request.getRequestTimeout());
+                        .setRequestTimeout(request.getRequestTimeout())
+                        // Dropped here until now, so a per-request read timeout reverted to the config default
+                        // on every hop after the first.
+                        .setReadTimeout(request.getReadTimeout());
```
hyperxpro, line 143: Worth noting this one changes behaviour outside the flag too. Anyone with a short per request read timeout and followRedirect was getting the config default on the later hops and now gets their own value. Correct, but it should probably get a line in the release notes.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -139,7 +139,19 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
                         .setNameResolver(request.getNameResolver())
                         .setProxyServer(request.getProxyServer())
                         .setRealm(stripAuth ? null : request.getRealm())
-                        .setRequestTimeout(request.getRequestTimeout());
+                        .setRequestTimeout(request.getRequestTimeout())
+                        // Dropped here until now, so a per-request read timeout reverted to the config default
```
hyperxpro: These read as a description of the change rather than of the code. AGENTS.md asks us not to, and "until now" will not mean much to whoever reads this next year. The revapi one in DefaultRequest is the same.

client/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
```diff
@@ -142,6 +146,30 @@ default boolean isUseEventLoopTimeouts() {
...
+     * working to an end-to-end budget wants it on; {@link Request#getUseAbsoluteRequestDeadline()} sets it for a
+     * single request.
+     * <p>
+     * As with every option on this interface, the {@code org.asynchttpclient.useAbsoluteRequestDeadline}
+     * property is read by {@link DefaultAsyncHttpClientConfig.Builder}, not here: an implementation of this
+     * interface that does not override this method gets {@code false} whatever the property says.
+     *
+     * @return {@code true} to treat the request timeout as a deadline for the whole exchange
+     */
+    default boolean isUseAbsoluteRequestDeadline() {
```
hyperxpro, line 169: Ignore my earlier comment on this one, you were right. Every default on this interface hardcodes its value, so delegating here would have made this the odd one out. No need to change the others either, leave it as it is.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: You are right, and the holder was the wrong thing to ask in the first place - the first attempt has none. Both moved: the check is now inside scheduleRequestTimeout, which every attempt passes through and which is the last point before the write, and the budget is a static on TimeoutsHolder taken off the future.

It returns whether the attempt may go ahead and the four call sites stop when it may not. Long.MAX_VALUE stands for an exchange that is not bounded as a whole, so a per-attempt timeout needs no special case at any of them. Resolving the configured timeout moved into that static too, so the constructor and the budget no longer resolve it separately.

The first attempt and a hop are now the same line of code, so the existing test covers both; I did not build a slow resolver plus ROUND_ROBIN plus a pooled hit to reach the same line a second way, but say so if you would rather that scenario were pinned down explicitly.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: It does, and it was there until an earlier edit of mine inserted the accessor between the javadoc and its method. Put back where it belongs, and the accessor keeps its own line comment.

client/src/test/java/org/asynchttpclient/netty/timeout/TimeoutsHolderTest.java
pavel-ptashyts: You are right, that one asserted nothing - the first conjunct answered it and the arithmetic never ran. It asserts on the deadline the holder computes instead, as you suggest: that a hop is handed the configured timeout of its own however long the exchange has already run. The spent-deadline case now goes through remainingBudget directly, so both of them exercise the arithmetic.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
pavel-ptashyts: Right, it reads the budget directly. The comment now says what actually depends on the field staying negative, which is startReadTimeout not arming a sibling for an exchange that is over.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
pavel-ptashyts: Agreed, and it is the more consequential half of that fix - it applies whether or not anyone turns the flag on. There is no release notes file in the repo, so I have called it out in the PR description under its own heading rather than burying it in the change list; happy to put it wherever you keep them.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
pavel-ptashyts: Fair, and it is the rule I should have caught myself. All three rewritten to describe the code: the two in Redirect30xInterceptor, and the revapi one in DefaultRequest, which now says what the constructor is for rather than what happened to it.

client/src/main/java/org/asynchttpclient/AsyncHttpClientConfig.java
pavel-ptashyts: Thanks for coming back to it. Left as it is.

I did keep the javadoc paragraph that came out of the original comment, since it says where the property is actually read and that is worth knowing at the point someone looks for it. Happy to drop it if you would rather one option did not carry a note about a convention all of them share.

pavel-ptashyts:
Round 2 addressed.

The one that mattered: the deadline check was only gating continuations, and asking a `TimeoutsHolder` that a first attempt does not have. It has moved into `scheduleRequestTimeout` - every attempt passes through it, and it is the last point before the write - and the budget is now a static taken off the future, with `Long.MAX_VALUE` for an exchange that is not bounded as a whole so no call site needs a special case. Resolving the configured timeout moved with it, so it is no longer resolved in two places.

The rest: the budget's javadoc is back on the budget rather than on the test accessor an earlier edit of mine slipped in front of it; the comment about the deadline being left negative now names what actually depends on it; three comments that described the change rather than the code are rewritten, `AGENTS.md` being clear on that; and the test that asserted nothing - a per-attempt exchange answered `isDeadlinePassed` on its first conjunct - asserts on the computed deadline instead.

Carrying the read timeout across a redirect changes behaviour outside this flag, so it now has its own heading in the description above rather than a line in the change list. There is no release notes file in the repo; say where you would like it and I will put it there.

`./mvnw clean verify` green, and the fail-fast case was re-checked by mutation after the guard moved. JDK 17 locally, no 11 on this machine.

hyperxpro (approved):

hyperxpro:
Thanks a lot!

Issue #1643, opened 2019-06-13
Multi-part Request body is not kept after 307 Redirects


prabhash1785:
In my current usage of async-http-client, I notice that multi-part request body is not kept after server sends a 307 redirect for HTTP Post. It does keep request body for non-multi part data though.

We are using Async Http Client lib Version: 2.2.0

Multi Part Request data is:
Part 1: jsonpart string
Part 2: byte array of mime data

Multi-part data is set as follows:
```
Request request = new RequestBuilder(Methods.POST)
                       .addBodyPart(jsonPart)
                       .addBodyPart(mimePart)
```

Looking at this code, I don't see 307Interceptor is setting MultiPart data. I am hoping I am looking at the right place.
https://github.com/AsyncHttpClient/async-http-client/blob/6e4c1a6eb8a33964aa181721aa96b80f69c3b7bf/client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java#L102-L114

slandelle:
Thanks for reporting

PR #2316, opened 2026-08-25
fix(netty): preserve all redirect body types

mkurz:
## Summary

- Build keep-body redirects from the original request, preserving every supported body representation and per-request setting.
- Clear target-specific routing and credential state when a redirect crosses an origin, including separately stored Cookie objects.
- Preserve an explicit `Content-Length` when replaying a raw `InputStream` or an unknown-length `InputStreamBodyGenerator`.
- Replay resettable streams and fail promptly for consumed raw streams, streamed multipart parts, and vanished files that cannot be replayed safely.
- Pin body bytes, headers, body-selection precedence, caller-owned `ByteBuf` references, and the new failure modes with focused tests.

## Problem

`Redirect30xInterceptor` rebuilds a request when it follows a strict 302, 307, or 308 redirect. Its keep-body copy chain handled form parameters, strings, byte arrays, `ByteBuffer`, body generators, and multipart bodies, but omitted four real request send paths:

- `List<byte[]>` / composite byte arrays
- Netty `ByteBuf`
- `InputStream`
- `File`

The redirected request therefore kept its method and Content-Type but sent zero bytes. The `File` case is especially risky for uploads because the target can accept an apparently valid empty PUT or POST. Reconstructing the request field by field also omitted unrelated per-request state such as the read timeout and range offset, and maintaining a second body-selection chain alongside `NettyRequestFactory` made future drift likely.

This is a pre-existing omission. AHC issue [#1643](https://github.com/AsyncHttpClient/async-http-client/issues/1643) previously fixed the same class of bug for multipart bodies. The copy chain was carried through pull request [#1843](https://github.com/AsyncHttpClient/async-http-client/pull/1843) without a policy discussion. Focused searches found no existing issue or pull request covering these four representations.

## Change

Build a keep-body redirect with `request.toBuilder()` and then replace only redirect-specific state. This preserves all current and future body representations and per-request options without duplicating `NettyRequestFactory.body`. Headers are copied before redirect-only values are removed, so the original request is not mutated.

On a cross-origin redirect, the copied request drops the previous resolved address, virtual host, realm, authorization headers, and Cookie objects before the cookie store adds cookies that legitimately match the new URI. The body is not covered by that boundary. It follows the existing keep-body policy, which means a `File` or `InputStream` body that a cross-origin redirect leg previously received as empty is now sent in full, and a target that keeps redirecting can receive it once per hop up to `maxRedirects`. That is the same exposure byte arrays, strings, form parameters, and multipart bodies already have today.

Composite byte arrays, caller-owned `ByteBuf`s, and files are repeatable. A resettable `InputStream`, such as `ByteArrayInputStream`, also replays. A caller-supplied `Content-Length` is retained for a raw `InputStream` or an `InputStreamBodyGenerator` without a declared length, because neither has an intrinsic size from which to recompute it. A consumed stream that cannot be reset reaches the existing fail-fast guard added in #2312 and completes the future with `IOException`; that is preferable to silently succeeding with an empty body.

An `InputStreamPart` is closed by the first multipart send and has no equivalent replay guard, so a keep-body redirect now fails promptly instead of risking a hang or incomplete multipart request. A selected `File` or `FileBodyGenerator` is also checked before dispatching the redirect; if it disappeared after the first send, the future fails with `IOException` before a target pooled channel can be removed and an unchecked constructor exception can escape. The validation follows `NettyRequestFactory` precedence so a sticky `File` field is ignored when a higher-priority body representation was actually sent.

The change does not alter which methods or status codes keep a body, nor does it introduce a new cross-origin policy. It makes the existing strict-302, 307, and 308 behavior complete for every supported request-body representation.

## Compatibility

There is no public API change. Requests that previously sent an empty body on a keep-body redirect now resend their configured body.

**Behavior changes:**

- A non-resettable `InputStream` on a keep-body redirect previously completed successfully after sending an empty redirected request. It now completes the request future exceptionally with `IOException`. This includes `FileInputStream`, which is closed after the first send and cannot be reset for replay.
- A multipart `InputStreamPart` now fails promptly with `IOException` when a keep-body redirect requires replay. Reusing its already-consumed and closed stream could previously hang or send incomplete multipart content.
- A selected file that disappears between the first request and redirect now fails with `IOException` before redirect dispatch rather than allowing an unchecked `IllegalArgumentException` to escape while constructing the next request.
- A `File` or `InputStream` body is now sent on a keep-body redirect to a different origin, where the redirected request previously carried no body. Credentials are still stripped at that boundary, but the payload is not.

Callers that accidentally relied on an empty or incomplete redirected request will observe an exception, but the failure is explicit instead of silently losing configured content. There is no public API change.

## AI disclosure

OpenAI Codex on behalf of Matthias Kurz. The commit includes `Co-Authored-By: OpenAI Codex <codex@openai.com>` per `AGENTS.md`.

## Test plan

- [x] On untouched `upstream/main`, the focused suite reproduced five failures: four body types arrived as zero bytes and a non-resettable stream incorrectly completed successfully.
- [x] Before the generator fix, a one-argument `InputStreamBodyGenerator` sent `Content-Length: 13` on the first leg and no `Content-Length` on the redirected leg.
- [x] `./mvnw -pl client -Dtest=RedirectBodyTest,RedirectCredentialSecurityTest test` on JDK 11: 40 tests passed, including Netty leak detection.
- [x] `./mvnw clean verify` on JDK 11: 1,496 tests passed and Revapi completed without failures (`BUILD SUCCESS`).

Generated with OpenAI Codex.


mkurz:
We use ahc in https://github.com/playframework/play-ws and I am in the process of upgrading to v3 - and found some thing worth adressing.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -219,6 +206,35 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
         return false;
     }
 
+    private static void copyBody(RequestBuilder requestBuilder, Request request) {
+        requestBuilder.setCharset(request.getCharset());
+
+        // Keep this precedence aligned with NettyRequestFactory.body. A Request can retain a File or
```
hyperxpro: Can we build this from `request.toBuilder()` and override method, uri, headers and realm after? `RequestBuilderBase` already copies all the body fields, plus charset, rangeOffset and readTimeout.

The builder above copies `setRequestTimeout` but not `setReadTimeout`, so a per-request read timeout silently falls back to the config default after the first redirect. `toBuilder()` would fix that too, and this wouldn't drift out of sync with `NettyRequestFactory.body` again.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -219,6 +206,35 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
+            requestBuilder.setBody(request.getByteData());
+        } else if (request.getCompositeByteData() != null) {
+            requestBuilder.setBody(request.getCompositeByteData());
+        } else if (request.getStringData() != null) {
+            requestBuilder.setBody(request.getStringData());
+        } else if (request.getByteBufferData() != null) {
+            requestBuilder.setBody(request.getByteBufferData());
+        } else if (request.getByteBufData() != null) {
+            requestBuilder.setBody(request.getByteBufData());
+        } else if (request.getStreamData() != null) {
```
hyperxpro: `WriteProgressListener` closes the stream after the first leg, so this only replays if `close()` is a no-op. `ByteArrayInputStream` is fine, `Files.newInputStream(...)` and anything buffered will fail the `reset()` in `NettyInputStreamBody`.

We used to send an empty body on the redirect leg, so this turns a silent bug into an `IOException` for the common case. That's the right trade I think, but it needs a release note, and a test with a real `FileInputStream` and not just `ByteArrayInputStream`.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -219,6 +206,35 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
+            requestBuilder.setBody(request.getStreamData());
```
hyperxpro: `propagatedHeaders` removes `CONTENT_LENGTH` and the rebuilt `NettyInputStreamBody` doesn't have a length, so a stream that went out with `Content-Length` on the first leg goes out chunked on the second. Some targets won't accept a chunked request body. Can we carry the original length over?

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -219,6 +206,35 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
+        } else if (isNonEmpty(request.getFormParams())) {
+            requestBuilder.setFormParams(request.getFormParams());
+        } else if (isNonEmpty(request.getBodyParts())) {
```
hyperxpro: `InputStreamMultipartPart` closes its stream once it has written it, and we don't have a consumed check here like the one `NettyInputStreamBody` got in #2312.

So for an `InputStreamPart` the replayed request still advertises the declared `Content-Length`, but `transferContentTo` sees `-1` straight away and writes nothing. The server then waits for a body that never arrives. Should we detect the parts we can't replay and fail fast the same way?

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -219,6 +206,35 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
+            requestBuilder.setBodyParts(request.getBodyParts());
+        } else if (request.getFile() != null) {
```
hyperxpro: Today the `File` is dropped on redirect, so a 307 to another origin sends nothing. After this we upload the whole file to whatever host `Location` points at, and a hostile target can bounce us `maxRedirects` times to get several copies.

Same origin is fine. Cross origin I'd like to be a deliberate decision, and same for `getStreamData()` above.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -219,6 +206,35 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
+            requestBuilder.setBody(request.getFile());
```
hyperxpro: `NettyFileBody` throws `IllegalArgumentException` if the file is gone by now, and we're on the event loop here. Nothing in `newNettyRequestAndResponseFuture` catches it, so it ends up in `exceptionCaught` after `drainChannelAndOffer` already gave the old channel back to the pool, and then we close a connection another request might be using. Check the file here and abort the future with an `IOException` instead?

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
```diff
@@ -121,4 +137,134 @@ public void regular307KeepsBody() throws Exception {
...
+            @Override
+            public synchronized void reset() throws IOException {
+                throw new IOException("reset not supported");
+            }
+        };
+        try (AsyncHttpClient c = asyncHttpClient(config().setFollowRedirect(true))) {
+            ExecutionException thrown = assertThrows(ExecutionException.class,
+                    () -> execute307(c.preparePost(getTargetUrl()).setBody(body)));
+
+            assertInstanceOf(IOException.class, thrown.getCause());
```
hyperxpro: Any channel-level failure is an `IOException` too, so this still passes if the replay guard in `NettyInputStreamBody` goes away. Assert on the message instead. The `FilterInputStream` is also never closed.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
```diff
@@ -121,4 +137,134 @@ public void regular307KeepsBody() throws Exception {
...
+
+            assertEquals("field=value", response.getResponseBody());
+        }
+    }
+
+    @RepeatedIfExceptionsTest(repeats = 5)
+    public void multipart307KeepsBody() throws Exception {
+        try (AsyncHttpClient c = asyncHttpClient(config().setFollowRedirect(true))) {
+            Response response = c.preparePost(getTargetUrl())
+                    .addBodyPart(new StringPart("field", "multipart value"))
```
hyperxpro, line 394: `StringPart` is replayable so this passes either way. The one that breaks is `InputStreamPart` with a declared length, please add that one.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Done - the keep-body path now builds from `request.toBuilder()` and only overrides method, URI, headers, realm and follow-redirect. `readTimeout` and `rangeOffset` are now carried on both paths (the non-keep-body branch sets them explicitly), and `copyBody` is gone.

Two things fell out of copying everything. `toBuilder()` also copies the resolved address, virtual host and Cookie *objects*, so those are cleared on a cross-origin redirect - the Cookie objects are separate from the Cookie header that `propagatedHeaders` strips. And `propagatedHeaders` was calling `.remove()` directly on `request.getHeaders()`, which is the live instance; that was harmless while the original request was discarded, but would have corrupted the prototype once we started building from it. It now copies first.

One place the precedence check does come back: `selectedBodyFile()` re-encodes the `NettyRequestFactory.body` order, because validating the `File` (your comment below) requires knowing whether the file is the representation that will actually be sent - a `File` can stay set behind a higher-priority body. I kept it as a pre-check rather than catching `NettyFileBody`'s `IllegalArgumentException`, since you asked for the failure to surface before dispatch. Happy to swap it for the catch if you'd rather not have the ordering repeated anywhere.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Added `fileInputStream307FailsPromptly`, which uses `Files.newInputStream(...)` rather than `ByteArrayInputStream` and asserts the exact guard message. The Compatibility section of the PR body now calls out `FileInputStream` by name as a behavior change: it previously completed successfully after sending an empty redirected request, and now fails with `IOException`.

On the release note - I don't see a `CHANGELOG` in the repo or a convention in `AGENTS.md`, so I've treated the PR body as the artifact. Is that what you had in mind, or is there somewhere else this should be recorded?

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Added - `ensureBodyReplayable` now rejects any `InputStreamPart` before the request is rebuilt, so it fails ahead of any channel work rather than advertising a `Content-Length` it can't satisfy. Test uses an `InputStreamPart` with a declared length, since `StringPart` is replayable and passes either way.

One thing worth your call: this `IOException` (and the vanished-file one below) propagates out of `exitAfterHandlingRedirect` into `HttpHandler.handleRead`, which routes `IOException` through `applyIoExceptionFiltersAndReplayRequest`. With no `IOExceptionFilter` configured, that is a straight abort. If a configured `IOExceptionFilter` requests replay and the future remains replayable, these failures can be retried up to `maxRequestRetry`. Avoiding that would require either a non-`IOException` exception or an explicit exclusion before applying the filters; an `IOException` subclass alone would not change the routing. Do you want either of those, or is the default-config behavior enough?

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Agreed that this is security-sensitive, so let me be precise about what does and does not change.

The exposure is not new in kind: several body representations, including byte arrays, strings, `ByteBuffer`, form parameters, multipart bodies, and replayable body generators, already replay across origins on 307, 308, and strict 302, once per hop up to `maxRedirects`.

This PR makes composite byte arrays, `ByteBuf`, `InputStream`, and `File` bodies follow that existing policy too. The most consequential change is for `File` and `InputStream`: a redirect leg that previously received an empty body can now receive the complete upload, which may be materially larger. That is worth stating plainly.

What I would rather not do is carve out particular body representations. Dropping only those preserves the method while sending an empty payload, which is exactly the silent data-corruption bug this PR fixes, and would make cross-origin behavior depend on which body setter the caller used.

The redirect boundary still strips Authorization, Proxy-Authorization, the Cookie header, Cookie objects, the realm, and target-specific routing state. The cookie store can only contribute cookies valid for the redirected URI.

If AHC should let callers refuse cross-origin body replay, I think that belongs as a representation-independent policy applied to every keep-body redirect. Because retaining the method while stripping its body is unsafe, I would prefer such a policy to decline or fail the redirect rather than silently remove the payload. I'll open a follow-up issue for that design and keep this PR focused on faithfully replaying bodies under the current policy. Say the word if you would prefer that policy to be included here.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Done in 6a7842dab. `propagatedHeaders` now preserves a caller-supplied `Content-Length` specifically when replaying a raw `InputStream` on a keep-body redirect; other body representations still have the header removed and recomputed normally. `inputStream307PreservesExplicitContentLength` asserts that both legs carry the original `Content-Length` and that the redirected request delivers the original bytes.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Done in 6268d4d8f. `ensureBodyReplayable` determines whether a `File` or `FileBodyGenerator` was the representation actually selected, checks it before rebuilding or dispatching the redirect, and throws `IOException` if it has disappeared. `vanishedFile307FailsPromptly` deletes the file between the first response and redirect and asserts the exact error. The coexistence test also confirms that a stale, lower-priority `File` is not rejected when another body representation was actually sent. The `IOExceptionFilter` caveat is covered in the `InputStreamPart` thread above.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
mkurz: Added in 858927ffe. `inputStreamMultipart307FailsPromptly` uses a real file-backed `InputStreamPart` with a declared length and asserts the exact fail-fast `IOException`. The `StringPart` test remains to pin the replayable multipart path.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
mkurz: Done in 67145c7a4. The `FilterInputStream` is now managed by try-with-resources, and the test asserts the exact replay-guard message, `HTTP/1 request body InputStream already consumed and cannot be reset for a retry`, in addition to checking the exception type. An unrelated channel-level `IOException` therefore no longer satisfies the test.

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
```diff
@@ -535,6 +535,26 @@ void crossDomainRedirectStripsCookieHeader() throws Exception {
         }
     }
 
+    @Test
+    void crossDomainRedirectStripsCookieObject() throws Exception {
```
hyperxpro, line 542: GET plus a 302 means `keepBody` is false, so we take the fresh builder, which never had cookies. `resetCookies()` is a no-op here, delete it and the test still passes. It needs a keep-body redirect to mean anything, and watch out that `/target-307` doesn't record the Cookie header the way `/target` does.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -229,10 +231,50 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
+
     private static HttpHeaders propagatedHeaders(Request request, Realm realm, boolean keepBody, boolean stripAuthorization) {
-        HttpHeaders headers = request.getHeaders()
-                .remove(HOST)
-                .remove(CONTENT_LENGTH);
+        HttpHeaders headers = request.getHeaders().copy().remove(HOST);
+
+        // A raw InputStream has no intrinsic length from which NettyRequestFactory can rebuild this header.
+        // Preserve a caller-supplied value when the stream itself is replayed.
+        if (!keepBody || request.getStreamData() == null) {
```
hyperxpro: `getStreamData() == null` isn't the same as having no intrinsic length. `InputStreamBodyGenerator` with the 1-arg ctor is also -1 and builds the same `NettyInputStreamBody`, so that one still goes out chunked. Can we key on the selected body having no length instead?

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
```diff
@@ -121,4 +152,243 @@ public void regular307KeepsBody() throws Exception {
             assertEquals(receivedContentType, contentType);
         }
     }
+
+    @RepeatedIfExceptionsTest(repeats = 5)
+    public void redirectPreservesPerRequestSettings() throws Exception {
```
hyperxpro, line 158: This is a 307, so `toBuilder()` copies `readTimeout` and `rangeOffset` anyway and the test passes even with the new `setReadTimeout` reverted. The gap is on the fresh builder, can you add a 301 or 303 case?

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -229,10 +231,50 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
+
+        File file = selectedBodyFile(request);
+        if (file != null && !file.isFile()) {
+            throw new IOException("Redirect request body file " + file.getAbsolutePath()
+                    + " is not a file or does not exist");
+        }
+    }
+
+    private static File selectedBodyFile(Request request) {
+        // Keep this precedence aligned with NettyRequestFactory.body. A File can remain set alongside a
```
hyperxpro, line 251: Pre-check is fine, I'd rather have it before dispatch than catch the `IllegalArgumentException` later. Keep it as is.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -229,10 +231,50 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
 
+    private static void ensureBodyReplayable(Request request) throws IOException {
+        for (Part part : request.getBodyParts()) {
+            if (part instanceof InputStreamPart) {
+                throw new IOException("Multipart InputStream body part '" + part.getName()
+                        + "' cannot be replayed after redirect");
+            }
+        }
+
+        File file = selectedBodyFile(request);
```
hyperxpro, line 243: Correction on my earlier comment: `drainChannelAndOffer` only arms an `OnLastHttpContentCallback`, so the channel wasn't in the pool yet and nothing was at risk. The guard is still right for the exception type. Two leftovers, both fine to leave: a file that vanishes between this check and `NettyFileBody`'s ctor still throws `IllegalArgumentException`, and a `FilePart` file isn't covered.

hyperxpro (changes requested):

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
mkurz: Fixed in aae06ef82.

The regression now uses a POST followed by a cross-origin 307, so it exercises the keep-body `request.toBuilder()` path. The origin handler verifies that the Cookie object was sent on the first leg, while `/target-307` records the redirected Cookie header and verifies that it is absent. I also confirmed that the test fails with `session=abc123` at the target if `resetCookies()` is removed.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Fixed in 6bec3fb24.

`propagatedHeaders` now bases preservation on the body representation actually selected, using the same precedence as `NettyRequestFactory.body()`, rather than checking only `getStreamData()`. It preserves an explicit `Content-Length` for both a raw `InputStream` and an `InputStreamBodyGenerator` whose declared length is negative. Known-length representations still have the header removed and recomputed.

A generic `BodyGenerator` can also produce an unknown length, but that length is only discoverable by calling `createBody()`, which is the one-shot body creation. Those generators therefore continue to fall back to chunked encoding on the redirected leg, as they do today. The other body representations have known lengths, and multipart cannot reach this path with an unknown length because `ensureBodyReplayable` rejects `InputStreamPart`.

Before the fix, the new one-argument `InputStreamBodyGenerator` regression observed `Content-Length` values `[13, null]` on the original and redirected legs. It now verifies `[13, 13]` and that the redirected request delivers the original bytes.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
mkurz: Added in 649f8a2a6.

The existing 307 test remains for the keep-body/`toBuilder()` path. The new test uses POST followed by 303, so it exercises the fresh `RequestBuilder` path, and observes both requests through a `ResponseFilter`. It verifies that both legs retain the per-request read timeout and range offset, while the redirected GET remains bodyless.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Thanks, agreed. I kept the pre-dispatch check as-is.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Thanks for checking and for correcting the pooled-channel rationale. The pre-check remains useful because it turns a missing file that is detectable at that point into a controlled `IOException` before redirect dispatch. I agree that it cannot eliminate the post-check TOCTOU window and that `FilePart` is a separate case; I am leaving both outside this PR.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -229,10 +232,73 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
+        if (request.getStreamData() != null) {
+            return true;
+        }
+        if (!request.getFormParams().isEmpty()
+                || !request.getBodyParts().isEmpty()
+                || request.getFile() != null) {
+            return false;
+        }
+        return request.getBodyGenerator() instanceof InputStreamBodyGenerator
+                && ((InputStreamBodyGenerator) request.getBodyGenerator()).getContentLength() < 0;
```
hyperxpro: `InputStreamBodyGenerator` isn't the only one that can come back with -1. Everything else lands in `NettyBodyBody`, and `PushBody` reports -1, so a feedable or third-party generator still loses the caller's `Content-Length` and goes out chunked on the redirect leg only.

```suggestion
        if (request.getBodyGenerator() instanceof InputStreamBodyGenerator) {
            return ((InputStreamBodyGenerator) request.getBodyGenerator()).getContentLength() < 0;
        }
        return request.getBodyGenerator() != null
                && !(request.getBodyGenerator() instanceof FileBodyGenerator);
```

That treats `ByteArrayBodyGenerator` as unknown as well, which just means we keep the caller's header instead of recomputing the same number. Add it to the exclusion if you'd rather be exact, it needs an import.

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
```diff
@@ -535,6 +538,27 @@ void crossDomainRedirectStripsCookieHeader() throws Exception {
...
+
+            client.preparePost("http://127.0.0.1:" + portA + "/redirect-307-to-b")
+                    .addCookie(new DefaultCookie("session", "abc123"))
+                    .setBody("request-body-content")
+                    .execute()
+                    .get(5, TimeUnit.SECONDS);
+
+            assertEquals("session=abc123", lastCookieHeaderOnA.get());
+            assertNull(cookieOn307Target.get(),
+                    "Cookie objects must not be copied to a cross-domain redirect target");
```
hyperxpro, line 560: Nothing here pins that leg 2 actually ran. If the redirect stopped being followed the 307 comes back normally and `assertNull` still passes. `/target-307` already records the body, so:

```suggestion
            lastCookieHeaderOnA.set(null);
            cookieOn307Target.set(null);
            bodyOn307Target.set(null);

            client.preparePost("http://127.0.0.1:" + portA + "/redirect-307-to-b")
                    .addCookie(new DefaultCookie("session", "abc123"))
                    .setBody("request-body-content")
                    .execute()
                    .get(5, TimeUnit.SECONDS);

            assertEquals("session=abc123", lastCookieHeaderOnA.get());
            assertEquals("request-body-content", bodyOn307Target.get());
            assertNull(cookieOn307Target.get(),
                    "Cookie objects must not be copied to a cross-domain redirect target");
```

hyperxpro (changes requested):

hyperxpro (approved):

hyperxpro:
Thanks a lot!

PR #2317, opened 2026-08-26
fix(netty): preserve QUERY across redirects

mkurz:
## Summary

- Preserve the QUERY method and request content when following 301, 302, 307, and 308 redirects, as required by RFC 10008.
- Keep the existing 303 behavior, which follows the redirect with GET and drops the request content.
- Add a `HttpConstants.Methods.QUERY` constant and name the redirect-policy decisions explicitly.
- Cover every QUERY redirect status, strict and non-strict 302 handling, existing POST/PUT/PATCH/DELETE behavior, repeatable and non-repeatable generated bodies, and cross-origin redirects.

## Problem

`Redirect30xInterceptor` treated every method other than GET, HEAD, and OPTIONS like POST when handling 301 and non-strict 302 responses. As a result, a QUERY request was changed to GET and its query content and Content-Type were dropped.

[RFC 10008 section 2.5](https://www.rfc-editor.org/rfc/rfc10008.html#section-2.5) explicitly says that the POST-to-GET exceptions for 301 and 302 do not apply to QUERY. A QUERY request must instead be repeated with its content for 301, 302, 307, and 308. Only a 303 response calls for a GET request to the redirect target.

## Change

Recognize QUERY in the redirect policy so 301 and 302 retain the original method and body. The existing 307, 308, 303, HEAD, OPTIONS, and POST behavior is unchanged. The implementation uses named decisions rather than embedding the QUERY exception in one compound expression.

Cross-origin QUERY redirects retain their method, content, and Content-Type, as required to repeat the query. This means a QUERY body now crosses origins on 301 and 302 where the old, incorrect GET rewrite dropped it. AHC already uses that body-replay trust model for 307 and 308. Existing redirect security still strips Authorization, Realm credentials, and user-supplied Cookie headers before sending the request to the new origin. A separate, representation-independent policy that refuses cross-origin keep-body redirects could be considered, but this conformance fix does not currently add one.

The broader pre-existing behavior that converts PUT, PATCH, DELETE, and other methods to GET after 301 and non-strict 302 responses is deliberately out of scope. It changes established behavior for existing users and is handled in a separate follow-up branch, `fix/non-post-redirects`.

That follow-up handles QUERY as an ordinary non-POST method and therefore supersedes this pull request's QUERY-specific interceptor condition if both land. Keeping this narrow pull request separate still allows the standardized QUERY behavior to land even if the broader compatibility change is rejected; the public method constant and QUERY-specific regression coverage remain useful either way.

This adds the public `HttpConstants.Methods.QUERY` string constant. It is an additive, user-facing API for constructing QUERY requests; there is no incompatible API change.

## AI disclosure

## Test plan

- [x] Reproduced the failure on JDK 11 with `./mvnw -pl client -Dtest=RedirectBodyTest#query301KeepsMethodAndBody test`.
- [x] `./mvnw -pl client -Dtest=RedirectBodyTest test` on JDK 11.
- [x] `./mvnw -pl client -Dtest=RedirectBodyTest,RedirectCredentialSecurityTest test` on JDK 11: 57 tests passed.
- [x] `./mvnw clean verify` on JDK 11: BUILD SUCCESS (full reactor, including tests, Javadocs, artifact signing, coverage, and Revapi).

Generated with OpenAI Codex.


mkurz:
We use ahc in https://github.com/playframework/play-ws and I am in the process of upgrading to v3 - and found some thing worth adressing.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -111,11 +112,19 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
+                        OPTIONS.equals(originalMethod) || HEAD.equals(originalMethod);
+                boolean strict302 = statusCode == FOUND_302 && config.isStrict302Handling();
+                boolean queryRedirect = isQuery &&
+                        (statusCode == MOVED_PERMANENTLY_301 || statusCode == FOUND_302);
+                boolean legacyRedirectToGet = statusCode == MOVED_PERMANENTLY_301 ||
+                        (statusCode == FOUND_302 && !strict302);
+                boolean switchToGet = !methodAlreadyPreserved &&
+                        (statusCode == SEE_OTHER_303 || (!isQuery && legacyRedirectToGet));
+                boolean keepBody = queryRedirect ||
+                        statusCode == TEMPORARY_REDIRECT_307 || statusCode == PERMANENT_REDIRECT_308 ||
```
hyperxpro, line 132: This sends the QUERY body to whatever host the 301 points at. We strip `Authorization` and `Cookie` when the origin changes, but with QUERY the body is the query, so a cached or injected 301 gets the whole thing where before it got an empty GET. 307 and 308 already behave like this, but 301 and 302 are the cacheable ones. Can we require `sameBase` for the body here, or gate it on a config flag?

hyperxpro, line 132: The block this feeds, at line 158, covers six of the body kinds. `getFile()`, `getStreamData()`, `getByteBufData()` and `getCompositeByteData()` all fall through, and QUERY is not POST/PUT/PATCH so we do not set `ZERO_CONTENT_LENGTH` either. A QUERY with a `File` body then gets redirected with its `Content-Type` and no content at all, which is the case this PR is meant to fix. It is pre existing for 307 and 308, but this PR is what makes it reachable for QUERY.

Its order does not match `NettyRequestFactory.body()` either. That one takes `byteData` before `formParams` and `bodyParts` before `bodyGenerator`, we do the reverse, so a request carrying both sends a different body after the redirect than it did on the first hop.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -111,11 +112,19 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
-                        !originalMethod.equals(HEAD) &&
-                        (statusCode == MOVED_PERMANENTLY_301 || statusCode == SEE_OTHER_303 || statusCode == FOUND_302 && !config.isStrict302Handling());
-                boolean keepBody = statusCode == TEMPORARY_REDIRECT_307 || statusCode == PERMANENT_REDIRECT_308 || statusCode == FOUND_302 && config.isStrict302Handling();
+                boolean isQuery = QUERY.equals(originalMethod);
+                boolean methodAlreadyPreserved = GET.equals(originalMethod) ||
+                        OPTIONS.equals(originalMethod) || HEAD.equals(originalMethod);
+                boolean strict302 = statusCode == FOUND_302 && config.isStrict302Handling();
+                boolean queryRedirect = isQuery &&
+                        (statusCode == MOVED_PERMANENTLY_301 || statusCode == FOUND_302);
+                boolean legacyRedirectToGet = statusCode == MOVED_PERMANENTLY_301 ||
```
hyperxpro, line 127: `queryRedirect` and `legacyRedirectToGet` are each read once, on the next line. `methodAlreadyPreserved` also reads backwards: it is false for QUERY, but the method is preserved for QUERY on 301. Did you consider scoping the rewrite to POST instead? RFC 9110 only defines it for POST, so QUERY would need no special case at all. It changes PUT and DELETE too so probably its own PR, but then you are not adding a branch you have to take out again later.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -111,11 +112,19 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
                 future.setScramContext(null);
 
                 String originalMethod = request.getMethod();
```
hyperxpro, line 119: Can you put the RFC 10008 2.5 reference in a comment here? Every other odd thing in this method says why it is there, and this is the least obvious line in it.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -111,11 +112,19 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
-                boolean switchToGet = !originalMethod.equals(GET) &&
-                        !originalMethod.equals(OPTIONS) &&
-                        !originalMethod.equals(HEAD) &&
-                        (statusCode == MOVED_PERMANENTLY_301 || statusCode == SEE_OTHER_303 || statusCode == FOUND_302 && !config.isStrict302Handling());
-                boolean keepBody = statusCode == TEMPORARY_REDIRECT_307 || statusCode == PERMANENT_REDIRECT_308 || statusCode == FOUND_302 && config.isStrict302Handling();
+                boolean isQuery = QUERY.equals(originalMethod);
+                boolean methodAlreadyPreserved = GET.equals(originalMethod) ||
```
hyperxpro: Nit: these three read `originalMethod.equals(GET)` before. `getMethod()` is never null, so can we leave them as they were and keep the diff to the actual change?

client/src/main/java/org/asynchttpclient/util/HttpConstants.java
```diff
@@ -33,6 +33,7 @@ public static final class Methods {
         public static final String PATCH = HttpMethod.PATCH.name();
         public static final String POST = HttpMethod.POST.name();
         public static final String PUT = HttpMethod.PUT.name();
+        public static final String QUERY = "QUERY";
```
hyperxpro: Netty 4.2.17 already has `HttpMethod.QUERY`, so this can be derived like the others.

```suggestion
        public static final String QUERY = HttpMethod.QUERY.name();
```

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
```diff
@@ -118,7 +142,101 @@ public void regular307KeepsBody() throws Exception {
 
             Response response = c.preparePost(getTargetUrl()).setHeader(CONTENT_TYPE, contentType).setBody(body).setHeader("X-REDIRECT", "307").execute().get(TIMEOUT, TimeUnit.SECONDS);
             assertEquals(response.getResponseBody(), body);
+            assertEquals(POST, receivedMethod);
             assertEquals(receivedContentType, contentType);
         }
     }
+
+    @RepeatedIfExceptionsTest(repeats = 5)
+    public void regular308KeepsBody() throws Exception {
```
hyperxpro, line 184: The condition you rewrote also drives PUT, PATCH and DELETE, but every new test is POST or QUERY. One parameterized test over those three on 301 and 302 would pin the behavior the PR says it is not touching.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
```diff
@@ -118,7 +142,101 @@ public void regular307KeepsBody() throws Exception {
...
+        queryRedirectKeepsMethodAndBody(307, false);
+    }
+
+    @RepeatedIfExceptionsTest(repeats = 5)
+    public void query308KeepsMethodAndBody() throws Exception {
+        queryRedirectKeepsMethodAndBody(308, false);
+    }
+
+    @RepeatedIfExceptionsTest(repeats = 5)
+    public void query301KeepsRepeatableBodyGenerator() throws Exception {
```
hyperxpro, line 240: Can we get one with a non repeatable generator too? A QUERY on 301 used to go out as a bodiless GET and complete, now it replays a consumed stream and fails the future through the guard from #2312. `ByteArrayBodyGenerator` is repeatable so this test does not see it.

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
```diff
@@ -508,6 +532,36 @@ void redirect308CrossDomainStripsAuthButPreservesBody() throws Exception {
...
+                    .execute()
+                    .get(5, TimeUnit.SECONDS);
+
+            assertNull(query301AuthOnTarget.get(),
+                    "Authorization must be stripped on a cross-origin QUERY redirect");
+            assertNull(query301CookieOnTarget.get(),
+                    "Cookie must be stripped on a cross-origin QUERY redirect");
+            assertEquals(QUERY, query301MethodOnTarget.get());
+            assertEquals("application/query", query301ContentTypeOnTarget.get());
+            assertEquals("sensitive-query", query301BodyOnTarget.get());
```
hyperxpro, line 564: This locks in the cross origin body replay I asked about in the interceptor. If we keep it that is fine, but I would rather decide that first than have a test assert it.

hyperxpro (changes requested):

mkurz:
- Putting on hold, waiting for #2316 first.

mkurz:
@hyperxpro sorry, hit the "review request" button to early. I will address you comments and ping you when I am done.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Agreed that this needs an explicit policy decision. My preference for this conformance fix is to retain AHC’s existing keep-body redirect policy: AHC already replays request bodies across origins for 307, 308, and strict 302 while stripping credentials.

Requiring `sameBase` only for QUERY on 301/302 cannot safely mean dropping the body: that would preserve the QUERY method while silently changing the query itself, contrary to RFC 10008.

On the config-flag option: I think that is the right mechanism if we want this restriction, but it should apply uniformly to every cross-origin keep-body redirect, not only QUERY on 301/302, and it should refuse the redirect rather than strip the body. Your caching point is well taken: when a 301/302 response is cached, it can repeatedly cause cross-origin body replay without a new redirect response. That strengthens the case for a configurable cross-origin body-replay policy; I do not think it argues for QUERY following a different policy from every other body-bearing method.

Would you prefer that uniform policy to be designed as part of this PR, or is retaining AHC’s existing trust model acceptable here while we discuss the broader policy separately?

mkurz: This is now fixed by #2316, which has merged into `main` and is included in this branch after rebasing. Redirect body replay is centralized in `copyBody`, covers every request-body representation, and follows the same precedence as `NettyRequestFactory.body()`. The QUERY keep-body path therefore no longer drops `File`, `InputStream`, `ByteBuf`, or composite byte-array bodies or selects a different coexisting representation.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: I have a local follow-up branch which already implements this: the legacy rewrite is scoped to POST, PUT/PATCH/DELETE/custom methods are tested on 301 and 302, and the QUERY-specific condition in this PR becomes unnecessary.

There are two possible ways to sequence it: land this narrow PR and follow it with the broader change, or close this PR and open the broader one with the QUERY constant and regression tests carried over. I lean toward the latter given your point about not adding a branch that is removed again, but the narrow PR would allow QUERY conformance to land independently if the compatibility change for existing PUT/PATCH/DELETE users needs more discussion. Which would you prefer?

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Added the RFC 10008 section 2.5 explanation directly above `queryRedirect` in d3f4d9ac1.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Done in 8171795e2. The three pre-existing comparisons use `originalMethod.equals(...)` again; only the new QUERY comparison retains its new form.

client/src/main/java/org/asynchttpclient/util/HttpConstants.java
mkurz: Done in f6a09efec: `Methods.QUERY` is now derived from `HttpMethod.QUERY.name()`.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
mkurz: Added in a977a518a. The parameterized test covers PUT, PATCH, and DELETE across both 301 and non-strict 302 and asserts the current GET rewrite, empty body, and removed Content-Type. If the broader POST-scoped rewrite lands later, these assertions will require an explicit update.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
mkurz: Added in 69393b5fb. The test uses a non-empty, non-resettable `InputStreamBodyGenerator` and asserts the exact replay-guard `IOException`. Before the QUERY change, the request became a bodyless GET and completed, so the test discriminates the changed behavior.

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
mkurz: Agreed. The test is intended to make the selected policy explicit, not decide it implicitly. If we retain AHC’s existing cross-origin keep-body policy, it should remain because it verifies both request preservation and credential stripping. If you prefer the uniform configuration discussed in the interceptor thread to be added here, I will update the implementation and this test together.

mkurz:
@hyperxpro I have addressed all concrete review feedback in five follow-up commits and rerun `./mvnw clean verify` successfully. The implementation is ready from my side.

Two policy decisions remain:

1. For cross-origin body replay, my preference is to retain AHC’s existing keep-body policy in this conformance fix and discuss a representation-independent option that refuses all cross-origin keep-body redirects separately.
2. For scope, I have a local branch that limits the legacy rewrite to POST and therefore subsumes the QUERY-specific interceptor condition here. I lean toward replacing this PR with that broader change, but keeping this PR allows QUERY conformance to land independently if the PUT/PATCH/DELETE compatibility change needs more discussion.

Please let me know which scope you prefer and whether retaining the existing cross-origin policy is acceptable here.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -116,11 +117,20 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
+                boolean strict302 = statusCode == FOUND_302 && config.isStrict302Handling();
+                // RFC 10008 section 2.5 excludes QUERY from the legacy POST-to-GET behavior.
+                boolean queryRedirect = isQuery &&
+                        (statusCode == MOVED_PERMANENTLY_301 || statusCode == FOUND_302);
+                boolean legacyRedirectToGet = statusCode == MOVED_PERMANENTLY_301 ||
+                        (statusCode == FOUND_302 && !strict302);
+                boolean switchToGet = !methodAlreadyPreserved &&
+                        (statusCode == SEE_OTHER_303 || (!isQuery && legacyRedirectToGet));
+                boolean keepBody = queryRedirect ||
+                        statusCode == TEMPORARY_REDIRECT_307 || statusCode == PERMANENT_REDIRECT_308 ||
```
hyperxpro: Fine by me, let's keep the existing trust model here. Couple of corrections to my own comment though. 301 and 308 are the heuristically cacheable ones, not 301 and 302, and we already replay on 308, so caching isn't what makes this different. And ignore the sameBase idea, you were right that it would keep the method and drop the body. What still nags me is that 301 and 302 are what a redirect helper emits by default, so they're what an open redirect returns, while 307 and 308 have to be asked for. That's a policy question for the whole interceptor though, not for this PR. I'll open a separate issue.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -116,11 +117,20 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
-                        (statusCode == MOVED_PERMANENTLY_301 || statusCode == SEE_OTHER_303 || statusCode == FOUND_302 && !config.isStrict302Handling());
-                boolean keepBody = statusCode == TEMPORARY_REDIRECT_307 || statusCode == PERMANENT_REDIRECT_308 || statusCode == FOUND_302 && config.isStrict302Handling();
+                boolean isQuery = QUERY.equals(originalMethod);
+                boolean methodAlreadyPreserved = originalMethod.equals(GET) ||
+                        originalMethod.equals(OPTIONS) || originalMethod.equals(HEAD);
+                boolean strict302 = statusCode == FOUND_302 && config.isStrict302Handling();
+                // RFC 10008 section 2.5 excludes QUERY from the legacy POST-to-GET behavior.
+                boolean queryRedirect = isQuery &&
+                        (statusCode == MOVED_PERMANENTLY_301 || statusCode == FOUND_302);
+                boolean legacyRedirectToGet = statusCode == MOVED_PERMANENTLY_301 ||
```
hyperxpro: I'd keep this one. Your follow up also makes PUT, PATCH and DELETE keep their bodies on 301 and 302, and that's a much bigger change to behaviour that's been there since 2015. QUERY has no installed base so it's the safe half to go first. Also for the squash message, the PR body mentions a `copyBody` helper from #2316 but there's no such method, it landed as `request.toBuilder()`.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
hyperxpro: Fine by me, let's keep the existing trust model here. Two things I got wrong above: 301 and 308 are the heuristically cacheable ones, not 301 and 302, and we already replay on 308, so caching isn't the difference. And forget the sameBase idea, you were right, it would keep the method and drop the body. My worry is just that 301 and 302 are what a redirect helper emits by default, so that's what an open redirect gives you, while 307 and 308 have to be asked for. But that's about the interceptor in general, not this PR. I'll open a separate issue.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
hyperxpro: I'd keep this one. Your follow up also makes PUT, PATCH and DELETE keep their bodies on 301 and 302, and that's a much bigger change to behaviour that's been there since 2015. QUERY is new so nobody depends on it yet, it's the safer one to go first. Also for the squash message, the PR body mentions a `copyBody` helper from #2316 but there's no such method, it landed as `request.toBuilder()`.

hyperxpro (changes requested):
Last round

mkurz:
Since I can no longer reply inline after the force-push:

> Fine by me, let's keep the existing trust model here. Two things I got wrong above: 301 and 308 are the heuristically cacheable ones, not 301 and 302, and we already replay on 308, so caching isn't the difference. And forget the sameBase idea, you were right, it would keep the method and drop the body. My worry is just that 301 and 302 are what a redirect helper emits by default, so that's what an open redirect gives you, while 307 and 308 have to be asked for. But that's about the interceptor in general, not this PR. I'll open a separate issue.

Thanks for clarifying, and for taking the general follow-up issue. I kept AHC's existing cross-origin keep-body trust model unchanged in this PR. The branch is now rebased onto latest `main` and squashed to 94bacf1cb; `./mvnw clean verify` passes on JDK 11.

> I'd keep this one. Your follow up also makes PUT, PATCH and DELETE keep their bodies on 301 and 302, and that's a much bigger change to behaviour that's been there since 2015. QUERY is new so nobody depends on it yet, it's the safer one to go first. Also for the squash message, the PR body mentions a `copyBody` helper from #2316 but there's no such method, it landed as `request.toBuilder()`.

Agreed. I kept this PR narrowly scoped to QUERY; the existing PUT, PATCH, and DELETE behavior remains unchanged and is pinned by regression tests. The broader POST-only rewrite remains a separate follow-up.

Also corrected: #2316 landed the replay preservation through `request.toBuilder()`, not a `copyBody` helper. Neither the current PR description nor the final squash commit message refers to `copyBody`.

The branch is now rebased onto latest `main` and squashed to 94bacf1cb; `./mvnw clean verify` passes on JDK 11.

hyperxpro (approved):

hyperxpro:
Thanks a lot!

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: I opened #2326 to collect the redirect-policy discussion here and the related HTTPS-to-HTTP concern from #2325. You mentioned opening a separate issue, so hopefully this saves you that step. This leaves the merged behavior unchanged; we can agree on the policy and defaults there before implementing a follow-up PR.

Issue #1233, opened 2016-08-22
Erroneous read timeouts when using StreamedAsyncHandler


twz123:
When using the Reactive Streams API to consume long running chunked HTTP responses, read timeouts occur on active connections that are actually receiving data constantly from the server. The only way to use Reactive Streams with AHC in this manner is to disable read timeouts.

It seems that `NettyResponseFuture.touch()` isn't called when passing `HttpResponseBodyPart` instances to the `StreamedResponsePublisher` in `AsyncHttpClientHandler`.

Is that intended behavior, or something that has just been overlooked?

In the latter case, what would be the right place to `touch` the future then? Somewhere below [here](https://github.com/AsyncHttpClient/async-http-client/blob/df522adcbce83b8ea0c6e17ab3f9911d6f0a9398/client/src/main/java/org/asynchttpclient/netty/handler/AsyncHttpClientHandler.java#L88) maybe?

slandelle:
Fixed, thanks for reporting!
cc @jroper

twz123:
:+1: for fixing this at speed of light ;)

PR #2318, opened 2026-08-26
Add response body flow control

mkurz:
## Summary

- Add a thread-safe `ResponseBodyControl` callback after final response headers.
- Support suspending, resuming, and cancelling HTTP/1.1 and HTTP/2 response bodies without coupling AHC to a streaming API.
- Pause the network read timeout while reads are intentionally suspended while leaving the request timeout active.
- Keep suspended HTTP/2 streams independent and cover transport backpressure, cancellation, timeouts, and connection reuse.

## Motivation

AHC 3 removed `StreamedAsyncHandler` and its Reactive Streams integration in pull request [#1843](https://github.com/AsyncHttpClient/async-http-client/pull/1843). That removal avoids coupling AHC to a particular streaming library, but `AsyncHandler` by itself has no way to stop transport reads while a downstream consumer has no demand.

[Play WS](https://github.com/playframework/play-ws) is the driving consumer for this change. Play WS needs transport backpressure to preserve its existing Pekko Streams and Reactive Streams response APIs while upgrading to AHC 3. A working adapter on a currently local Play WS development branch turns this control into a single-subscriber Reactive Streams publisher and has been tested against this AHC branch.

This pull request adds only the transport primitive. Streaming-library policy and dependencies remain in Play WS, so AHC does not regain a dependency on Reactive Streams or JDK Flow.

## Semantics

- `AsyncHandler.onResponseBodyStart` runs after final headers and before body parts, including for a response with no body.
- Calls to the supplied control are thread-safe, idempotent, and ignored after response completion.
- `suspend()` stops requesting new transport data, although body parts already read may still be delivered.
- `resume()` permits transport reads again.
- Returning `State.ABORT` is the synchronous callback-time way to stop processing. A handler can retain the control and call `cancel()` when an asynchronous decision is made after the callback returns.
- Cancelling an HTTP/1.1 body closes its connection when bytes may remain unread; cancellation after terminal content has been received can reuse a keep-alive connection. Cancelling an HTTP/2 body closes only its stream.
- Fully consumed responses retain the existing HTTP/1.1 pooling and HTTP/2 parent-connection reuse behavior.
- Suspension pauses only the network read timeout. The request timeout remains active; if the request timeout is disabled, an application that never resumes or cancels can retain the exchange and its transport resources indefinitely.

## HTTP/2 flow control

Connections with no actively suspended response retain Netty's normal connection-level receive-window accounting. The default 65,535-byte connection window therefore continues to cap unconsumed flow-controlled DATA across all streams for users that never call `suspend()`.

When the first response on a connection is suspended, AHC returns connection-level credit that has already accumulated and continues returning that shared credit as DATA arrives. This prevents the suspended stream from exhausting the connection window and starving sibling streams. Per-stream credit remains consumption-driven at all times, so each suspended stream is still bounded by its own receive window. Multiple simultaneous suspensions are counted, and normal connection accounting resumes after the last one ends. Credit already returned and data already queued cannot be revoked, and the controller tracks credit returned early so later application consumption does not return it twice.

During an active suspension, aggregate buffering can still scale with the number of concurrent streams; the relevant controls are `http2InitialWindowSize`, `http2MaxConcurrentStreams`, and the connection limits. The defaults do not impose a hard client-side aggregate bound during that interval: the initial per-stream window is 16 MiB and `http2MaxConcurrentStreams = -1` leaves concurrency server-controlled. A rough upper-bound estimate is connection count times effective concurrent streams times the initial window, excluding network and decoder overhead. Applications requiring a finite policy must configure these values together. A hard aggregate byte budget is a separate design and is outside this pull request.

Netty's connection auto-refill state is private and fixed when `DefaultHttp2LocalFlowController` is constructed, so it cannot be enabled only for the lifetime of a suspension through composition or subclassing. The package-private `SuspensionAwareHttp2LocalFlowController` therefore adapts `DefaultHttp2LocalFlowController` from Netty 4.2.17.Final, preserving its normal behavior while making connection refill suspension-scoped.

This adaptation has a maintenance cost: AHC owns the copied flow-control logic and every Netty upgrade must compare it with the corresponding upstream implementation for correctness and security fixes. The exact source version and that obligation are recorded in the class Javadoc. An upstream Netty API that permits connection auto-refill to be changed at runtime would provide the exit path and allow AHC to remove the adaptation; no such API exists in Netty 4.2.17.Final.

The auto-refill mode requires a custom `Http2Connection`. Netty's builder treats `server()` and `connection()` as mutually exclusive, so `ClientHttp2FrameCodecBuilder` supplies the connection through the protected builder API and overrides `isServer()` to retain client mode.

The explicit `gracefulShutdownTimeoutMillis(0)` is not a new shutdown policy. `Http2FrameCodecBuilder.forClient()` selects zero through its package-private client constructor; the subclass must use the protected no-argument constructor, so it sets zero explicitly to preserve the existing client-factory behavior.

## Scope and commit structure

The API/lifecycle work and the HTTP/2 independence work are kept as separate logical commits, but they belong in one pull request. Without the HTTP/2 work, a suspended response can consume the shared connection window and block unrelated sibling streams, so the public control would not have correct multiplexed behavior. Follow-up review fixes are also split into focused commits covering exchange ownership, terminal cleanup, terminal HTTP/1.1 cancellation, bodyless responses, interim responses, indefinite-suspension diagnostics, and suspension-scoped HTTP/2 refill.

## History checked

- Issue [#544](https://github.com/AsyncHttpClient/async-http-client/issues/544) originally identified the lack of `AsyncHandler` backpressure, and pull request [#963](https://github.com/AsyncHttpClient/async-http-client/pull/963) addressed it by adding Reactive Streams support.
- Issues [#1233](https://github.com/AsyncHttpClient/async-http-client/issues/1233) and [#1721](https://github.com/AsyncHttpClient/async-http-client/issues/1721) document the interaction between downstream demand and read timeouts in the former streamed handler.
- Pull request [#1843](https://github.com/AsyncHttpClient/async-http-client/pull/1843) removed `StreamedAsyncHandler` for AHC 3.
- Discussion [#1925](https://github.com/AsyncHttpClient/async-http-client/discussions/1925) asks how to migrate streamed consumers to AHC 3; the maintainer response declines restoring Reactive Streams because other libraries provide that policy and maintaining it adds overhead.
- Focused GitHub issue, pull-request, discussion, and local history searches found no existing AHC 3 proposal for a streaming-library-neutral suspend/resume/cancel response-body control.

## Compatibility

- `AsyncHandler.onResponseBodyStart` is a new Java `default` method, so existing handler implementations remain source- and binary-compatible and retain their previous behavior unless they override it.
- `ResponseBodyControl` is a new public interface.
- `NettyResponseBodyControl` is public only to support AHC's cross-package transport integration and is marked `@ApiStatus.Internal`; consumers should depend on `ResponseBodyControl` instead.
- HTTP/1.1 informational responses from 102 through 199 are now treated as interim and no longer reach `onStatusReceived` or `onHeadersReceived`; this matches the existing HTTP/2 behavior and prevents a 103 Early Hints response from completing the exchange before the final response. Existing 100 Continue handling and 101 protocol switching are preserved.
- Cancelling HTTP/1.1 processing from a terminal trailer or body callback now reuses a keep-alive connection because `LastHttpContent` has already been received; it previously closed that fully read connection.
- The full JDK 11 verification, including Revapi, passes.

## AI disclosure

## Test plan

- [x] On the first commit alone, the new HTTP/2 sibling-stream test reproduced shared connection-window starvation: a sibling response did not progress while the first response remained suspended.
- [x] The focused `ResponseBodyControlTest`, `Http2ResponseBodyControlTest`, `NettyResponseFutureTest`, `TimeoutTimerTaskTest`, `SuspensionAwareHttp2LocalFlowControllerTest`, and `Continue100InterceptorTest` suites pass 38 tests: 12 HTTP/1.1 control tests, 6 HTTP/2 integration tests, 9 future-lifecycle tests, 4 timeout-task tests, 5 flow-controller accounting tests, and 2 HTTP/1.1 Continue tests.
- [x] The flow-controller tests cover unchanged connection accounting without suspension, connection-only refill during suspension, DATA received before the suspension callback, overlapping suspensions, and avoiding double credit after normal accounting resumes.
- [x] `JAVA_HOME=<jdk-11> ./mvnw clean verify`: BUILD SUCCESS for the full reactor, including tests, Javadocs, coverage, and Revapi.
- [x] The locally published AHC snapshot passes the full Play WS Scala 2.13 and Scala 3.3.8 test matrices on Java 17 and Java 21: 183 integration tests and 76 unit tests pass in each combination, with 2 expected pending tests.
- [x] Play WS code validation, documentation, Scala 2 and Scala 3 MiMa checks, and dependency-tree verification pass against the locally published AHC branch.


client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -1284,6 +1286,28 @@ private static final class ConnectionCounts {
...
+        private ClientHttp2FrameCodecBuilder() {
+            // Http2FrameCodecBuilder.forClient() sets this through its package-private constructor. This subclass must
+            // use the protected no-argument constructor, so set it explicitly to preserve the client factory behavior.
+            gracefulShutdownTimeoutMillis(0);
+
+            DefaultHttp2Connection connection = new DefaultHttp2Connection(false);
+            // Refill shared credit on receipt so a suspended stream cannot starve siblings. Per-stream windows retain
+            // application backpressure, at the deliberate cost that aggregate queued data can scale with the number of
+            // suspended streams. ResponseBodyControl documents the relevant configuration bounds.
+            connection.local().flowController(new DefaultHttp2LocalFlowController(
```
hyperxpro: This turns the connection level bound off for every H2 connection, not only the ones that use the new control. With the shipped defaults (`http2InitialWindowSize` 16 MiB, `http2MaxConcurrentStreams` -1) nothing aggregate is left, so the ceiling becomes the per stream window times whatever concurrency the server picks. Before this a slow consumer was capped at 64 KiB per connection.

Can we auto refill only while a control on that connection is actually suspended, or keep an aggregate byte budget? I would rather not change the default for users who never touch `ResponseBodyControl`. The README points at `http2MaxConcurrentStreams` as the mitigation, but that is off by default too.

client/src/main/java/org/asynchttpclient/netty/handler/AsyncHttpClientHandler.java
```diff
@@ -122,6 +123,7 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable e) {
         }
 
         Channel channel = ctx.channel();
+        NettyResponseBodyControl.discardForChannelClose(channel);
```
hyperxpro: This runs before the two branches below that keep the exchange alive on purpose, the IOException filter one and `recoverOnReadOrWriteException`. In both we do not close the channel, but the control is already discarded, so the handler is left holding one where `suspend()`, `resume()` and `cancel()` are silent no ops. `detach(false)` also leaves autoRead off on a channel that survives. Should we move it down to the paths that really tear the channel down?

client/src/main/java/org/asynchttpclient/netty/handler/AsyncHttpClientHandler.java
```diff
@@ -179,7 +181,9 @@ public void channelActive(ChannelHandlerContext ctx) {
 
     @Override
     public void channelReadComplete(ChannelHandlerContext ctx) {
-        readIfNeeded(ctx);
+        if (!NettyResponseBodyControl.isSuspended(ctx.channel())) {
```
hyperxpro: Nit: this does an attribute lookup on every read complete for every channel, including each H2 stream child, and in the normal case `readIfNeeded` then no ops anyway. `suspend0()` always sets autoRead to false, so checking `isAutoRead()` first is equivalent and keeps the lookup off the hot path.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
```diff
@@ -180,14 +180,26 @@ private void handleHttp2HeadersFrame(Http2HeadersFrame headersFrame, Channel cha
             if (!abort) {
                 abort = handler.onHeadersReceived(responseHeaders) == State.ABORT;
             }
+            if (!abort) {
+                NettyResponseBodyControl control = NettyResponseBodyControl.create(
+                        channel, future::touch, () -> finishUpdate(future, channel, false));
+                abort = handler.onResponseBodyStart(control) == State.ABORT;
```
hyperxpro, line 190: When the HEADERS frame ends the stream we still create a control and call this, and `finishUpdate` runs a few lines down. A handler that suspends here gets no data and no error, and the `resume()` its consumer sends later hits `!active` and returns silently. Both new tests suspend unconditionally in `onResponseBodyStart`, so the path is exercised but the ignored suspend is never asserted. Either skip the callback when there is no body, or state on `ResponseBodyControl` that suspending in that case does nothing.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
```diff
@@ -180,14 +180,26 @@ private void handleHttp2HeadersFrame(Http2HeadersFrame headersFrame, Channel cha
...
+                if (abort) {
+                    NettyResponseBodyControl.complete(channel);
```
hyperxpro: `finishUpdate` calls `complete()` as its first statement, and if `cancel()` ran inline we already detached, so this never does anything. Same for the copy in `HttpHandler.abortAfterStartingResponseBody`. I would drop both and keep `finishUpdate` as the single completion point.

client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java
```diff
@@ -56,6 +56,17 @@ private static boolean abortAfterHandlingHeaders(AsyncHandler<?> handler, HttpHe
         return !responseHeaders.isEmpty() && handler.onHeadersReceived(responseHeaders) == State.ABORT;
     }
 
+    private boolean abortAfterStartingResponseBody(Channel channel, NettyResponseFuture<?> future,
+                                                   AsyncHandler<?> handler) throws Exception {
+        NettyResponseBodyControl control = NettyResponseBodyControl.create(
```
hyperxpro, line 64: `handleHttp2HeadersFrame` returns for every 1xx before it reaches the equivalent code, this path does not. `exitAfterIntercept` only special cases 100, so a 103 Early Hints lands here, gets a control and calls `onResponseBodyStart` on an interim response. The decoder treats 1xx as empty and hands us `LastHttpContent` right away, so we complete the future on the 103 and the real response then hits the `future.isDone()` branch in `handleRead`. `onStatusReceived` and `onHeadersReceived` have the same hole today, but this is new API and it would be good not to inherit it.

client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java
```diff
@@ -56,6 +56,17 @@ private static boolean abortAfterHandlingHeaders(AsyncHandler<?> handler, HttpHe
...
+                channel, future::touch, () -> finishUpdate(future, channel, true));
```
hyperxpro: `close` is hardcoded to true, so a `cancel()` from the last `onBodyPartReceived` throws away a connection that has nothing unread on it. The reason given in the description, that unread bytes cannot be pooled, does not apply once the body is fully read, and `cancellationFromTerminalBodyCallbackCompletesOnce` asserts `connectionCount == 2` which locks that in. Can we derive `close` from whether we have seen `LastHttpContent` yet?

client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java
```diff
@@ -92,6 +106,10 @@ private void handleChunk(HttpContent chunk, final Channel channel, final NettyRe
         if (!abort && (buf.isReadable() || last)) {
             HttpResponseBodyPart bodyPart = config.getResponseBodyPartFactory().newResponseBodyPart(buf, last);
             abort = handler.onBodyPartReceived(bodyPart) == State.ABORT;
+            // cancel() may have completed the future inline from the handler callback.
+            if (future.isDone()) {
```
hyperxpro, line 143: The same guard is missing after `onTrailingHeadersReceived` a few lines up. `cancel()` from a trailer callback completes the future and runs `onCompleted()`, then we come in here because `last` is true and call `onBodyPartReceived` on a handler that is already done. `handleHttp2TrailingHeadersFrame` got the guard, this one did not.

client/src/main/java/org/asynchttpclient/netty/handler/NettyResponseBodyControl.java
```diff
@@ -0,0 +1,162 @@
...
+    private final boolean previousAutoRead;
+
+    private volatile boolean suspended;
+    private boolean active = true;
+
+    static NettyResponseBodyControl create(Channel channel, Runnable resumeAction, Runnable cancelAction) {
+        if (!channel.eventLoop().inEventLoop()) {
+            throw new IllegalStateException("A response body control must be initialized on its channel event loop");
+        }
+        if (get(channel) != null) {
```
hyperxpro: Keying this on the channel is fragile. `HttpHandler.handleRead` returns early when `applyIoExceptionFiltersAndReplayRequest` replays and nothing detaches the control, so the channel enters `drainChannelAndOffer` still suspended with autoRead off. `LastHttpContent` never arrives, the drain never finishes, and the channel, its pool slot and the future are gone for good. If such a channel is reused later this throw fails a request that has nothing wrong with it. Keying the control to the exchange and replacing instead of throwing would remove both.

client/src/main/java/org/asynchttpclient/netty/handler/NettyResponseBodyControl.java
```diff
@@ -0,0 +1,162 @@
...
+        if (restoreAutoRead && previousAutoRead && !channel.config().isAutoRead()) {
+            channel.config().setAutoRead(true);
+        }
+    }
+
+    private void execute(Runnable task) {
+        if (channel.eventLoop().inEventLoop()) {
+            task.run();
+        } else {
+            channel.eventLoop().execute(task);
```
hyperxpro: `ResponseBodyControl` says calls made after completion have no effect, but this throws `RejectedExecutionException` at the caller once the event loop group is shutting down, which is exactly when a consumer calls `cancel()`. Catch it and treat it as the documented no op.

client/src/main/java/org/asynchttpclient/netty/timeout/ReadTimeoutTimerTask.java
```diff
@@ -42,6 +43,12 @@ public void run(Timeout timeout) {
             return;
         }
 
+        if (NettyResponseBodyControl.isSuspended(nettyResponseFuture.channel())) {
```
hyperxpro: With the request timeout disabled this re-arms forever and never reaches `clean()`, so a handler that suspends and then loses its consumer pins the future, the channel and the `AsyncHandler` for the life of the process. The description mentions the risk but there is no bound in the code. Can we cap the total suspended time, or at least log once when it runs long?

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
```diff
@@ -96,7 +103,7 @@ void startReadTimeout(ReadTimeoutTimerTask task) {
         }
     }
 
-    public void cancel() {
+    public synchronized void cancel() {
```
hyperxpro: `cancelTimeouts()` is called from `finishUpdate` on the event loop for every request, so this monitor now sits on the I/O path. The timer thread holds the same monitor inside `startReadTimeout` across `nettyTimer.newTimeout()`, which blocks on first use and contends on the wheel's queue. On an H2 connection that parks the loop for every other stream on it. Re-checking `cancelled` after scheduling, or a CAS on the `readTimeout` field, closes the same race without a shared lock.

client/src/test/java/org/asynchttpclient/Http2ResponseBodyControlTest.java
```diff
@@ -0,0 +1,420 @@
...
+    private final LinkedBlockingQueue<Boolean> cancelledLargeStreamClosed = new LinkedBlockingQueue<>();
+
+    private NioEventLoopGroup serverGroup;
+    private Channel serverChannel;
+    private ChannelGroup serverChildChannels;
+    private SslContext serverSslContext;
+    private int serverPort;
+
+    @BeforeEach
+    public void prepareServer() throws Exception {
```
hyperxpro, line 98: Nit: this is the same bootstrap as `Http2StreamingBodyFlowControlTest` and five other H2 tests. Worth pulling into a helper next to `TestUtils` at some point so the next pkitesting or ALPN change is one edit instead of eight.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
mkurz: Agreed. I chose suspension-scoped connection auto-refill and implemented it in 7bf901119.

Connections with no suspended response now retain Netty's normal 65,535-byte shared connection-window bound. When the first response on a connection is suspended, AHC returns connection credit that had already accumulated and continues returning connection-level credit while leaving per-stream credit consumption-driven. Multiple suspensions are counted, and normal connection accounting resumes after the last one ends. Credit already returned and data already queued cannot be revoked, so the remaining aggregate-buffering trade-off exists only during an active suspension and is documented in the API, README, and PR text.

The implementation also tracks connection credit returned early so application consumption after resume cannot return it twice. Tests cover the unchanged no-suspension path, DATA received before the suspension callback, overlapping suspensions, no double credit after resume, and the sibling-stream integration case.

This required adapting Netty's default local flow controller because its connection auto-refill state is private and fixed at construction; composition or subclassing cannot switch it for the lifetime of a suspension. That is a real maintenance cost, so `08be4ab5f` records that the package-private implementation is forked from Netty 4.2.17.Final and that every Netty upgrade must compare it with upstream for correctness and security fixes. A runtime-switchable Netty API is the exit path that would let AHC delete the adaptation. I did not add a hard aggregate byte budget in this PR, and I have not claimed that an upstream Netty issue already exists.

client/src/main/java/org/asynchttpclient/netty/handler/AsyncHttpClientHandler.java
mkurz: Agreed. Fixed in 631bdb2b4.

The control is no longer discarded at the start of `exceptionCaught`. It is discarded only on paths that actually close the channel. Replay now completes the old control before the old response is drained, restoring reads, while a genuinely reused exchange receives a replacement control. The ownership also moved from the channel to `NettyResponseFuture`, so it follows the exchange and stale controls cannot affect a later exchange on the same channel.

The regression test suspends the first response, triggers an `IOExceptionFilter` replay, verifies that the old channel has `autoRead` restored for draining, verifies that a replacement control is delivered, and confirms that calls through the stale control do not affect the replayed response.

client/src/main/java/org/asynchttpclient/netty/handler/AsyncHttpClientHandler.java
mkurz: Agreed. Fixed in 631bdb2b4.

`channelReadComplete` now checks `channel.config().isAutoRead()` first and returns immediately in the normal case. It only looks up the exchange/control when auto-read is already disabled, so the extra lookup is off the hot path for ordinary HTTP/1.1 channels and HTTP/2 stream children.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
mkurz: I kept the callback for bodyless responses so a handler receives one consistent response-body lifecycle callback after final headers, and documented the exact limitation in dfd26a6ed: if the final headers also end the response, `suspend()` cannot defer completion. The control becomes inactive when the callback returns and later calls are no-ops.

There are now explicit HTTP/1.1 and HTTP/2 tests that suspend in this callback, verify that completion is not deferred, and verify that later control calls have no effect.

client/src/main/java/org/asynchttpclient/netty/handler/Http2Handler.java
mkurz: Agreed. Fixed in 31434bb92: both callback-abort copies of `complete()` are gone, and `finishUpdate` is the single control-completion point.

client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java
mkurz: Agreed. Fixed in 375d57d89, with the terminator ownership tightened in ab8109aea.

HTTP/1.1 interim responses are now ignored until the final response, including the synthetic `LastHttpContent` Netty emits after them. This also avoids delivering the pre-existing status/header callbacks for a 103. The 100 Continue paths retain their request-body behavior, and 101 remains the protocol-switch exception.

The synthetic terminator is represented by AHC's existing `OnLastHttpContentCallback`, bound to the current `NettyResponseFuture`, rather than by a free-standing channel boolean. The callback restores the future when the terminator arrives and cannot leave a marker for a later exchange. The unsolicited-100 path now explains that `Continue100Interceptor` replaces the future attribute only when it actually deferred a body; if the attribute remains the same future, the 100 was unsolicited and only its synthetic terminator is consumed.

The regression test sends 103 Early Hints followed by 200 on the same connection and verifies that status, headers, `onResponseBodyStart`, body delivery, and completion each apply only to the final response, after which the connection is still reusable.

client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java
mkurz: Agreed. Fixed in 31434bb92.

The HTTP/1.1 path now records that `LastHttpContent` has been received before invoking terminal trailer/body callbacks. Cancellation closes the connection only when bytes may remain unread or keep-alive is unavailable; cancellation from a terminal callback completes normally and returns a keep-alive connection to the pool. The updated regression test verifies one handler completion and reuse of the same connection.

I also corrected the PR description: it now says that HTTP/1.1 cancellation closes the connection only when bytes may remain unread, while cancellation after terminal content has arrived can reuse a keep-alive connection.

client/src/main/java/org/asynchttpclient/netty/handler/HttpHandler.java
mkurz: Agreed. Fixed in 31434bb92.

There is now a `future.isDone()` guard immediately after `onTrailingHeadersReceived`, matching the HTTP/2 path. The regression test cancels from the trailer callback and verifies that the terminal body callback is skipped, `onCompleted` runs once, and the fully read HTTP/1.1 connection is reused.

client/src/main/java/org/asynchttpclient/netty/handler/NettyResponseBodyControl.java
mkurz: Agreed. Fixed in 631bdb2b4.

The current control is now held atomically by `NettyResponseFuture`, not in a channel attribute. Creating a control replaces and deactivates the previous one instead of throwing, and stale controls become harmless no-ops. Replay explicitly completes the old control before draining its channel, restoring reads, and the next response installs a new control for the same exchange.

The replay regression test covers the suspended-channel drain, replacement control, and stale `suspend()`/`cancel()` calls without letting those calls affect the replayed response.

client/src/main/java/org/asynchttpclient/netty/handler/NettyResponseBodyControl.java
mkurz: Agreed. Fixed in 631bdb2b4.

`NettyResponseBodyControl.execute` now catches `RejectedExecutionException`; once the channel event loop is shutting down there is no remaining transport for the control to affect, so this follows the documented post-completion no-op behavior. A regression test closes the client and then calls `suspend()`, `resume()`, and `cancel()` through the retained control, verifying that none throws.

client/src/main/java/org/asynchttpclient/netty/timeout/ReadTimeoutTimerTask.java
mkurz: I took the logging option and implemented it in ea01b27aa.

When the read-timeout interval expires while the response is still suspended and the request timeout is disabled, AHC now logs one WARN for that exchange explaining that transport resources remain retained until the application resumes or cancels it. It logs only once even though the read timeout continues to re-arm. I did not impose a new suspension deadline because that would turn an intentionally disabled request timeout into a different cancellation policy.

The test verifies the warning level, the request-timeout-disabled wording, and one-shot behavior.

client/src/main/java/org/asynchttpclient/netty/timeout/TimeoutsHolder.java
mkurz: Agreed. After rebasing onto current `main`, the synchronized implementation is gone. The upstream timeout changes now use an `AtomicBoolean` for one-shot cancellation, schedule without a shared monitor, and call `cancelIfRaced` after recording the scheduled handle. That closes the arm/cancel race without putting a monitor on the event-loop completion path. The suspension warning in ea01b27aa builds on that lock-free implementation.

client/src/test/java/org/asynchttpclient/Http2ResponseBodyControlTest.java
mkurz: Agreed that the repeated HTTP/2 bootstrap is worth consolidating. I kept that test-infrastructure refactor out of this PR because doing it properly would touch the existing HTTP/2 test classes as well and would add unrelated churn around the flow-control change. I will leave this as a focused follow-up rather than extracting a helper used only by the new test here.

mkurz:
@hyperxpro if you have time to look at this PR that would be really great, so I know how to move on in play-ws. Thanks!

client/src/main/java/org/asynchttpclient/netty/handler/AsyncHttpClientHandler.java
```diff
@@ -86,14 +87,19 @@ public void channelRead(final ChannelHandlerContext ctx, Object msg) throws Exce
 
     @Override
     public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+        Channel channel = ctx.channel();
+        Object attribute = Channels.getAttribute(channel);
+        NettyResponseFuture<?> controlFuture = responseFuture(attribute);
```
hyperxpro, line 92: I think this lookup isn't enough. Everything that ends an exchange from outside the pipeline stamps `DiscardEvent.DISCARD` on the attribute before it closes the channel, `NettyResponseFuture.cancel()` and `ChannelManager.closeChannel()` both do it, and both timeout tasks come through the latter. So `responseFuture` returns null there and the control is never deactivated.

For HTTP/1.1 that's fine, the channel is dead anyway. For HTTP/2 we close the stream child, the parent stays in the registry, and `suspendedResponses` is still incremented, so that connection keeps auto refilling for everything that lands on it later.

`active` also stays true, so a `cancel()` from `onThrowable` after a timeout still runs `cancelAction` and we go through `finishUpdate` twice.

Can we complete the control from the future side instead? `terminateAndExit()` covers abort and done, `cancel(boolean)` would need its own call.

client/src/main/java/org/asynchttpclient/netty/channel/SuspensionAwareHttp2LocalFlowController.java
```diff
@@ -0,0 +1,540 @@
...
+        }
+
+        void startAutoRefill() throws Http2Exception {
+            // DATA may already have reached a stream child channel before its handler calls suspend(). Return that
+            // outstanding connection credit too, otherwise the pre-suspension bytes could still starve siblings.
+            int unconsumedBytes = unconsumedBytes();
+            if (unconsumedBytes > 0) {
+                super.consumeBytes(unconsumedBytes);
+                autoConsumedBytes += unconsumedBytes;
+            }
```
hyperxpro, line 283: We drop the return value here. `super.consumeBytes` returns true when it wrote a WINDOW_UPDATE, and the upstream `onStreamClosed` above flushes on it for a reason, those two are the sites that can run outside a read cycle.

`suspend()` from an application thread comes in through `eventLoop().execute`, so this runs as its own task with no `channelReadComplete` after it. If ~33 KB of DATA already arrived the WINDOW_UPDATE just sits in the parent's outbound buffer, the peer has no window left to send anything, and nothing flushes it until some unrelated write comes along.

```suggestion
            if (unconsumedBytes > 0) {
                if (super.consumeBytes(unconsumedBytes)) {
                    ctx.flush();
                }
                autoConsumedBytes += unconsumedBytes;
            }
```

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/handler/AsyncHttpClientHandler.java
mkurz: Agreed. Fixed in 8b120fa9c.

The response-body control is now completed by the future's own terminal transitions. `terminateAndExit()` covers both `done()` and `abort()`, while `cancel(boolean)` completes the control after winning cancellation and before closing the channel or invoking `onThrowable`. This clears the future's control and makes retained calls no-ops even when the channel attribute has already been replaced with `DiscardEvent.DISCARD`. For a suspended HTTP/2 response, the event-loop detach invokes the suspension-ended action exactly once, so the parent connection's suspended-response count is decremented and normal connection-window accounting resumes.

`finishUpdate` still completes the control first because it must restore `autoRead` before the channel is offered back to the pool; deferring to the `future.done()` call at the end of `finishUpdate` would pool a channel with reads still suspended. The two calls are idempotent because `deactivate` is CAS-guarded: the future is the authoritative completion point for terminal transitions, while `finishUpdate` ensures transport cleanup happens before pooling.

I kept `discardForChannelClose` in `channelInactive`: it is still needed when one channel dies without terminating the exchange, particularly an I/O-filter replay that moves the same future to a replacement channel.

Two regression tests cover both external abort (including the timeout path) and explicit future cancellation. Each suspends a control, terminates the future, calls the retained control from `onThrowable`, and verifies that suspension ends once, future ownership is cleared, and the control's cancellation action is not invoked. Both tests failed before the fix and now pass.

hyperxpro (approved):

hyperxpro:
Thanks a lot!

PR #2320, opened 2026-08-28
Mark an HTTP/2 connection instead of looking it up

pavel-ptashyts:
## Problem

`ChannelManager.isHttp2(channel)` asks the pipeline for the multiplex handler by name:

```java
return channel.pipeline().get(HTTP2_MULTIPLEX) != null;
```

The write path asks it of every request, twice: once in `sendRequestWithOpenChannel` to decide
whether to store the per-request future on the channel, once in `writeRequest` to route the
write. `DefaultChannelPipeline.get(String)` walks the handler chain comparing names, and an
HTTP/1.1 connection - which has no such handler - is walked to the end to answer no. That is the
common case for anyone who has not turned HTTP/2 on.

In a profile of a client running with `setHttp2Enabled(false)`,
`DefaultChannelPipeline.context0` accounted for 83 CPU samples on this alone.

## Change

The multiplex handler is installed in exactly one place, `upgradePipelineToHttp2`, so a channel
attribute is set beside it and `isHttp2` reads that instead:

```java
public static boolean isHttp2(Channel channel) {
    return channel.hasAttr(HTTP2_CONNECTION_ATTRIBUTE);
}
```

A binary search over integer keys in a small array, rather than a walk with a string compare per
handler. `hasAttr` rather than `attr(...).get()`: the latter would add an entry to the attribute
map of every HTTP/1.1 channel just to find nothing in it.

Behaviour is unchanged in every configuration, including the ones a config check would get wrong
- see below.

## Why not check the config instead

Reading `config.isHttp2Enabled()` first would be cheaper still, and it was the first thing tried.
It is not safe: the two can disagree.

`NettyConnectListener` upgrades the pipeline on the ALPN result alone, not on the config:

```java
boolean http2Negotiated = ApplicationProtocolNames.HTTP_2.equals(alpnProtocol);
if (http2Negotiated && !uri.isWebSocket()) {
    channelManager.upgradePipelineToHttp2(channel.pipeline());
```

And ALPN can select `h2` with the flag off. `DefaultSslEngineFactory` advertises `h2` only when
`isHttp2Enabled()`, but it leaves a caller-supplied `SslContext` alone
(`config.getSslContext() != null || !config.isHttp2Enabled()`), and a caller-supplied
`SslEngineFactory` is free to advertise whatever it likes - which the WebSocket guard beside the
upgrade already accounts for in as many words: *"this guard is the backstop for a custom
SslEngineFactory that still advertises h2"*. `upgradePipelineToHttp2AfterProxyConnect` is gated on
ALPN the same way.

With `http2Enabled(false)` and such a context, a config check would route a genuine HTTP/2
connection down the HTTP/1.1 branch and write an HTTP/1.1 request onto an HTTP/2 pipeline. The
attribute costs nothing more than the config read would have saved, and cannot disagree with the
handler, being set where the handler is.

If HTTP/2 negotiated against `http2Enabled(false)` is considered unsupported, a config
short-circuit could be layered on top of this - but that is a behaviour decision rather than a
micro-optimisation, so it is not made here.

## Can the attribute go stale

No. Nothing removes `HTTP2_MULTIPLEX` from a pipeline - the only reference to it besides the
lookup is the `addLast` in the upgrade - so there is no downgrade for the two to diverge across.
Stream child channels carry neither the handler nor the attribute, so `isHttp2` answers no for
them exactly as it did before.

## Tests

`ChannelManagerHttp2MarkerTest` pins the attribute to the handler: either both say HTTP/2 or
neither does, so a later change to the upgrade cannot set one and forget the other. Three cases -
a connection never upgraded, one upgraded, and a stream channel.

Checked by mutation rather than assumption: dropping the attribute assignment fails that test and
46 of the 52 in `BasicHttp2Test`, the write path having routed HTTP/2 connections down the
HTTP/1.1 branch.

## Verification

`mvnw clean verify` - BUILD SUCCESS, 1488 tests, 0 failures, 0 errors, 26 skipped. Error Prone,
NullAway and Revapi all clean, with no revapi entries: `isHttp2` keeps its signature and the
attribute key is private.

installed on this machine, so it was run on **JDK 17** (also in the CI matrix). The JDK 11 legs of
CI on this PR are the real gate.


client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -125,6 +126,8 @@ public class ChannelManager {
     public static final String LOGGING_HANDLER = "logging";
     public static final String HTTP2_FRAME_CODEC = "http2-frame-codec";
     public static final String HTTP2_MULTIPLEX = "http2-multiplex";
+    // Set beside HTTP2_MULTIPLEX and nowhere else, so that isHttp2 can answer without a pipeline lookup.
+    private static final AttributeKey<Boolean> HTTP2_CONNECTION_ATTRIBUTE = AttributeKey.valueOf("http2Connection");
```
hyperxpro, line 130: `AttributeKey.valueOf(String)` lands in the global constant pool, so any other code in the JVM that uses "http2Connection" ends up with the same key. `SuspensionAwareHttp2LocalFlowController` already uses the class scoped form, can we do the same here ?

```suggestion
    private static final AttributeKey<Boolean> HTTP2_CONNECTION_ATTRIBUTE = AttributeKey.valueOf(ChannelManager.class, "http2Connection");
```

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -1027,10 +1030,16 @@ protected void initChannel(Channel channel) throws Exception {
...
+     * its stream children, whose own pipelines carry neither the multiplex handler nor this attribute.
+     * <p>
+     * Answered from an attribute rather than by looking {@link #HTTP2_MULTIPLEX} up in the pipeline. The two are
+     * set together and so always agree, but a pipeline lookup compares handler names down the chain, and an
+     * HTTP/1.1 connection, which has no such handler, is walked to the end to say no. The write path asks this
+     * of every request.
      */
     public static boolean isHttp2(Channel channel) {
-        return channel.pipeline().get(HTTP2_MULTIPLEX) != null;
+        return channel.hasAttr(HTTP2_CONNECTION_ATTRIBUTE);
```
hyperxpro, line 1042: `hasAttr` is true as soon as an entry exists for the key, not when the value is `TRUE`. A plain `channel.attr(HTTP2_CONNECTION_ATTRIBUTE)` read is enough to make this say HTTP/2 for an HTTP/1.1 channel, and a later `set(false)` would not undo it. `attr(...).get()` is the idiom used everywhere else in this class, so I think we hit this at some point.

```suggestion
        return Boolean.TRUE.equals(channel.attr(HTTP2_CONNECTION_ATTRIBUTE).get());
```

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -1096,6 +1105,7 @@ protected void initChannel(Channel ch) {
 
         pipeline.addLast(HTTP2_FRAME_CODEC, frameCodec);
         pipeline.addLast(HTTP2_MULTIPLEX, multiplexHandler);
+        pipeline.channel().attr(HTTP2_CONNECTION_ATTRIBUTE).set(Boolean.TRUE);
```
hyperxpro, line 1108: We set `HTTP2_STATE_KEY` on this same channel three lines below, and `writeHttp2Request` reads it back as its first statement. Why not route on that and drop the new key ? `writeRequest` does one `attr(HTTP2_STATE_KEY).get()` and passes the state in, which loses the pipeline walk, the second lookup inside `writeHttp2Request` and the second marker to keep in sync.

client/src/test/java/org/asynchttpclient/netty/channel/ChannelManagerHttp2MarkerTest.java
```diff
@@ -0,0 +1,84 @@
...
+class ChannelManagerHttp2MarkerTest {
+
+    private ChannelManager channelManager;
+    private Timer timer;
+    private EmbeddedChannel channel;
+
+    @BeforeEach
+    void setUp() {
+        timer = new HashedWheelTimer();
+        channelManager = new ChannelManager(config().build(), timer);
```
hyperxpro, line 45: Nit: nothing here needs a `ChannelManager` per test, the state under test lives on the `EmbeddedChannel`. `@BeforeAll` with a close in `@AfterAll` takes the class from about 8s to roughly a third of that. The sibling ChannelManager tests run in under 0.2s.

client/src/test/java/org/asynchttpclient/netty/channel/ChannelManagerHttp2MarkerTest.java
```diff
@@ -0,0 +1,84 @@
...
+        channel = new EmbeddedChannel();
+    }
+
+    @AfterEach
+    void tearDown() {
+        channel.finishAndReleaseAll();
+        timer.stop();
```
hyperxpro, line 52: `channelManager.close()` is missing. The constructor builds an `SslContext` and an event loop group per test, and `ChannelManagerHttp2DrainPermitTest`, `ChannelManagerHttp2SiblingLookupTest` and `NettyConnectListenerPermitLeakTest` all close it in `@AfterEach`. I replayed this setUp and tearDown three times and the fork gained about 6 fds each round, never given back. Worth guarding for null too, like the sibling tests do, otherwise a failure in `setUp` gets reported as an NPE from here.

```suggestion
        if (channel != null) {
            channel.finishAndReleaseAll();
        }
        if (channelManager != null) {
            channelManager.close();
        }
        if (timer != null) {
            timer.stop();
        }
```

client/src/test/java/org/asynchttpclient/netty/channel/ChannelManagerHttp2MarkerTest.java
```diff
@@ -0,0 +1,84 @@
...
+                "the upgrade should install the multiplex handler");
+        assertTrue(ChannelManager.isHttp2(channel), "and should report the connection as HTTP/2");
+    }
+
+    @Test
+    void aStreamChannelIsNotItsParentsConnection() {
+        // Nothing marks a stream child, and its own pipeline carries no multiplex handler either, so the two
+        // agree here as well: a stream is not the connection that multiplexes it.
+        channelManager.upgradePipelineToHttp2(channel.pipeline());
+        EmbeddedChannel stream = new EmbeddedChannel();
```
hyperxpro, line 76: This is a plain `EmbeddedChannel` with no parent, so the test asserts the same thing as `aConnectionThatWasNeverUpgradedIsNotHttp2`. The case worth pinning is a real stream child not picking up the parent attribute, which is new with this change:

```java
channel.runPendingTasks();
Channel stream = new Http2StreamChannelBootstrap(channel)
        .handler(new ChannelInboundHandlerAdapter())
        .open().syncUninterruptibly().getNow();
```

That gives a real `Http2StreamChannel` and `isHttp2` on it is false, so the assert still holds, and it would break if a Netty version ever made `attr` delegate to the parent.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
pavel-ptashyts: Moot now - the key is gone entirely, see the HTTP2_STATE_KEY thread.

One thing worth saying while it is here, though, in case it matters for a separate change: I could not find `SuspensionAwareHttp2LocalFlowController` in the repo, and there is no class-scoped `AttributeKey.valueOf` anywhere in it. All six keys in the client use the plain form - "default", "activeToken", "channelCreation", "channelIdleState", "http2ConnectionState" and the one I was adding. So the global-pool point stands, but as a point about the keys that are already there, "default" most of all, rather than about this one. Happy to open that separately if you want it.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
pavel-ptashyts: Agreed, and it goes with the key. The routing now reads the value rather than asking whether an entry exists:

```java
channel.attr(Http2ConnectionState.HTTP2_STATE_KEY).get() != null
```

which is the idiom you are pointing at, and it cannot be turned true by a read.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
pavel-ptashyts: Taken, and it is the better shape - one marker instead of two, and the second lookup goes with it. `writeRequest` reads the state and passes it in; `writeHttp2Request` takes it as a parameter and drops both its own `attr(...).get()` and the `state != null` guard around it, since the only caller now routes on that being non-null.

One difference from the pipeline lookup worth recording rather than leaving to be found: neither the state nor the multiplex handler is ever taken away by this client, so the two agree for the life of a connection, but Netty removes handlers itself during teardown on close. After that `isHttp2` keeps saying HTTP/2 where a pipeline lookup would have stopped. Both call sites sit behind an active-channel check, so nothing asks at that point - but it is a real difference and not one the tests would catch.

client/src/test/java/org/asynchttpclient/netty/channel/ChannelManagerHttp2MarkerTest.java
pavel-ptashyts: Done - one `ChannelManager` for the class in `@BeforeAll`, since what is under test lives on the `EmbeddedChannel` and the manager is there only to be called. The class reports 0.058s in a suite run now.

client/src/test/java/org/asynchttpclient/netty/channel/ChannelManagerHttp2MarkerTest.java
pavel-ptashyts: Good catch, and thank you for actually replaying it - `close()` was simply missing. It is in `@AfterAll` now with the null guards, matching the sibling tests.

client/src/test/java/org/asynchttpclient/netty/channel/ChannelManagerHttp2MarkerTest.java
pavel-ptashyts: You are right, a bare `EmbeddedChannel` has no parent, so that case was asserting what the first one already did. It opens a real stream child through `Http2StreamChannelBootstrap` now, as you suggested, which pins the part that is actually new: a stream does not inherit the connection state its parent carries, and that state is what identifies an HTTP/2 connection after this change.

pavel-ptashyts:
Round 1 addressed. The `HTTP2_STATE_KEY` suggestion was the right call and it took the other two main-code comments with it: there is no new attribute any more, `writeRequest` reads the state once and hands it to `writeHttp2Request`, which loses its own lookup and the null guard around it. So the change is now one lookup where there were two, rather than a cheaper lookup plus a second marker.

The one difference from the old pipeline lookup that survives is recorded inline: Netty removes handlers during teardown while the attribute stays, so `isHttp2` keeps answering true after a close where the lookup would have stopped. Both callers are behind an active-channel check.

Tests: one `ChannelManager` per class instead of per test, `close()` added - it was missing, and your fd count was right - and the stream case opens a real `Http2StreamChannel` rather than a bare `EmbeddedChannel` that was re-testing the first case.

`./mvnw clean verify` green. JDK 17 locally, no 11 on this machine, so the JDK 11 legs of CI are the real gate.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
pavel-ptashyts: Correction to my reply above: you were right and I was wrong. `SuspensionAwareHttp2LocalFlowController` does exist and does use the class-scoped form:

```java
AttributeKey.valueOf(SuspensionAwareHttp2LocalFlowController.class, "controller")
```

I branched from `511cd44d7`, which predates #2318, so it was not in my tree when I looked, and I reported "not in the repo" instead of "not in the tree I was looking at". My apologies for the noise.

So there is a precedent, and the rest of what I said stands with it: the five older keys still use the plain form, `"default"` most exposed of all. Nothing to change here since the key is gone, but I am happy to send that as its own change if you want the older ones brought in line.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -1027,10 +1027,17 @@ protected void initChannel(Channel channel) throws Exception {
...
     /**
-     * Checks whether the given channel is an HTTP/2 connection (i.e. has the HTTP/2 multiplex handler installed).
+     * Checks whether the given channel is an HTTP/2 connection: the parent that multiplexes streams, not one of
+     * its stream children, which carry neither the multiplex handler nor connection state of their own.
+     * <p>
+     * Answered from the {@link Http2ConnectionState} attached to the connection rather than by looking
+     * {@link #HTTP2_MULTIPLEX} up in the pipeline. The two are attached together, in
+     * {@link #upgradePipelineToHttp2}, and neither is ever taken away, so they say the same thing; but a
+     * pipeline lookup compares handler names down the chain, and an HTTP/1.1 connection, which has no such
+     * handler, is walked to the end to say no. The write path asks this of every request.
```
hyperxpro: Two claims in here are not right.

`neither is ever taken away` is wrong for the handler. `DefaultChannelPipeline.HeadContext.channelUnregistered` calls `destroy()` on close and strips every handler, so a closed connection reads not HTTP/2 by pipeline and HTTP/2 by attribute. That is the divergence you describe yourself in the `HTTP2_STATE_KEY` thread, and this sentence says the opposite of it.

`The write path asks this of every request` stopped being true in this commit. `writeRequest` reads the attribute itself now and never calls `isHttp2`. That rationale fits on the new comment in `writeRequest`, not here.

Can we say instead that the two agree for as long as the connection is live, and that after close only the attribute is left ? That is the part someone calling this from outside would actually notice.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
```diff
@@ -1097,7 +1104,8 @@ protected void initChannel(Channel ch) {
         pipeline.addLast(HTTP2_FRAME_CODEC, frameCodec);
         pipeline.addLast(HTTP2_MULTIPLEX, multiplexHandler);
```
hyperxpro, line 1124: Can we attach the state above these two lines instead of below them ?

Not a bug report: I went through all three callers and they are all on the event loop with the channel not yet in the registry or the pool, so nothing can observe the gap today. But the old code did not care about the order, the `state != null` guard covered it, and now the order is load bearing. And the failure got worse rather than better. A write landing in that gap used to go down the HTTP/2 path with an unaccounted stream slot; now it takes the HTTP/1.1 branch on an HTTP/2 pipeline and dies in `UnsupportedMessageTypeException`.

Attaching before the frame codec makes it structural rather than something every future caller has to keep true, and a racing write would then fail cleanly out of `Http2StreamChannelBootstrap` and retry. It is also what okhttp and jetty get for free, the connection object exists before any of its machinery does.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -772,10 +775,12 @@ public <T> void writeRequest(NettyResponseFuture<T> future, Channel channel) {
...
      * and the {@link NettyResponseFuture} attached to it, mirroring the HTTP/1.1 channel model.
      */
-    private <T> void writeHttp2Request(NettyResponseFuture<T> future, Channel parentChannel) {
-        Http2ConnectionState state = parentChannel.attr(Http2ConnectionState.HTTP2_STATE_KEY).get();
-
-        if (state != null && !state.tryAcquireStream()) {
+    /**
+     * @param state the connection's HTTP/2 state, which is what identified it as an HTTP/2 connection in the
+     *              first place, so the caller has it in hand
+     */
```
hyperxpro: Two javadoc blocks stacked here, so only the second one attaches to the method and the description above it is dropped from the published javadoc.

```suggestion
     *
     * @param state the connection's HTTP/2 state, which is what identified it as an HTTP/2 connection in the
     *              first place, so the caller has it in hand
     */
```

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
```diff
@@ -703,9 +703,12 @@ public <T> void writeRequest(NettyResponseFuture<T> future, Channel channel) {
...
 
-        // Route to HTTP/2 path if the parent channel has the HTTP/2 multiplex handler installed
-        if (ChannelManager.isHttp2(channel)) {
-            writeHttp2Request(future, channel);
+        // Route to HTTP/2 when the connection carries HTTP/2 state, which is attached where the multiplex
+        // handler is. Read here rather than asked of ChannelManager.isHttp2, because the HTTP/2 path needs the
+        // state itself and would otherwise look up what this line has already found.
+        Http2ConnectionState http2State = channel.attr(Http2ConnectionState.HTTP2_STATE_KEY).get();
+        if (http2State != null) {
+            writeHttp2Request(future, channel, http2State);
```
hyperxpro, line 711: Two leftovers from this: `state != null` at :858 and at :904 can never be false now that the only caller passes a non-null state. Worth dropping them along with the one you already removed at the top.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
pavel-ptashyts: Both wrong, and the first one contradicts what I wrote two threads over, which is worse than either claim on its own.

The handler does get taken away - `HeadContext.channelUnregistered` calls `destroy()` on close and strips the pipeline - so "neither is ever taken away" is exactly backwards for it. And "the write path asks this of every request" is something this commit falsified: `writeRequest` reads the state itself now, so the only production caller left is `sendRequestWithOpenChannel`.

Rewritten along the lines you suggest: the two agree for as long as the connection is live, they part on close where Netty strips the pipeline and leaves the attribute, and the callers reach this having just taken a channel from the pool and checked it is active.

client/src/main/java/org/asynchttpclient/netty/channel/ChannelManager.java
pavel-ptashyts: Agreed, and the argument for it is the right one: the ordering was free before and this branch made it load bearing, so it should be a property of the method rather than something the callers happen to make true. Moved above both `addLast` calls.

The failure inverting is the part that decided it. A racing write now reads HTTP/2 on a pipeline that has no multiplex handler yet, so it fails out of `Http2StreamChannelBootstrap` and is retried, rather than writing an HTTP/1.1 request onto an HTTP/2 pipeline.

One honest counterpart, since the window does not vanish so much as change sides: between the attribute and the multiplex handler the connection says HTTP/2 while the pipeline is still being built, and a custom outbound handler writing during the preface could observe that. Nothing in the client does, and it is the safe direction of the two, but the marker is not a pipeline-readiness signal and I would rather say so than let it read as one.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Fixed - merged into one block, so the stream child description is published with the `@param` rather than dropped.

client/src/main/java/org/asynchttpclient/netty/request/NettyRequestSender.java
pavel-ptashyts: Both gone. Same reasoning as the one at the top: `openHttp2Stream` is reached only from `writeHttp2Request` and from the pending opener it queues, and both carry the state the caller routed on, so neither guard can be false.

pavel-ptashyts:
Round 2 addressed.

The ordering one was the substantive change: the state is attached before the frame codec and the multiplex handler now, so the upgrade itself keeps the invariant rather than the callers keeping it by accident. As you say, that also inverts the failure - a write racing into the window opens a stream on a pipeline without a multiplex handler, which fails out of the bootstrap and retries, instead of an HTTP/1.1 write onto an HTTP/2 pipeline. I had a second look at what the reorder could disturb: nothing between the two points reads the attribute, adding either handler does not replay `channelActive` or start an inbound read, and the SETTINGS listener is still added after both with the state already in place either way.

The `isHttp2` javadoc is rewritten - both claims in it were wrong, and the first one contradicted what I had written on the `HTTP2_STATE_KEY` thread, which is the part I should have caught. The stacked javadoc blocks on `writeHttp2Request` are one block, and the two dead `state != null` guards in `openHttp2Stream` are gone.

One thing beyond the review, because this commit is what made it stale: `Continue100Interceptor` explained itself in terms of `writeRequest` looking the multiplex handler up in the pipeline, which it no longer does. The reason it gives still holds, so only the wording changed.

`./mvnw clean verify` green. JDK 17 locally, no 11 on this machine.

hyperxpro (approved):

hyperxpro:
Thanks a lot!

PR #2322, opened 2026-08-31
Provide an opt-in zero-copy response body view

### LINKED ISSUE #2321: Provide an opt-in zero-copy response body byte-array view
## Problem

The standard response path

```text
executeRequest()
  -> AsyncCompletionHandlerBase
  -> ResponseBuilder
  -> NettyResponse
```

copies the response body twice before returning it as a byte array:

1. `EagerResponseBodyPart` copies every incoming Netty `ByteBuf` into an
   AHC-owned `byte[]`.
2. `NettyResponse.getResponseBodyAsBytes()` calls
   `getResponseBodyAsByteBuffer()`, which allocates another array for the
   complete body and copies all body parts into it.

The second copy is unnecessary when a response arrived in one chunk because
the `EagerResponseBodyPart` already owns one complete AHC-managed array.

#2303 removed this extra copy from `getResponseBody(Charset)`, but deliberately
kept it in `getResponseBodyAsBytes()` because that method hands a mutable array
to application code. Byte-array consumers therefore cannot opt into the same
optimization even when they can guarantee that they will not mutate the body.

## Goal

Add an explicit opt-in API that may return shared response-body storage. Keep
all existing behavior of `getResponseBodyAsBytes()` unchanged.

## Proposed API

Add this default method to `org.asynchttpclient.Response`:

```java
/**
 * Returns the entire response body as a byte array that may share
 * its storage with this response.
 *
 * <p>The returned array must be treated as read-only. Modifying it may
 * change the content subsequently returned by this response's other
 * body accessors.
 *
 * <p>The implementation is not required to return shared storage.
 * Depending on the response representation, this method may still
 * return a copy. No array identity is guaranteed between calls.
 *
 * <p>Use {@link #getResponseBodyAsBytes()} when an independently owned,
 * mutable array is required.
 *
 * @return the entire response body as a possibly shared byte array
 */
default byte[] getResponseBodyAsBytesView() {
    return getResponseBodyAsBytes();
}
```

A default method keeps third-party `Response` implementations binary- and
source-compatible. Implementations that cannot expose shared storage preserve
the existing safe behavior. Aliasing is enabled explicitly by the caller.

The name and Javadoc must make it clear that the returned array can be shared
and must be treated as read-only.

## NettyResponse implementation

`NettyResponse` should override the method:

```java
@Override
public byte[] getResponseBodyAsBytesView() {
    return bodyParts.size() == 1
            ? bodyParts.get(0).getBodyPartBytes()
            : getResponseBodyAsBytes();
}
```

Expected behavior:

| Response shape | Behavior |
| --- | --- |
| No body parts | Return an empty array through the normal fallback |
| One `EagerResponseBodyPart` | Return its internal `byte[]` without copying |
| One `LazyResponseBodyPart` | Use `getBodyPartBytes()`; perform the one necessary copy from the `ByteBuf`, but do not allocate a second aggregate array |
| Multiple parts | Use the existing concatenation through `getResponseBodyAsBytes()` |

The implementation must call `getBodyPartBytes()` and must not access
`ByteBuf.array()` directly. A `ByteBuf` can be direct, can be a slice with a
non-zero `readerIndex` or `arrayOffset`, and its backing array can contain bytes
outside the readable region.

## Integration with #2303

Remove or replace the existing private `sharedBodyBytes()` method so the fast
path has a single implementation:

```java
@Override
public String getResponseBody(Charset charset) {
    return new String(getResponseBodyAsBytesView(), charset);
}
```

This preserves the string-decoding optimization from #2303 while exposing it to
byte-array consumers only through the new opt-in method.

## Required constraints

The change must not:

- change the behavior of `getResponseBodyAsBytes()`;
- change the behavior of `getResponseBodyAsByteBuffer()`;
- make any existing `byte[]` API implicitly shared;
- retain the original network `ByteBuf` or change its reference count;
- access `ByteBuf.array()` directly;
- change `ResponseBodyPartFactory.EAGER` or `ResponseBodyPartFactory.LAZY`;
- add synchronization or atomic operations to the response hot path;
- cache multi-part concatenation as part of this change.

`getResponseBodyAsBytes()` must continue to return an independently owned
array. Mutating an array returned by the existing method must not affect the
response.

### One EAGER part

- Assert that the view is the exact array owned by the part.
- Assert that repeated view calls return the same array for this
  `NettyResponse` shape. This is an implementation performance property, not a
  general `Response` API identity guarantee.

### Defensive-copy preservation

- Keep the #2303 coverage proving that the old method does not return the
  part's array.
- Prove that two calls to `getResponseBodyAsBytes()` return distinct arrays.
- Prove that mutating the old method's result does not affect the response.

### Lazy part over a slice/direct buffer

Create a `LazyResponseBodyPart` over the readable `Hello World` region in
`XXXHello WorldYYY` and assert that the view contains no surrounding bytes.
Also verify that the call does not change `readerIndex`, `writerIndex`, or
`refCnt`. Cover a direct buffer as well.

### Multiple parts

- Verify correct concatenation order with no missing or duplicate bytes.
- Split a multi-byte UTF-8 character across two parts.
- Verify that `getResponseBody(Charset)` produces the same result for one and
  multiple parts.

### Empty body

Verify that the view returns an empty array.

### Default implementation

Add a test for a `Response` implementation that does not override the new
method. It must delegate to `getResponseBodyAsBytes()` without requiring a
third-party implementation change.

## Benchmark

Compare:

```java
response.getResponseBodyAsBytes();
response.getResponseBodyAsBytesView();
```

Dataset matrix:

- one EAGER part: 512 B, 4 KiB, 16 KiB, and 128 KiB;
- one LAZY part: 4 KiB and 16 KiB;
- 2 and 8 EAGER parts;
- empty body.

Expected result:

- one EAGER part: the view has no body-sized allocation or copy;
- one LAZY part: one body-sized copy instead of two;
- multiple parts: performance is equivalent to the existing method;
- `getResponseBodyAsBytes()` behavior and performance remain unchanged.

No hard numerical CI threshold is required. Include allocation benchmark
results in the pull request.

## Acceptance criteria

- An opt-in shared/read-only byte-array API is available.
- Existing `Response` implementations do not need to change.
- One `EagerResponseBodyPart` is returned without another copy.
- Sliced and direct lazy parts preserve readable-region correctness.
- `getResponseBodyAsBytes()` still returns an independent array.
- The `getResponseBody(Charset)` optimization from #2303 is preserved.
- Netty buffer lifecycle and reference counts are unchanged.
- `mvnw.cmd clean verify` passes on JDK 11, including Error Prone, NullAway,
  and Revapi.
- The pull request includes allocation benchmark results.

## Out of scope

- Removing the initial `ByteBuf` -> `EagerResponseBodyPart` copy.
- Preallocating from `Content-Length`.
- A new multi-chunk accumulator.
- Caching the aggregated multi-part body.
- Changing the LAZY response lifecycle.
- A zero-copy `InputStream`.
- Changing the semantics of `getResponseBodyAsBytes()` or
  `getResponseBodyAsByteBuffer()`.

After release, downstream byte-array consumers can opt in by replacing a method
reference with `response::getResponseBodyAsBytesView`. For single-part responses,
JFR should then show the aggregate allocation in
`NettyResponse.getResponseBodyAsByteBuffer()` disappearing; the
`EagerResponseBodyPart` allocation remains.

pavel-ptashyts:
## Summary

- add `Response#getResponseBodyAsBytesView()` as a compatible default method
  with an explicit possibly-shared, read-only contract
- return the existing byte array for a single eager Netty body part
- use the body part accessor for lazy/direct/sliced buffers, avoiding a second
  aggregate copy without exposing `ByteBuf` storage
- preserve defensive-copy behavior for `getResponseBodyAsBytes()`,
  `getResponseBodyAsByteBuffer()`, and `getResponseBodyAsStream()`
- reuse the new accessor for string decoding so the fast path from #2303 stays
  centralized

Closes #2321.

## Compatibility and safety

The default implementation delegates to `getResponseBodyAsBytes()`, so existing
third-party `Response` implementations keep their current behavior. Netty uses
the optimized path only for exactly one body part. Empty and multipart bodies
continue through the existing aggregation path.

Tests cover eager identity and repeated access, defensive-copy isolation,
heap and direct lazy slices, reader/writer indices, reference counts, empty
bodies, multipart ordering and split UTF-8 characters, stream isolation, and
default-method delegation.

## Verification

- `mvnw.cmd -B -ntp -pl client -Dtest=NettyAsyncResponseTest test`
- `mvnw.cmd clean verify`
  - Amazon Corretto 11.0.32.1
  - 1,489 tests; 0 failures; 0 errors; 26 skipped
  - Error Prone, NullAway, Javadocs, GPG signing, and Revapi passed

## Allocation benchmark

JMH 1.37, Corretto 11.0.32.1, one fork, three 500 ms warmups, five
500 ms measurements, `gc` profiler. Values are allocated bytes per operation,
rounded to whole bytes; zero means below the profiler's resolution.

| Body shape | `getResponseBodyAsBytes()` | `getResponseBodyAsBytesView()` |
| --- | ---: | ---: |
| Empty | 16 B/op | 16 B/op |
| 1 eager part, 512 B | 576 B/op | 0 B/op |
| 1 eager part, 4 KiB | 4,160 B/op | 0 B/op |
| 1 eager part, 16 KiB | 16,448 B/op | 0 B/op |
| 1 eager part, 128 KiB | 131,136 B/op | 0 B/op |
| 1 lazy direct part, 4 KiB | 8,272 B/op | 4,112 B/op |
| 1 lazy direct part, 16 KiB | 32,848 B/op | 16,400 B/op |
| 2 eager parts, 4 KiB each | 8,256 B/op | 8,208 B/op |
| 8 eager parts, 4 KiB each | 32,832 B/op | 32,832 B/op |

The eager single-part view measured about 2.3 ns/op regardless of body size,
versus 322 ns/op for 4 KiB and 11.0 us/op for 128 KiB defensive copies. Lazy
single-part allocation is halved as expected. Multipart bodies retain the
payload-sized aggregation allocation.


client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
```diff
@@ -193,6 +193,11 @@ public byte[] getResponseBodyAsBytes() {
         return getResponseBodyAsByteBuffer().array();
     }
 
+    @Override
+    public byte[] getResponseBodyAsBytesView() {
+        return bodyParts.size() == 1 ? bodyParts.get(0).getBodyPartBytes() : getResponseBodyAsBytes();
```
hyperxpro, line 198: `bodyParts.size() == 1` is not a property of the response, it is the number of non-empty chunks we happened to read. The same 48 byte body handed over in one chunk gives the caller our array, in two chunks gives them a copy. Any proxy that re-chunks flips it. gzip flips it the other way, because the decompressor output does not get split at `httpClientCodecMaxChunkSize`, so a 20000 byte body arrives as 4 parts uncompressed and 1 part compressed.

It is not even stable for a fixed server. On a fresh connection the split is at 2048 bytes, on a warm pooled one it is around 8192, so the caller gets a copy on request 1 and our array on request 2. A test that makes one client per test with a 4 KiB fixture never sees the sharing path.

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
```diff
@@ -193,6 +193,11 @@ public byte[] getResponseBodyAsBytes() {
         return getResponseBodyAsByteBuffer().array();
     }
 
+    @Override
+    public byte[] getResponseBodyAsBytesView() {
```
hyperxpro, line 197: This array is not only ours. `EagerResponseBodyPart` returns its `bytes` field by identity, and that is the same array we already passed to the user's `AsyncHandler.onBodyPartReceived`, the same one `TransferCompletionHandler` gives to every `TransferListener`, and the one `ResumableAsyncHandler` wraps writable in a `ByteBuffer`. So the caller can corrupt those, and they can corrupt the caller's view after we returned it.

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
```diff
@@ -224,19 +229,9 @@ public String getResponseBody() {
...
-     * because a multi-byte character can straddle a part boundary.
-     */
-    private byte[] sharedBodyBytes() {
-        return bodyParts.size() == 1 ? bodyParts.get(0).getBodyPartBytes() : getResponseBodyAsBytes();
-    }
-
     @Override
     public String getResponseBody(Charset charset) {
-        return new String(sharedBodyBytes(), charset);
+        return new String(getResponseBodyAsBytesView(), charset);
```
hyperxpro, line 234: This used to go through the private `sharedBodyBytes()`. Now it is a virtual call to a public method on a non-final class, so any subclass that overrides the view silently changes `getResponseBody()`, `getResponseBody(Charset)` and `toString()` as well, and the String and the bytes of one response can disagree. `getStatusCode`, `getHeaders` and `isRedirected` right next to it are `final`. Can we keep a private helper here and leave the public method for callers?

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
```diff
@@ -224,19 +229,9 @@ public String getResponseBody() {
         return getResponseBody(withDefault(extractContentTypeCharsetAttribute(getContentType()), UTF_8));
     }
 
-    /**
```
hyperxpro, line 227: This comment was carrying two things. The reason multi part gets concatenated, a multi byte char can straddle a part boundary, which is the only place that is written down. And the invariant from #2303, that the array does not escape the method. #2303 landed a month ago and its tests are still in this file. If we are reversing it, the PR description should say why instead of just deleting the comment.

client/src/main/java/org/asynchttpclient/Response.java
```diff
@@ -55,6 +55,23 @@ public interface Response {
      */
     byte[] getResponseBodyAsBytes();
 
+    /**
+     * Returns the entire response body as a byte array that may share its storage with this response.
+     *
+     * <p>The returned array must be treated as read-only. Modifying it may change the content subsequently returned
```
hyperxpro, line 61: "this response's other body accessors" is too narrow. The array belongs to the `HttpResponseBodyPart`, not to the response, and it is reachable from the handler callback, from `TransferListener` and from `getResponseBodyAsByteBuf()`. It also goes the other way: `bb.setByte(10, '9')` on that ByteBuf changes what the caller's read-only view says. Worth spelling out who the other holders are.

client/src/main/java/org/asynchttpclient/Response.java
```diff
@@ -55,6 +55,23 @@ public interface Response {
...
+     * by this response's other body accessors.
+     *
+     * <p>The implementation is not required to return shared storage. Depending on the response representation, this
```
hyperxpro, line 64: Can we say here that it also depends on `ResponseBodyPartFactory`? EAGER shares, LAZY always copies through `ByteBufUtil.getBytes(buf.duplicate())`, and the caller cannot see which one the client was built with. Someone who develops an in-place transform against LAZY will ship it green and corrupt bodies on the default config.

client/src/main/java/org/asynchttpclient/Response.java
```diff
@@ -55,6 +55,23 @@ public interface Response {
...
+     * method may still return a copy. No array identity is guaranteed between calls.
+     *
+     * <p>Use {@link #getResponseBodyAsBytes()} when an independently owned, mutable array is required.
```
hyperxpro, line 67: `getResponseBodyAsBytes()` never promised this. Its own javadoc just says it returns the body, and it is an interface method anyone can implement, so returning a cached array is legal there. I would not promise ownership here on behalf of a method that does not guarantee it.

client/src/main/java/org/asynchttpclient/Response.java
```diff
@@ -55,6 +55,23 @@ public interface Response {
...
+     *
+     * @return the entire response body as a possibly shared byte array
+     */
+    default byte[] getResponseBodyAsBytesView() {
```
hyperxpro, line 71: Please add an `@implSpec` saying the default calls `getResponseBodyAsBytes()`. After reading the text above, the obvious thing for an implementor is `getResponseBodyAsBytes() { return getResponseBodyAsBytesView().clone(); }` and override nothing else, and that recurses until it blows the stack.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
```diff
@@ -113,32 +120,108 @@ public void testGetResponseBodyDecodesOnePartAndSplitPartsIdentically() {
...
+            assertEquals(refCnt, slice.refCnt());
+        } finally {
+            slice.release();
+        }
+    }
+
+    @Test
+    public void testGetResponseBodyAsBytesViewReadsDirectLazyPart() {
+        byte[] backing = "XXXHello WorldYYY".getBytes(StandardCharsets.UTF_8);
+        ByteBuf direct = Unpooled.directBuffer(backing.length);
```
hyperxpro, line 156: `direct` is allocated, written and sliced before the `try`, so anything throwing in there leaks a direct buffer and the leak detector fails the build. Open the `try` right after the allocation.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
```diff
@@ -113,32 +120,108 @@ public void testGetResponseBodyDecodesOnePartAndSplitPartsIdentically() {
...
         List<HttpResponseBodyPart> bodyParts = new LinkedList<>();
-        bodyParts.add(new LazyResponseBodyPart(Unpooled.wrappedBuffer(backing, 3, 11), true));
+        bodyParts.add(new EagerResponseBodyPart(Unpooled.wrappedBuffer("Hello World".getBytes(StandardCharsets.UTF_8)), true));
         NettyResponse response = new NettyResponse(new NettyResponseStatus(null, null, null), null, bodyParts);
 
-        assertEquals("Hello World", response.getResponseBody(StandardCharsets.UTF_8));
-        assertEquals("Hello World",
-                new String(response.getResponseBodyAsStream().readAllBytes(), StandardCharsets.UTF_8));
+        byte[] view = response.getResponseBodyAsBytesView();
+        assertSame(bodyParts.get(0).getBodyPartBytes(), view);
```
hyperxpro, line 183: The javadoc says no array identity is guaranteed between calls and that the implementation may return a copy, and then we assert both of those here. Swap the fixture to `LazyResponseBodyPart` and the same assertion fails. One of the two has to give, otherwise the next person cannot tell which one is the contract.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
```diff
@@ -113,32 +120,108 @@ public void testGetResponseBodyDecodesOnePartAndSplitPartsIdentically() {
...
+        assertSame(view, response.getResponseBodyAsBytesView());
     }
 
     @Test
     public void testGetResponseBodyAsBytesDoesNotShareTheBodyPartArray() {
+        byte[] expected = "Hello World".getBytes(StandardCharsets.UTF_8);
```
hyperxpro, line 189: `expected` is both the array we wrap into the part and the oracle we compare against. It holds today only because `EagerResponseBodyPart` copies. Add the obvious `hasArray()` fast path there and `expected` becomes the part's live storage, so `firstCopy[0] = 'X'` mutates it too and both assertions pass on a corrupt response. `Unpooled.wrappedBuffer(expected.clone())` keeps it honest.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
```diff
@@ -113,32 +120,108 @@ public void testGetResponseBodyDecodesOnePartAndSplitPartsIdentically() {
...
+            assertEquals(readerIndex, slice.readerIndex());
+            assertEquals(writerIndex, slice.writerIndex());
+            assertEquals(refCnt, slice.refCnt());
+        } finally {
+            slice.release();
+        }
+    }
+
+    @Test
+    public void testGetResponseBodyAsBytesViewReadsDirectLazyPart() {
```
hyperxpro, line 154: This state cannot happen on a real request. `AsyncHttpClientHandler.channelRead` releases the message in a finally as soon as `onBodyPartReceived` returns and `LazyResponseBodyPart` never retains, so by the time anyone holds a `Response` the buffer is at refCnt 0 and this call throws `IllegalReferenceCountException`. With `-Dio.netty.buffer.checkAccessible=false` it does not throw, it returns recycled pool memory instead. The test keeps the buffer alive by hand, which makes LAZY look supported when it is not. Not something you introduced, but adding coverage here signs off on it.

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
```diff
@@ -193,6 +193,11 @@ public byte[] getResponseBodyAsBytes() {
         return getResponseBodyAsByteBuffer().array();
     }
 
+    @Override
+    public byte[] getResponseBodyAsBytesView() {
+        return bodyParts.size() == 1 ? bodyParts.get(0).getBodyPartBytes() : getResponseBodyAsBytes();
+    }
```
hyperxpro, line 199: Nit: an empty body still goes down the aggregating path and allocates a `byte[0]` plus a `ByteBuffer` on every call. HEAD, 204 and 304 are common enough that a `static final byte[] EMPTY` short circuit is worth it, and it is the one case here that really is zero allocation.

hyperxpro (changes requested):

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
pavel-ptashyts: You are right that it is not a property of the response, and it cannot be made into one while the method returns an array - a lone part either exists to share from or it does not, and that is decided by how the body arrived.

So the javadoc now says it rather than glossing it: sharing depends on the chunking, on a proxy having re-chunked, on compression, and on the parts the implementation was given, none of it visible from the call site, and the same body from the same server may share on one response and copy on the next. A caller who reads that and still wants the method wants it for what it is: an accessor that will not copy when it does not have to, with no promise about when that is.

If you would rather have no such accessor than one with that contract, say so and I will close this - the honest version of the javadoc is not a flattering advertisement for the method.

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
pavel-ptashyts: Agreed, and the javadoc understated it badly. It now names the holders: the array is reachable from the part handed to `onBodyPartReceived`, it is what a `TransferCompletionHandler` passes each `TransferListener` - `fireOnBytesReceived(content.getBodyPartBytes())`, by identity - and it is what `getResponseBodyAsByteBuf()` wraps. And it says the traffic goes both ways: a write through any of those changes what this returns.

One correction to my own first draft of that text, which said the array is "passed to `onBodyPartReceived`". It is not; the *part* is, and the array is reachable through it. Fixed before pushing.

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
pavel-ptashyts: Fixed - `getResponseBody(Charset)` goes back through the private helper, so overriding the view cannot change what a response's text says.

I nearly wrote that the neighbouring methods are `final` "for the same reason", which would have been me inventing a rationale for someone else's code, so that is not in there. And the helper does not make a subclass consistent in general: it still calls the overridable `getResponseBodyAsBytes()` for the multi-part case, and the string accessors are overridable too. What it does is stop the *view* being in that path, which is what this change put there.

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
pavel-ptashyts: Both halves taken. The multi-byte straddle note is back in the helper's comment, and it is the reason it is worth having.

The invariant is a different matter: it has genuinely changed, and the comment now says so instead of repeating a claim that stopped being true. #2303 kept the array in because nothing exposed it; this exposes it deliberately, so the comment says the array does leave the client through the view, and the view's own javadoc carries the terms. That is the part the description should have argued rather than the comment quietly losing, and it now does.

client/src/main/java/org/asynchttpclient/Response.java
pavel-ptashyts: Agreed on both counts - the scope and the direction. It names the part as the owner and the response as one holder among several, and it says a write through the ByteBuf changes what the view returns, not only the other way round.

client/src/main/java/org/asynchttpclient/Response.java
pavel-ptashyts: Added, and it turned into the sharpest sentence in the javadoc. What I did not do is state it as "shares only under EAGER": `NettyResponse`'s constructor takes whatever `HttpResponseBodyPart` it is handed, so the absolute would not have been true either. It says instead that whether a lone part hands over storage is the part's business - eager returns the array it holds, lazy copies out of its buffer every call - so a response built from lazy parts never shares whatever the method says.

client/src/main/java/org/asynchttpclient/Response.java
pavel-ptashyts: Right, and dropped. It no longer says anything about what `getResponseBodyAsBytes()` guarantees; it says a caller who needs a modifiable array should copy what it receives, and that the other accessor is implemented by whoever implements the interface, so its own contract is the thing to read.

client/src/main/java/org/asynchttpclient/Response.java
pavel-ptashyts: Added in substance, though not as a tag: `@implSpec` is not configured for the javadoc plugin here and is used nowhere else in the client, so the build fails on `unknown tag: implSpec`. Rather than put a doclet tag configuration into the shared pom for one comment, it is an "Implementation note" paragraph saying the default returns `getResponseBodyAsBytes()`.

Also narrowed while writing it: the recursion needs the default to be *left in place*. Overriding both methods and implementing bytes as `view().clone()` is fine and is what an implementor may well want. The text says that.

Happy to send the `-tag implSpec` plugin configuration as its own change if you would like the tag available - there are other places that could use it.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
pavel-ptashyts: Gone with the test - see the LAZY thread below. Worth saying it was a real bug and not a nit: the allocation succeeded and everything up to the `try` could have thrown with a live direct buffer, so any failure in that window would have been reported as a leak rather than as itself.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
pavel-ptashyts: You are right that one of the two had to give, and it is the test that was claiming too much rather than the javadoc that was too weak. The interface keeps the weak contract - it cannot promise identity when it does not control the chunking - and the assertion stays, but as what it actually is: `NettyResponse`'s own behaviour, that when this response *can* share it does rather than copying anyway. The test says so in as many words now, and `NettyResponse#getResponseBodyAsBytesView` documents the same thing, so the two do not read as the same claim at different strengths.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
pavel-ptashyts: Fixed with the `expected.clone()` you suggested. I also added the content assertions the test was missing: it checked `assertNotSame` on both copies without ever checking what was in them, so an implementation that copied faithfully and one that copied garbage were indistinguishable. Same for the sharing test next to it, which asserted identity and nothing about the bytes.

client/src/test/java/org/asynchttpclient/netty/NettyAsyncResponseTest.java
pavel-ptashyts: Agreed, and the new coverage is gone: the direct-buffer test is deleted and the pre-existing lazy test is back to exactly what #2303 left, without the view assertions this branch had added to it. Adding coverage for a mode that cannot occur on a real request would have been this PR quietly signing off on it, which is your point.

The lazy branch of the accessor is still exercised, through `getResponseBody(Charset)` - it shares the same private helper - so removing this costs no reachable path.

The underlying thing stands on `main` and I would rather raise it separately than have it ride along here: `channelRead` releases in a `finally` and `LazyResponseBodyPart` never retains, so a `Response` built from lazy parts holds buffers at refCnt 0, and `ResponseBodyPartFactory.LAZY` is not usable as a client-wide setting. Say the word and I will open it, with or without a fix.

client/src/main/java/org/asynchttpclient/netty/NettyResponse.java
pavel-ptashyts: Done - a shared `EMPTY_BODY` returned before the aggregating path, so HEAD, 204 and 304 allocate neither the array nor the buffer to wrap it. Safe to share because nothing can be written through a zero-length array. It sits in the private helper, so the view and the string accessors both get it; `getResponseBodyAsBytes()` keeps its copying path untouched, since that one hands the array out as the caller's own.

pavel-ptashyts:
Round 1 addressed. Two of the comments are about the contract rather than the code, and they are right that neither can be designed away while the method returns a `byte[]`: whether a lone part exists to share from is decided by the chunking, and where an array is shared it is also reachable from `onBodyPartReceived`, from a `TransferCompletionHandler`'s listeners and from `getResponseBodyAsByteBuf()`. So the javadoc states both instead of glossing them, and states that array identity between calls is not guaranteed either.

That leaves an accessor whose honest description is "will not copy when it does not have to, and will not tell you when that is". If that is not worth an interface method, I would rather you said so than have me argue it - it closes cleanly, and `getResponseBodyAsByteBuf()` already covers the multi-part case zero-copy for anyone who can take a buffer.

The rest:

* `getResponseBody(Charset)` goes back through the private helper, so overriding the view cannot change a response's text as well. The helper keeps #2303's note about multi-byte characters straddling a part boundary, and now says the array does leave the client rather than claiming it does not.
* The javadoc no longer promises anything on behalf of `getResponseBodyAsBytes()`, and records the recursion an implementor would otherwise write - narrowed to the case that actually recurses, which is leaving the default in place.
* `@implSpec` is a paragraph rather than a tag: it is not configured for the javadoc plugin here and the build fails on `unknown tag`. Offered separately.
* Empty bodies short-circuit to a shared empty array.
* Tests: content assertions where there were only identity ones, `expected.clone()` so the oracle is not the fixture, and the LAZY coverage dropped rather than signing off on a mode that cannot occur on a real request.

`./mvnw clean verify` green. JDK 17 locally, no 11 on this machine.

hyperxpro (approved):

PR #1736, opened 2020-09-28
Use original method in redirects for HEAD / OPTIONS requests

TomGranot:
Originally, when setFollowRedirect is set to true (like in Head302Test), the expected flow is HEAD --> GET (based on the behaviour of Redirect30xInterceptor - see https://github.com/AsyncHttpClient/async-http-client/commit/3c25a42289bf679cfff40ca5ef0e77be85215deb for the reasoning).

Following a change to the interceptor (to fix https://github.com/AsyncHttpClient/async-http-client/issues/1728), Head302Test started going on an infinite loop caused by the way the handler in the test was set up (to account for the original HEAD --> GET flow and not the new HEAD --> HEAD --> HEAD.... flow).

This PR does 3 things:

* Changes Redirect30xInterceptor to set all redirects of a HEAD request to be HEADs as well
* Change Head302Test to account for the new flow
* Notes a flaky test in AsyncStreamHandlerTest that was found during work on the PR

Issue #1042, opened 2015-11-19
POST body is dropped on following 301 redirect


yuppie-flu:
### Environment:
async-http-client 1.9.31
netty 3.10.4.Final
AsyncHttpClientConfig.Builder().setFollowRedirect(true)

### Preconditions
"url-original" accepts POST and redirect to "url-redirected" with HTTP 301 status

### Steps:
Send request with asyncHttpClient.preparePost("url-original").setBody("not empty body").execute().get()

### Expected 
Request redirected to "url-redirected" with original body

### Actual
Request redirected to "url-redirected" with empty body

### Notes
1.8.14 version works as expected for this case

slandelle:
No, actually, behaviour was wrong in 1.8.
Check what browsers do: on 301, method is transformed to GET and body is dropped.

yuppie-flu:
Ok, may be this is the case for default behaviour.
But don't you think It will be worth to provide some configuration option to make client follow the rfc2616 specification.
For example, **curl** has the one called: **--post301**

slandelle:
As you API obviously doesn't work with web browsers, I really suspect that you don't need the "Permanent Redirect" semantics you get with 301 that makes the redirect cached by web browsers.

Then, why not use another status code that would keep both original method and body: 307?

PR #2325, opened 2026-09-05
fix(netty): preserve non-POST redirect methods

mkurz:
## Summary

- Limit AHC's historical 301/302 POST-to-GET rewrite to POST requests.
- Keep request content whenever a redirect preserves the request method, except for 303's explicit body-dropping behavior.
- Validate only the body representation that AHC selected for transmission and reject consumed, non-resettable streams before connecting to the redirect target.
- Cover standard methods, extension methods, caller-added redirect statuses, and cross-origin behavior.

## Problem

`Redirect30xInterceptor` applied the historical POST-to-GET behavior for 301 and non-strict 302 responses to every method other than GET, HEAD, and OPTIONS. A PUT, PATCH, DELETE, or extension-method request therefore became a bodyless GET after either redirect.

The body decision was also coupled to a fixed list of methods and status codes. GET, HEAD, and OPTIONS requests can mechanically carry bodies in AHC, but those bodies were dropped on 301 and 302 even though the methods were retained. A caller-added status in the public mutable `REDIRECT_STATUSES` set likewise retained the method while silently dropping its body.

[RFC 9110 sections 15.4.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.4.2) and [15.4.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.4.3) scope the compatibility allowance to changing POST to GET. They do not permit rewriting every other method in the same way. [RFC 10008 section 2.5](https://www.rfc-editor.org/rfc/rfc10008.html#section-2.5) explicitly requires QUERY not to use the POST exceptions.

## Change

Apply the legacy 301/302 rewrite only when the original method is POST. Derive body handling from the method decision: keep the body whenever the method is preserved, except for 303, which drops the body regardless. This differs deliberately from checking whether a request is "not POST": a POST receiving a caller-added status such as 300 also keeps its method and must therefore keep its body.

The resulting behavior is:

- POST on 301 or non-strict 302: switch to GET and drop the body.
- POST on strict 302: retain POST and the body.
- PUT, PATCH, DELETE, QUERY, and extension methods on 301 or 302: retain the method and body.
- GET, HEAD, and OPTIONS on 301 or 302: retain the method and any explicitly attached body.
- 303: drop the body; methods that require rewriting switch to GET, while GET, HEAD, and OPTIONS remain unchanged.
- 307 and 308: retain the method and body under the existing policy.
- Caller-added redirect statuses: retain the body whenever the method is retained.

This is a deliberate compatibility change. It corrects the standards scope and removes method-preserving, body-dropping combinations while retaining the long-established POST behavior for 301 and non-strict 302.

## Redirect-body validation

The replay checks now use one private `BodyRepresentation` selection that mirrors `NettyRequestFactory.body()` precedence. This matters because some request-builder setters leave lower-priority representations in place. Validation now examines only the body AHC selected for transmission; for example, a stale multipart `InputStreamPart` no longer rejects a redirect whose selected body is a byte array.

Selected raw `InputStream` bodies and `InputStreamBodyGenerator` instances that report no mark/reset support are rejected before AHC connects to the redirect target only if the stream has already been consumed. An early redirect in response to `Expect: 100-continue` can leave the stream untouched, so the target can still receive its first transmission without reset support. This preflight is intentionally partial: a stream may report mark support yet be closed or fail to reset after the first send. The existing write-time replay guard remains the final authority for those cases. A generic `BodyGenerator` can also produce an unknown-length or non-repeatable body, but discovering that would require calling its one-shot `createBody()` early, so its existing write-time behavior is unchanged.

## Redirect security

For a cross-origin 301 or 302, newly preserved request content is replayed to the redirect target. This follows the body-replay trust model AHC already uses for strict 302, 307, and 308. Existing redirect security continues to strip Authorization, Proxy-Authorization, Realm credentials, user-supplied Cookie headers, and Cookie objects when the origin changes. Tests assert that credentials reach the original server, do not reach the target, and that the method, content type, and body reach the target intact.

The same consideration applies to HTTPS-to-HTTP redirects: this pull request can replay content on 301 or 302 that the previous method/body rewrite discarded. Gating only `keepBody` on a scheme downgrade would preserve the method while silently deleting its payload, reproducing the data-corruption shape this change removes. If AHC adopts a downgrade restriction, it should refuse the redirect itself and apply uniformly to all keep-body statuses, including strict 302, 307, and 308. That transport-policy decision is left to a focused follow-up; this pull request does not alter AHC's existing downgrade policy.

## History checked

The broad conversion is established behavior rather than a recent accident. Issue #989 requested browser-compatible 301 handling, issue #1042 retained the body drop for POST, and pull request #1736 later exempted HEAD and OPTIONS from method rewriting. This change retains the established POST rule while making body preservation follow the resulting method decision. A search did not find an existing issue or pull request specifically correcting PUT, PATCH, DELETE, extension methods, or caller-added redirect statuses.

## Compatibility

There is no public API change.

The following redirect behavior changes intentionally:

- PUT, PATCH, DELETE, and extension methods on 301 and non-strict 302 retain their original method and content instead of becoming bodyless GET requests.
- GET, HEAD, and OPTIONS requests with explicitly attached content retain it on 301 and 302 instead of sending a bodyless second request.
- Requests followed through caller-added redirect statuses retain content whenever their method is retained, including POST on a registered 300.
- A selected raw `InputStream` or `InputStreamBodyGenerator` that has already been consumed and lacks mark/reset support now fails before connecting to the redirect target. For newly preserved non-POST 301/302 requests, this replaces the previous silent success as a bodyless GET with an explicit replay failure. Existing keep-body redirects fail earlier and use the new redirect-level error message. Untouched streams deferred by `Expect: 100-continue` remain usable at the redirect target. Streams that advertise mark support but cannot actually reset still fail at write time.
- A request with a replayable selected body and a stale, lower-priority non-replayable representation no longer fails validation.
- Cross-origin and HTTPS-to-HTTP 301/302 redirects can now receive content that the old body-dropping behavior suppressed; credential stripping remains unchanged.

POST on 301 and non-strict 302, strict-302 method policy, 303 method policy, and 307/308 method policy remain unchanged.

## AI disclosure

## Test plan

- [x] The new focused regressions reproduced eight body-loss failures for GET, HEAD, OPTIONS, and caller-added 300 before the generalized body rule was applied.
- [x] The selected-body coexistence regression reproduced the stale multipart `InputStreamPart` false rejection before validation was aligned with outbound-body precedence.
- [x] Both deferred-stream regressions failed with the unconditional preflight and pass with the consumption check. They assert no stream reads or body bytes at the origin before its 307 response and exact body bytes at the target, for raw `InputStream` and `InputStreamBodyGenerator` bodies.
- [x] `./mvnw -pl client -Dtest=RedirectBodyTest,RedirectCredentialSecurityTest test` on JDK 11: 73 tests passed.
- [x] `./mvnw clean verify` on JDK 11: BUILD SUCCESS (full reactor, including tests, Javadocs, artifact signing, coverage, and Revapi).

Generated with OpenAI Codex.


client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -117,18 +117,17 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
-                        (statusCode == FOUND_302 && !strict302);
-                boolean switchToGet = !methodAlreadyPreserved &&
-                        (statusCode == SEE_OTHER_303 || (!isQuery && legacyRedirectToGet));
-                boolean keepBody = queryRedirect ||
+                // RFC 9110 limits the historical 301/302 POST-to-GET rewrite to POST.
+                boolean switchToGet = !bodylessMethod &&
+                        (statusCode == SEE_OTHER_303 ||
+                                (isPost && (statusCode == MOVED_PERMANENTLY_301 ||
+                                        (statusCode == FOUND_302 && !strict302))));
+                boolean keepBody = (!bodylessMethod && !isPost &&
```
hyperxpro: schemeDowngrade is already computed further down and only feeds stripAuth, so nothing stops an https to http 301 from replaying the body over cleartext now. I ran it with a card number in a PUT and it goes out in the clear, where main sends a bodyless GET.

The JDK 11 client refuses that hop entirely and hands the 301 back to the caller. Can we gate keepBody on the scheme too ? 307 and 308 have the same hole today so the full fix is probably its own PR, but this one widens it to 301 and 302.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -117,18 +117,17 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
                 future.setScramContext(null);
 
                 String originalMethod = request.getMethod();
-                boolean isQuery = QUERY.equals(originalMethod);
-                boolean methodAlreadyPreserved = originalMethod.equals(GET) ||
+                boolean isPost = originalMethod.equals(POST);
+                boolean bodylessMethod = originalMethod.equals(GET) ||
```
hyperxpro: bodylessMethod is not really true, we do support GET and OPTIONS with a body and BasicHttpTest.getShouldAllowBody depends on it.

The result is that the same GET with content keeps its body on 307 and loses it on 301. Neither status changes the method, so I do not think the drop is justified in either case. Dropping !bodylessMethod from keepBody and renaming this to something like neverRewrittenMethod would make the two lines agree with each other.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -117,18 +117,17 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
-                boolean switchToGet = !methodAlreadyPreserved &&
-                        (statusCode == SEE_OTHER_303 || (!isQuery && legacyRedirectToGet));
-                boolean keepBody = queryRedirect ||
+                // RFC 9110 limits the historical 301/302 POST-to-GET rewrite to POST.
+                boolean switchToGet = !bodylessMethod &&
+                        (statusCode == SEE_OTHER_303 ||
+                                (isPost && (statusCode == MOVED_PERMANENTLY_301 ||
+                                        (statusCode == FOUND_302 && !strict302))));
+                boolean keepBody = (!bodylessMethod && !isPost &&
+                        (statusCode == MOVED_PERMANENTLY_301 || statusCode == FOUND_302)) ||
```
hyperxpro: REDIRECT_STATUSES is public and mutable and the class javadoc sells that as a feature. Because this arm lists 301 and 302 only, a caller who registers 300 gets the method preserved and the body silently dropped, so a PUT goes out with Content-Length 0 and no error.

I could not find another client that produces that combination. Would it be better to keep the body for anything that is not 303 and not the POST case, rather than enumerating ?

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -117,18 +117,17 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
                 boolean strict302 = statusCode == FOUND_302 && config.isStrict302Handling();
-                // RFC 10008 section 2.5 excludes QUERY from the legacy POST-to-GET behavior.
-                boolean queryRedirect = isQuery &&
-                        (statusCode == MOVED_PERMANENTLY_301 || statusCode == FOUND_302);
-                boolean legacyRedirectToGet = statusCode == MOVED_PERMANENTLY_301 ||
-                        (statusCode == FOUND_302 && !strict302);
-                boolean switchToGet = !methodAlreadyPreserved &&
-                        (statusCode == SEE_OTHER_303 || (!isQuery && legacyRedirectToGet));
-                boolean keepBody = queryRedirect ||
+                // RFC 9110 limits the historical 301/302 POST-to-GET rewrite to POST.
```
hyperxpro: Nit: can we keep the RFC 10008 reference here as well ? QUERY still comes out right after this change, but only as a side effect of !isPost, and someone widening the legacy branch later has nothing in the file telling them why QUERY must not be rewritten.

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
```diff
@@ -565,6 +588,36 @@ void query301CrossOriginStripsCredentialsAndPreservesRequest() throws Exception
...
+                    .execute()
+                    .get(5, TimeUnit.SECONDS);
+
+            assertNull(put301AuthOnTarget.get(),
+                    "Authorization must be stripped on a cross-origin PUT redirect");
+            assertNull(put301CookieOnTarget.get(),
+                    "Cookie must be stripped on a cross-origin PUT redirect");
+            assertEquals("PUT", put301MethodOnTarget.get());
+            assertEquals("application/octet-stream", put301ContentTypeOnTarget.get());
+            assertEquals("sensitive-content", put301BodyOnTarget.get());
```
hyperxpro: This is the assertion I keep coming back to. The test name is about stripping credentials, but this line pins the body being replayed to the other origin as intended behaviour, and that is the part actually worth discussing.

I am not saying the behaviour is wrong, curl and HC5 and the JDK all do it. But it should be named for what it pins rather than arriving under a hardening name.

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
```diff
@@ -565,6 +588,36 @@ void query301CrossOriginStripsCredentialsAndPreservesRequest() throws Exception
...
+
+            client.preparePut("http://127.0.0.1:" + portA + "/redirect-put-301-to-b")
+                    .setHeader("Authorization", "Bearer secret-token")
+                    .setHeader("Cookie", "session=secret-session")
+                    .setHeader("Content-Type", "application/octet-stream")
+                    .setBody("sensitive-content")
+                    .execute()
+                    .get(5, TimeUnit.SECONDS);
+
+            assertNull(put301AuthOnTarget.get(),
```
hyperxpro: These two only check that the target never saw the header. If we ever stopped sending Authorization on the first hop, this still passes and still reports the stripping as working.

crossDomainRedirectStripsCookieHeader further down records the value on A and asserts it before checking B. Worth doing the same here.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
```diff
@@ -281,16 +281,45 @@ public synchronized void reset() throws IOException {
...
             "PATCH, 301",
             "PATCH, 302",
             "DELETE, 301",
-            "DELETE, 302"
+            "DELETE, 302",
+            "CUSTOM, 301",
+            "CUSTOM, 302"
     })
-    public void putPatchAndDelete301And302KeepExistingBehavior(String method, int statusCode) throws Exception {
+    public void nonPost301And302KeepMethodAndBody(String method, int statusCode) throws Exception {
```
hyperxpro: These all redirect to getTargetUrl(), so every case in this class is same origin. Fine for checking the method and body survive, but it means nothing here covers the cross origin case, which is the part that changes what actually leaves the process.

That is also why the suite stays green. Could we add at least one row that redirects to a different host ?

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
```diff
@@ -281,16 +281,45 @@ public synchronized void reset() throws IOException {
         }
     }
 
-    @ParameterizedTest(name = "{0} on {1} keeps the existing GET rewrite")
+    @RepeatedIfExceptionsTest(repeats = 5)
+    public void put301WithNonRepeatableBodyGeneratorFailsPromptly() throws Exception {
```
hyperxpro: ensureBodyReplayable runs before we touch the new connection, but it only looks at InputStreamPart and a missing File, so this case gets all the way into NettyInputStreamBody.write and fails after we have already connected to the redirect target.

Fetch fails this before issuing the redirect request at all. Can we teach ensureBodyReplayable about streamData and InputStreamBodyGenerator so both shapes fail in the same place ? Worth noting this also turns a FileInputStream PUT that worked on main into a hard failure.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
hyperxpro: `schemeDowngrade` is computed below but only feeds `stripAuth`, so an https to http 301 now replays the body in cleartext. The JDK client refuses that hop completely. Can we gate `keepBody` on the scheme as well ? 307 and 308 have the same problem already so maybe that part is a follow up.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
hyperxpro: `bodylessMethod` is not accurate, we do support GET and OPTIONS with a body, see `BasicHttpTest.getShouldAllowBody`. So the same GET keeps its body on 307 but loses it on 301. Neither changes the method, so why drop it ?

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
hyperxpro: `REDIRECT_STATUSES` is public and mutable, so a registered 300 keeps the method here but loses the body and you get a `PUT` with `Content-Length: 0`. Should this keep the body for anything that is not 303 and not POST, instead of listing the two ?

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
hyperxpro, line 125: Nit: can we keep the RFC 10008 reference ? QUERY is still correct after this but only as a side effect of `!isPost` now.

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
hyperxpro, line 627: This pins the body being replayed cross origin as expected behaviour, under a name that reads like a hardening test. Can we rename it to say what it asserts ?

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
hyperxpro, line 621: These pass even if we stopped sending `Authorization` on the first hop. `crossDomainRedirectStripsCookieHeader` records the value on A and asserts it first, we should do the same here.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
hyperxpro, line 332: These all use `getTargetUrl()` so every case here is same origin, which is also why the suite stays green. Can we add a row that redirects to a different host ?

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
hyperxpro, line 290: `ensureBodyReplayable` only looks at `InputStreamPart` and `File`, so this one fails in `NettyInputStreamBody.write` after we already connected to the target. Can we check `streamData` and `InputStreamBodyGenerator` up front too ?

hyperxpro (changes requested):
Overall looks solid

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Thanks, this is a real concern. I did not gate `keepBody` on the scheme because that would preserve PUT/QUERY/etc. while silently deleting their payload, which recreates the data-corruption shape this change is intended to remove. If AHC restricts HTTPS-to-HTTP body replay, I think it should refuse the redirect itself and apply the policy uniformly to every keep-body redirect, including strict 302, 307, and 308, which can already replay bodies on a downgrade.

I updated the PR description to disclose that 301/302 can now replay content on a downgrade where the old method/body rewrite discarded it. I agree that a uniform downgrade policy is best handled as a focused follow-up rather than changing only these newly corrected paths here.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Fixed in ba32a323d. I renamed `bodylessMethod` to `methodAlreadyPreserved` and now derive body handling from the redirect decision: the body is kept whenever the method is preserved, except for 303. I added 301/302 coverage for GET, HEAD, and OPTIONS with explicit bodies; all three retain their method, content type, and bytes.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Fixed in ba32a323d. I used `statusCode != SEE_OTHER_303 && !switchToGet` for `keepBody`, which is slightly broader than "not 303 and not POST": it keeps the body whenever the method is preserved. That distinction matters for a caller-added 300 with POST; excluding POST there would retain POST while sending `Content-Length: 0`, the same corruption we want to avoid. The regression test temporarily registers 300 and covers both POST and PUT, asserting the original method, content type, and body bytes.

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Kept in ba32a323d. The policy comment now cites RFC 9110 for the POST-scoped historical rewrite and RFC 10008 section 2.5 for QUERY, even though QUERY is handled by the general non-POST rule.

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
mkurz: Renamed in 271010164 to `put301CrossOriginReplaysBodyAndStripsCredentials`, so the test name states both sides of the behavior it pins rather than reading only as a hardening test.

client/src/test/java/org/asynchttpclient/RedirectCredentialSecurityTest.java
mkurz: Fixed in 271010164. The origin handler now records Authorization and Cookie independently of the target handler, and the test first asserts the exact values on the original PUT before asserting that both are absent from the redirected request. It therefore fails if either credential was never sent on the first hop.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
mkurz: Added a dedicated different-host case in 271010164. It starts the PUT at `127.0.0.1` and follows the 301 Location to `localhost`, then asserts that the redirected request retains PUT, its content type, and its body bytes. I kept this as a separate test rather than adding an origin flag to every method/status row so the origin-boundary condition is explicit. The security test separately verifies credential stripping across a different-port origin.

client/src/test/java/org/asynchttpclient/RedirectBodyTest.java
mkurz: Fixed in two commits. 872800aa1 centralizes the selected body representation using the exact precedence from `NettyRequestFactory.body()` and routes multipart, file, and length checks through it. That also fixed a false rejection in the merged #2316 code: a stale lower-priority multipart `InputStreamPart` no longer rejects a redirect when the selected body is a replayable byte array, with a regression test for that coexistence case.

`78930e8dc` then preflights the selected raw `streamData` or `InputStreamBodyGenerator`. If it reports no mark/reset support, the interceptor now fails before connecting to the redirect target with `Redirect request body InputStream does not support mark/reset and cannot be replayed`.

The preflight is necessarily partial: a stream can advertise mark support but still be closed or fail to reset after the first write, so the existing write-time replay guard remains the final authority. A generic `BodyGenerator` can also conceal replayability until its one-shot `createBody()` is called, so that path retains its existing write-time behavior. The affected tests now assert the redirect-level preflight message.

mkurz:
@hyperxpro looking forward to your review ;)

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
```diff
@@ -243,66 +241,106 @@ public boolean exitAfterHandlingRedirect(Channel channel, NettyResponseFuture<?>
...
-            return null;
+        InputStream inputStream = null;
+        if (bodyRepresentation == BodyRepresentation.STREAM_DATA) {
+            inputStream = request.getStreamData();
+        } else if (bodyRepresentation == BodyRepresentation.INPUT_STREAM_BODY_GENERATOR) {
+            inputStream = ((InputStreamBodyGenerator) request.getBodyGenerator()).getInputStream();
         }
-        if (request.getFile() != null) {
-            return request.getFile();
+        if (inputStream != null && !inputStream.markSupported()) {
```
hyperxpro: Could we check whether the stream has actually been consumed before rejecting it here? With Expect: 100-continue, the server can return a 307 before the body is sent, so the redirect target would be reading it for the first time. I tested both a raw InputStream and an InputStreamBodyGenerator. 

Both succeed on the base revision but fail here before any stream reads. Checking future.isStreamConsumed() before requiring mark/reset support should preserve that behavior.

hyperxpro:
> @hyperxpro looking forward to your review ;)

Sorry took longer than expected

client/src/main/java/org/asynchttpclient/netty/handler/intercept/Redirect30xInterceptor.java
mkurz: Fixed in d732c06e8. You're right: my preflight required reset support before checking whether the stream had been consumed, unlike the existing write-time guard. It now checks `future.isStreamConsumed()` before requiring mark/reset support for the selected raw `InputStream` or `InputStreamBodyGenerator`. An early redirect before `100 Continue` therefore leaves the target free to read the stream for the first time.

I added a parameterized regression covering both representations with a non-empty, non-resettable stream. The origin sends a 307 while still expecting `100 Continue`, without reading the request body; the test asserts zero client-side stream reads and zero body bytes received at the origin before the redirect, then checks the original bytes and PUT method at the target. Both cases failed with the unconditional preflight and pass with this fix. The existing consumed-stream rejection tests still pass.

Multipart validation deliberately does not use this flag: `NettyMultipartBody` does not update it, even after transmission, so gating the `InputStreamPart` check on it would disable that protection. The pre-existing multipart guard from #2316 can still reject an untouched `InputStreamPart` after an early redirect; safely supporting that case needs separate multipart transmission tracking. This fix leaves that protection unchanged.

mkurz:
> > @hyperxpro looking forward to your review ;)
> 
> Sorry took longer than expected

No worries at all - thanks for taking the time to test both stream representations and catch this.

mkurz:
@hyperxpro assuming this gets merged, would it be possible to cut a new release so I can use it downstream? Thanks!

hyperxpro (approved):

hyperxpro:
> @hyperxpro assuming this gets merged, would it be possible to cut a new release so I can use it downstream? Thanks!

I plan to cut release next week as there are some fixes I need to perform. Is it possible to use SNAPSHOT till then?

hyperxpro:
Thanks a lot!

mkurz:
> > @hyperxpro assuming this gets merged, would it be possible to cut a new release so I can use it downstream? Thanks!
> 
> I plan to cut release next week as there are some fixes I need to perform. Is it possible to use SNAPSHOT till then?

Next week is ok, thanks!

mkurz:
@hyperxpro any eta? thanks!

hyperxpro:
Should be released by Sunday night.

